PR #31, opened 2013-12-11
Adding `interactplot` function

mwaskom:
The `interactplot` function is being added to help visualize models with two-way interactions between continuous variables. The plot is made by fitting an interaction model and then drawing a contour plot with the two predictor variables on each axis and the fitted values (y hats) color-encoded in the contours. So the joint distribution of the data observations is clear, a scatterplot is then drawn over the contours. The resulting image looks like this:

screen shot 2013-12-10 at 11 18 39 pm

This PR has the new function and some additions to the linear models notebook (including an example for an alternative presentation that bins one of the variables).


cmardiros:
Hi! I'm loving Seaborn. This might not be the best place to ask, but I have real problems understanding how to interpret this plot with respect to the interaction between x1 and x2 and the response y. Yes, I can visually see where yhat is highest/lowest but I don't understand what it's telling me about the interaction between x1 and x2. Using the above plot as an example, would you please provide an interpretation? Seems like a very useful tool for EDA, I just need to understand how to interpret it.

Incidentally I posted a question on Cross Validated exactly about this, about an interactplot made using seaborn, would be great to get your input on it:

http://stats.stackexchange.com/questions/240978/how-to-interpret-this-interaction-plot-contour-in-the-context-of-eda-for-regre

Thanks!

mwaskom:
Here's where I got the idea for this plot from, which may be helpful http://mbjoseph.github.io/2013/08/18/vague_intxn.html

cmardiros:
Thanks so much for posting that link. It's very useful. However, I still find the plot unintuitive. The article says:

"the expected value of y increases as either covariate increases and the interaction term affects the magnitude this increase"

Looking at the plot, though, in both your example above and the one in the link, the expected value of y increases as the covariates increase __or__ decrease (the bottom left corner is nearly as intensely coloured as the top right corner). So I don't understand how to interpret this. It would be really helpful if you could offer your own interpretation of the above example.

I'm assuming you've developed this plot to use in your own work. When have you found it most useful, if I may ask?

Thank you so much.

mwaskom:
For the figure in the first post in this thread, I would say something like "The effect of x1 on y depends on x2. When x2 is low, x1 has a negative effect on y. When x2 is high, x1 has a positive effect on y". Of course, that formulation implies some causation although without knowing anything else about the variables it would be equally valid to swap x1 and x2 in that description.

cmardiros:
Thanks, that's really helpful.

Issue #13, opened 2013-11-20
Allow for importing without manipulating the rcParams


mwaskom:
As much as I like the default style (that's why I made it the default), there are situations where people might want to use the plotting functions, but not the style. It actually used to be the case that you would have to call

```
seaborn.set()
```

after importing, but I decided that was kind of gross. Once #10 is sorted out it will be easy to revert the style changes that get made, but even then I think it might be nice to have something like

```
import seaborn.apionly as sns
```

which would give you the whole namespace, but avoid the call to `rcmod.set()` in `__init__.py`.

I'm not sure I've ever seen this design pattern in a Python package so I'm not certain of the correct/clean way to do it. Please pipe up if you have an idea.

PR #126, opened 2014-03-07
Add seaborn.apionly pseudomodule to get the api without the style

### LINKED ISSUE #10: restore **origin** theme and not **default** theme.
Hi, 

I found of about `seaborn.reset_defaults()` , may I suggest a `seaborn.reset_origin()` ? Which will partialy take into account user matplotlibrc file and user modification ?

I also found that `matplotlib.rcParams.update(matplotlib.rcParamsOrig)` is not reliable and that you should make a copy of default param before importing seaborn.

If you are not against and I find some time I could make a PR.

mwaskom:
I'm not sure this is the right way to do this, but it seems to work. The idea is that if you have you own rc parameters and don't want seaborn to mess with them, you can do `import seaborn.apionly as sns` (or `from seaborn.api import regplot`, etc.) and carry on as normal.

This works by importing seaborn as usual and then calling the new `reset_orig()` function, which updates the rcParams with `matplotlib.rcParamsOrig`, which I think respects any custom `matplotlibrc` files.

Closes #10 and #13.

### DIFF
--- a/doc/v0.3.0.txt
+++ b/doc/v0.3.0.txt
@@ -26,6 +26,8 @@ Other changes
 Using the package
 ^^^^^^^^^^^^^^^^^
 
+- If you want to use plotting functions provided by the package without setting the matplotlib style to a seaborn theme, you can now do `import seaborn.apionly as sns` or `from seaborn.apionly import regplot`, etc. This is using the (also new) `seaborn.reset_origin()` function, which returns the rc parameters to what they are at matplotlib import time — i.e. they will respect any custom `matplotlibrc` settings on top of the matplotlib defaults.
+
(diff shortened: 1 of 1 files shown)


dubourg:
Unfortunately `seaborn` is not the only python module that can change the `rcParams`. Indeed, jupyter/ipython-notebook also happens to tweak params such as the backend... So when I `import seaborn.apionly as sns` within a notebook it reverts my backend to the default `TkAgg` while it was set to `NotebookAgg` for my ipython-notebook session. Hence figure won't display unless I `%matplotlib inline` again.

How about writing `seaborn`'s `apionly.py` like its `__init__.py` without the call to `set()`? I understand it's harder to maintain if you create/delete/rename namespaces...

PR #140, opened 2014-03-21
Add bias-corrected, accelerated (BCa) conf. intervals

phobson:
I pulled the `known_ci_bca` values in the tests from the [referenced paper](http://staff.ustc.edu.cn/~zwp/teach/Stat-Comp/Efron_Bootstrap_CIs.pdf), but had to tweak them *inward* (narrower) but 0.1 on each side. Not sure if that's because slight numeric differences in computation from 1996 vs now, or what, but I feel ok about it. I've validated this algorithm before with one of Efron's other papers (can't find it at the moment.

I figure after we iterate on this and get it in a good spot, I'll think more about this https://github.com/mwaskom/moss/issues/7#issuecomment-38298687 and add in the ABC method

### DIFF
--- a/seaborn/algorithms.py
+++ b/seaborn/algorithms.py
@@ -77,6 +77,140 @@ def bootstrap(*args, **kwargs):
     return np.array(boot_dist)
 
 
+def percentiles(a, pcts, axis=None):
+    """Like scoreatpercentile but can take and return array of percentiles.
+
+    Parameters
+    ----------
+    a : array
+        data
+    pcts : sequence of percentile values
+        percentile or percentiles to find score at
+    axis : int or None
+        if not None, computes scores over this axis
+
+    Returns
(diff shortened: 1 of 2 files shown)


phobson:
Thinking we should close this. Ideally, I think it'd be nice to just pass a bootstrapper to the function itself, and the arch library has just about everything one could want (https://github.com/bashtage/arch). no point in reinventing wheels.

mwaskom:
Yeah, sorry, this was good work, but I wanted to make sure I understood that math before this landed, and I never got around to that :)

phobson:
No problem. I'm exploring putting a bootstrapped kwarg in functions that'll
just take a lambda that'll return the lower and upper limits (or whatever
values your functions need).

Does that sound worth pursuing to you?

PR #182, opened 2014-04-26
Add the scale_font option to plotting_context

mwaskom:
This should make it easier to change the fontsize of the plots. `set`, `set_context`, and `plotting_context` now take a `font_scale` option to make all the font elements larger or smaller. (e.g. `font_scale=2` makes the fonts twice as large as they would otherwise be in that context.

This closes #159


keisuke-nakata:
I found that `font_scale` parameter does not work with `context=None` (i.e., default context).  
`rcmod.py:plotting_context()` uses the `font_scale` parameter only if `context is not None`. (https://github.com/mwaskom/seaborn/blob/14078fe9b4bb0b6b6fc957d6dfa0d18dc0adbbef/seaborn/rcmod.py#L381)

Is this an intended behavior? (Just passing `context` can avoid this so there is no problem, but I feel it a bit confusing).  
Thank you in advance.

PR #214, opened 2014-06-02
add flip_axes option to factorplot

fdeheeger:
I create a new PR so it comes from a specific branch instead of master.

From the previous discussion #212, i know it is not your preferred option, but it might help.


fdeheeger:
Nothing to add to the past days commits ! Let's drop this patch.
Amazing work, thanks.

PR #400, opened 2014-12-16
Add hue to JointGrid

oesteban:
This PR adds the hue option to JointGrid, enabling the representation of mixture models.

I'm totally unaware of the internals of seaborn, so I apologize in the case of hard-coded features (such as colors).


oesteban:
This PR allows the creation of this kind of plots:
paper-joint-kde-realdata

(From our recent paper: http://dx.doi.org/10.1101/018945, Figure 4)

oesteban:
@mwaskom , is it worth fixing the tests? Do you find this PR interesting?

mwaskom:
This has come up a couple of times and I agree it would nice to have hue-based conditional jointplots, but I don't think this is the way to do it.

PR #415, opened 2015-01-09
fixing issue where setting vmin/vmax affects row/col_colors

funnell:
When setting vmin/vmax and row/col_colors in clustermap, the row/col_colors are affected (I set vmax=0.5 here):
ipython

This pull request should fix that.


mwaskom:
Thanks!

Issue #361, opened 2014-11-11
Seaborn should respect categorical order when sorting pd.Categorical objects


shoyer:
For example, this adaption of the "Grouped boxplots" example should work (if using pandas 0.15 or higher) even without specifying `x_order`:
```python
import seaborn as sns
sns.set(style="ticks")

tips = sns.load_dataset("tips")
days = ["Thur", "Fri", "Sat", "Sun"]
tips['days'] = pd.Categorical(tips['day'], days)

g = sns.factorplot("day", "total_bill", "sex", tips, kind="box",
                   palette="PRGn", aspect=1.25)
g.despine(offset=10, trim=True)
g.set_axis_labels("Day", "Total Bill")
```

If you using a pandas method to do the sorting, then this is a pandas bug.

mwaskom:
The strange behavior of unordered pandas categoricals sort of defeats the utility of relying on `.unique()`, @shoyer, cf this issue thread: https://github.com/pydata/pandas/issues/9148

JanSchulz:
IMO if categorical variables (dtype "category") are to be plotted, the categroies should be used directly instead of `unique()`. E.g. I would expect that if I plot a lickert scale via bar plots, I expect that the complete scale is shown, not only the categories which are used. That was also once upon the time the reasoning why unique returned (all) categories (but this was changed in  https://github.com/pydata/pandas/issues/8559#issuecomment-64189570).

mwaskom:
Interesting, I think that's a reasonable point @JanSchulz

mwaskom:
So to make explicit what we want to happen on a categorical axis:

- For objects with ordered categorical datatype, show all categories in the correct category order
- For objects with unordered categorical datatype, show all categories in the order they appear in the Series
- Otherwise, show all unique values in the order they appear in the Series/array/list

Also this has to happen by inspecting the object attributes, not with any special pandas functions, because seaborn has to run on pandas < 0.15.

Does that sound right? If so, smarter pandas folk, what is the cleanest way to go about doing this?

shoyer:
@mwaskom Here is my suggested implementation:

```python
def get_categories(values):
    if hasattr(values, 'categories'):
        # values is a pd.Categorical
        return np.asarray(values.categories)
    else:
        return pd.unique(values)
```

[...]

PR #431, opened 2015-01-23
Fix bug in Grid legends

mwaskom:
Previously, `add_legend` was passing a list of handles for the plots that had been drawn and the list of all hue levels to the matplotlib legend code. Because matplotlib just zipped these and didn't raise
when they were of differenet lengths, this meant that sometimes the right colors but wrong labels were used.

This isn't globablly optimal, because I think the behavior we want is to show the hue levels that could have existed in the plot (because they are in `hue_order` but don't, because they weren't in the DataFrame.

However, doing that is a bit tricky because we'll have to make a proxy artist, you can't just pass `None` or something similar as a handle.

This fixes the bug identified in https://github.com/mwaskom/seaborn/issues/361#issuecomment-71140105. @shoyer, this does make the labels correct, but doesn't address the broader point that all levels of the variable should show up in the legend even when they're not in the plot data. Opening now but holding off on merge until I can think a little bit about what could be done to get a "full legend".

Thanks for the complete test case :+1:

```
import pandas as pd
import numpy as np
import seaborn as sns

def cut_diverging(array, n=9):
    mag = max(-array.min(), array.max())
    return pd.cut(array, np.linspace(-mag, mag, num=(n + 1)))

rs = np.random.RandomState(0)
df = pd.DataFrame({'x': rs.rand(100),
                   'y': rs.rand(100),
                   'z': 1 + rs.randn(100)})
df['z_cat'] = cut_diverging(df.z, 9)

categories = df.z_cat.cat.categories
palette = sns.color_palette('RdBu_r', 9)

g = sns.FacetGrid(df, hue='z_cat', hue_order=categories,
                  palette=palette, aspect=1.3, size=3)
g.map(plt.scatter, 'x', 'y', s=50)
g.add_legend()
```
shoyer

### DIFF
--- a/doc/releases/v0.6.0.txt
+++ b/doc/releases/v0.6.0.txt
@@ -18,6 +18,8 @@ This is a major release from 0.5. The main objective of this release is to unify
 
 - Added the ``line_kws`` parameter to :func:`residplot` to change the style of the lowess line, when used.
 
+- Added open-ended ``**kwargs`` to the ``add_legend`` method on :class:`FacetGrid` and :class:`PairGrid`, which will pass additional keyword arguments through when calling the legend function on the ``Figure`` or ``Axes``.
+
 - Added a catch in :func:`distplot` when calculating a default number of bins. For highly skewed data it will now use 10 bins, where previously the reference rule would return "infinite" bins and cause an exception in matplotlib.
 
 - Fixed a bug in :class:`FacetGrid` and :class:`PairGrid` that lead to incorrect legend labels when levels of the ``hue`` variable appeared in ``hue_order`` but not in the data.
--- a/doc/releases/v0.6.0.txt
+++ b/doc/releases/v0.6.0.txt
@@ -20,3 +20,4 @@ This is a major release from 0.5. The main objective of this release is to unify
 
 - Added a catch in :func:`distplot` when calculating a default number of bins. For highly skewed data it will now use 10 bins, where previously the reference rule would return "infinite" bins and cause an exception in matplotlib.
 
+- Fixed a bug in :class:`FacetGrid` and :class:`PairGrid` that lead to incorrect legend labels when levels of the ``hue`` variable appeared in ``hue_order`` but not in the data.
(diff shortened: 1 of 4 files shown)


shoyer:
Thanks!

I am a little surprised that this handles the palette colors properly even with the missing values, but that does appear to be the case.

mwaskom:
Most of the logic to handle missing values was already in the body of `map()`, this just addresses a bug in how the handles and labels were paired.

Poking around a little bit, there's no obvious way to make correct proxy handles to show all levels in the legend. You can extract the artist class from the other handles in `_legend_data`, but `FacetGrid` needs to be as general as possible and there's not a uniform API.

An alternative would be to add an invisible handle by creating a line artist or something with 0 alpha and then add that with the correct label. That would at least show all of the labels, and in some sense give a stronger hint that certain levels are absent, though for the kind of thing you're doing in the example you won't get as pretty a "colorbar".

mwaskom:
Alright, I think I like this solution. When levels of the hue variable are missing, the label will show up in the legend but the artist patch will be blank:

shoyer

shoyer:
Hmm, I'm still not really happy with that, though I suppose it's better than not plotting the missing labels at all.

Is there a reason why this solution would not generalize? It seems like simply calling the plot method with empty data still creates a marker.
```python
plt.scatter([0], [1], c='blue', label='x')
plt.scatter([0], [0], c='red', label='y')
plt.scatter([], [],  c='green', label='z')
plt.legend()
```
image

mwaskom:
I don't think that would always work (for example, it fails with `plt.hist`, and I do not think many seaborn functions are robust to empty data). It would also be a more involved change that would require altering the logic of how the various plotting methods on the objects that inherit from `Grid` work.

mwaskom:
This also, in some ways, is more consistent with what you'll get when you have a missing level in a boxplot or barplot or what have you, where the label will show up on the categorical axis but there will just be empty space in the plot.

I suppose one alternative would be to use a general purpose "missing" glyph so at least you get the color, like an X, maybe.

mwaskom:
(I'm not averse to finding a better solution, but I did want to merge this as it's an acceptable fix for a problematic bug).

shoyer:
> (I'm not averse to finding a better solution, but I did want to merge this as it's an acceptable fix for a problematic bug).

:)

PR #434, opened 2015-01-28
Fixed inconsistent plurals ('sticks'->'stick'). 

ahjulstad:
Also added documention default setting ('box')


mwaskom:
Thanks, as you'll note in the code, passing either will work because it just checks what the string starts with. But good to be consistent, nice catch! Generally the numpy docstring rules are not to state default values in the `Parameters` section because they're redundant with the function signature, and provide an opportunity for the docs to get out of sync if that changes. Also the addition of that clause is causing the linting tests to fail as the line is too long.

ahjulstad:
OK. Removed the default statement and rebased.

(Really small change, but I like practicing my git skills on simple stuff first...)

mwaskom:
Looks great, Thanks!

PR #438, opened 2015-02-05
Changed what qualifies as a 'numeric' type in PairGrid

betatim:
Instead of classing everything which can be converted
to a float as 'numeric' we now check the dtype.kind
attribute.

int, unsigned int, float and complex are numeric, but
bool has stopped being classed as numeric type.

Try out the following:
```
df = pd.DataFrame(dict(A=np.arange(3),
                       B=np.random.randn(3),
                       C=['foo','bar','bah'],
                       D=pd.Timestamp('20130101'),
                       E=[False, True, False]
                       ))
for (name,col) in df.iteritems():
    print name, col.dtype.kind,
    try:
        col.astype(np.float)
        print "numeric"
    except:
        print "not numeric"
```

With the old definition this prints:
```
A i numeric
B f numeric
C O not numeric
D M not numeric
E b numeric```
```

I am not sure if a bool should be a numeric or not. I propose this change as I often use bools to specify a category and then have to specify all columns by hand because `sns.pairplot(dataframe, hue="boolean_col")` classes the column boolean_col as numeric and will an extra row/col to the pair grid for that variable.

Maybe the better fix would be to remove the variable used for `hue` from the list of columns used in the pair plot?

I can not decide if I think a boolean should be numeric or not. In the case of `pairplot()` it seems you would not expect it to be "numeric" though.

Thoughts?

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1082,15 +1082,13 @@ class PairGrid(Grid):
 
     def _find_numeric_cols(self, data):
         """Find which variables in a DataFrame are numeric."""
-        # This can't be the best way to do this, but  I do not
-        # know what the best way might be, so this seems ok
+        # This marks int, unsigned int, float and complex
+        # as numeric types. Booleans are not numeric anymore
         numeric_cols = []
         for col in data:
-            try:
-                data[col].astype(np.float)
+            if col.dtype.kind in "iufc":
                 numeric_cols.append(col)
-            except (ValueError, TypeError):
-                pass
+                
         return numeric_cols
 
 
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1085,9 +1085,9 @@ class PairGrid(Grid):
         # This marks int, unsigned int, float and complex
         # as numeric types. Booleans are not numeric anymore
         numeric_cols = []
-        for col in data:
-            if col.dtype.kind in "iufc":
-                numeric_cols.append(col)
+        for name,col_type in zip(data.columns, data.dtypes):
+            if col_type.kind in "iufc":
+                numeric_cols.append(name)
                 
         return numeric_cols
 


betatim:
Realised that:

* I forgot to commit everything
* there is a smarter way to do this

Locally not all tests pass, which is bad, however I can not tell whether it is because of my change or something else. Any ideas how to find that out?

mwaskom:
Looks like failures are because of the linter.

What are the advantages to this approach over the way it was being done?

PS some tests would be good.

mwaskom:
Oops, wrong button.

betatim:
The original motivation for doing it like this is that you can tell the difference between bool and float. Which gives you the chance to say `bool` that isn't a numeric type.

Having thought about it a bit more though I am starting to think maybe the real solution is to remove what ever column you are using for the `hue` parameter from the list of columns which are plotted in the grid. It seems like you'd not want that. In which case you don't need to change anything in `find_numeric_cols()` logic.

mwaskom:
It's not clear that doing so would always be ideal, though. e.g., this seems useful:

pairplot

betatim:
Ok, maybe too stuck in my "gotta build a classifier, let's use class as the
hue" mind set.

mwaskom:
I mean I see how it could go either way. But allowing the user to provide just the columns that are desired seems more straightforward than dropping columns that are possible to numerically encode (and providing some additional parameter to override this behavior, lest you make it impossible to do something people might want).

Though even in the classification case, having a scatterplot with the class variable doesn't seem totally useless, e.g. it's a good way to check for perfect separation, which is a problem with some models.

pairplot

Could be nice to drop non-numeric columns even if they appear in the `vars` list so that it's easier to pass `df.columns.drop("i_don't_want_this_one")` than to build a list of the ones you do want.

betatim:
The more I think about it the more I agree with you. How frequent is the
use case of wanting to compare all variables except for the one used as
hue? If it is common then it might be worth adding a parameter for that,
instead of having to do `df.colums.drop("my_hue")`.

From my side it might be best to close this PR and discuss it further on
the issue tracker.

Issue #437, opened 2015-02-02
Specifying height/width of {column/row}_colors in cluster map for non-rectangular fig sizes


jrderuiter:
As far as I can see it is not (yet) possible to specify the height/width of side colors in clustermap. If the matrix is far from rectangular, this seems to lead to overly large/small side colors on the axes. Would it be possible to change the height/width of the side colors (possibly using an extra argument) to account for these situations. 

As a side note, it would be very nice if the side colors could also be labeled, any plans to accommodate such a feature?

![screen shot 2015-02-02 at 16 25 53](https://cloud.githubusercontent.com/assets/307739/6003010/69cc8d9e-aaf8-11e4-9905-3ffe137fb3c5.png)

mwaskom:
cc @olgabot

mwaskom:
A related point is that there's a lot of dead space in the figure when one of the dendrograms is turned off.

olgabot:
The sizing of the figure for non-square datasets isn't ideal, I agree. An argument to specify the fraction of the figure taken up by the side colors would be good, especially when there's complex colors as you have for the column labelings. 

As for the labeling of the side colors, were you thinking of something like a legend? Or an additional colorbar?

@mwaskom - agreed. Your comment that a generic side-color labeled heatmap with or without clustering, such as combining `heatmap` and `clustermap` together makes a lot of sense.

I'm currently bogged down with papers and my upcoming advancement so I can't guarantee to dedicate a lot of time to this.

mwaskom:
I'm not sure it needs to be configurable, I think ideally the side colors should just have the same relative "thickness".

jrderuiter:
I agree that more sensible 'defaults' would be great, it might not have to be exactly configurable. Specifying a fraction would however be flexible enough for most cases. 

Labeling might be best in a legend, as plotting the labels directly next to the colors (in case of the column colors for example) might make the labels too small to be legible.

mwaskom:
I am skeptical that it will be easy to find a good general-purpose approach to labeling, though see this [stackoverflow question](http://stackoverflow.com/questions/27988846/how-to-express-classes-on-the-axis-of-a-heatmap-in-seaborn) where I show one way it can be implemented by the user.

olgabot:
Wow that is a really great way! I like the use of the phantom barplot.

jrderuiter:
[...]

PR #441, opened 2015-02-09
Fix clustermap annotation

jrderuiter:
I added some changes to the generation of the clustermap axes, which now avoids excessive whitespace if any of the two dendrograms are omitted. Additionally, I added to extra parameters to the clustermap function, col_colors_ratio and row_colors_ratio, which essentially sets a relative size for the side_color axes in the generation of the overall grid. I also added a small update to include annotation from the index of a passed dataframe containing side_colors for rows and/or columns.

See issue https://github.com/mwaskom/seaborn/issues/437 for more details.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -923,10 +923,13 @@ def clustermap(data, pivot_kws=None, method='average', metric='euclidean',
     {row,col}_linkage : numpy.array, optional
         Precomputed linkage matrix for the rows or columns. See
         scipy.cluster.hierarchy.linkage for specific formats.
-    {row,col}_colors : list-like, optional
+    {row,col}_colors : list-like or DataFrame, optional
         List of colors to label for either the rows or columns. Useful to
         evaluate whether samples within a group are clustered together. Can
-        use nested lists for multiple color levels of labeling.
+        use nested lists for multiple color levels of labeling. Colors can
+        also be provided as a pandas DataFrame, of which each column specifies
+        an entry for the corresponding side color. The column labels of the
+        DataFrame are used to label the colors.
     {row,col}_colors_ratio: float, optional
         A float value indicating the relative proportion of the axes that
         should be used for plotting the {row,col}_colors.
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -505,7 +505,9 @@ def dendrogram(data, linkage=None, axis=1, label=True, metric='euclidean',
 
 class ClusterGrid(Grid):
     def __init__(self, data, pivot_kws=None, z_score=None, standard_scale=None,
-                 figsize=None, row_colors=None, col_colors=None):
+                 figsize=None, row_colors=None, col_colors=None,
+                 row_cluster=True, col_cluster=True, 
+                 row_colors_ratio=0.05, col_colors_ratio=0.05):
         """Grid object for organizing clustered heatmap input on to axes"""
 
         if isinstance(data, pd.DataFrame):
@@ -528,39 +530,19 @@ class ClusterGrid(Grid):
             col_colors = _convert_colors(col_colors)
         self.col_colors = col_colors
 
-        width_ratios = self.dim_ratios(self.row_colors,
-                                       figsize=figsize,
-                                       axis=1)
+        # Setup axes for current configuration.
+        axes = self.setup_axes(self.fig, figsize, 
+                               row_colors, col_colors,
+                               row_cluster, col_cluster,
+                               col_colors_ratio=col_colors_ratio,
+                               row_colors_ratio=row_colors_ratio)
 
-        height_ratios = self.dim_ratios(self.col_colors,
-                                        figsize=figsize,
-                                        axis=0)
-        nrows = 3 if self.col_colors is None else 4
-        ncols = 3 if self.row_colors is None else 4
-
-        self.gs = gridspec.GridSpec(nrows, ncols, wspace=0.01, hspace=0.01,
-                                    width_ratios=width_ratios,
-                                    height_ratios=height_ratios)
-
-        self.ax_row_dendrogram = self.fig.add_subplot(self.gs[nrows - 1, 0:2],
-                                                      axisbg="white")
-        self.ax_col_dendrogram = self.fig.add_subplot(self.gs[0:2, ncols - 1],
-                                                      axisbg="white")
-
-        self.ax_row_colors = None
-        self.ax_col_colors = None
-
-        if self.row_colors is not None:
-            self.ax_row_colors = self.fig.add_subplot(
-                self.gs[nrows - 1, ncols - 2])
-        if self.col_colors is not None:
-            self.ax_col_colors = self.fig.add_subplot(
-                self.gs[nrows - 2, ncols - 1])
-
-        self.ax_heatmap = self.fig.add_subplot(self.gs[nrows - 1, ncols - 1])
+        self.ax_row_dendrogram, self.ax_col_dendrogram, \
+        self.ax_row_colors, self.ax_col_colors, \
+        self.ax_heatmap, self.cax = axes
 
         # colorbar for scale to left corner
-        self.cax = self.fig.add_subplot(self.gs[0, 0])
+        # self.cax = self.fig.add_subplot(self.gs[0, 0])
 
         self.dendrogram_row = None
         self.dendrogram_col = None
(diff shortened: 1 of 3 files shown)


jrderuiter:
I fixed the lint errors, however the builds on Python 2 are still failing due to lint errors in other modules. Anything I can do about this? (Aside from changing the other modules, which is outside the scope of this pull-request).

jrderuiter:
Just added two tests that should cover the added code. The only lines missed in coverage of matrix.py on my machine are related to fastcluster, which I do not have installed. I also added the description of the added ratio parameters to the clustermap function.

mwaskom:
Huh, looks like an update to pep8.py got a lot more strict about some things.

For the time being I guess we could force an install of pep8 1.5 so that the linting rules are at least consistent.

mwaskom:
Oops I didn't see that you had put the pep8 fix in yourself and just merged it into master. No worries, I'll do the merge manually after I've had some time to review and test out this code. Olga, if you want to give it a spin, that would be helpful too.

jrderuiter:
Any update on this?

mwaskom:
Hi Julian,

Playing with this a little bit, building on the clustermap gallery example, I notice some things:

- The colorbar has moved by default and now often overlaps the row labels, e.g. just taking the gallery example literally it obscures them. I guess this is because if column clustering is turned off it would have nowhere to go. (Also I think I've said on github that I'd like it to be possible to put the colorbar over there). But it will need a little more work to position it so that overlaps won't happen. Also, aesthetically, I'm not crazy about it taking up the full vertical expanse of the heatmap.

- The labels for row colors look off when using a DataFrame, I think they go in the wrong order.

- In general I have some issues with the size and aspect ratio of the various plot elements, not sure if they're introduced by this PR but might be good to take care of them while we're touching that code. For example, this seems like a weird way to set up the figure:

```python
sns.clustermap(df.corr().loc["16"], method="average",
               row_colors=colors.loc["16"],
               col_colors=colors,
               cmap=cmap, figsize=(15, 4))
```
wonky_aspect

- More specifically related to this PR, it seems weird that the "height" of the side heatmaps depends so much on the size and aspect ratio of the overall figure. With your additions those will be configurable, but it still seems like it will require a fair amount of trial and error and tweaking to, e.g., get the height the same for the row and column heatmaps in that example. It seems like it would be ideal to make that easier for the user, but I'm not exactly sure what wold be best. Does anything come to mind?

- I think the automatic figsize should reflect whether there is clustering and keep the aspect ratio of the main heatmap the same.

- It's great that `setup_axes` got refactored into its own method, I think it would be good to get some more tests that exercise it directly.

Also:

- Passing colors in a dataframe is not as robust as passing them as a Series. i.e., building on the gallery example, `network_colors` ends up as a Series of lists, I guess, which works fine when passing that directly to `{row,col}_colors`, but if you put that Series into a DataFrame an error gets raised from out code:

```python
b, _ , r = sns.color_palette("deep", 3)
hemi_colors = df.columns.get_level_values("hemi").map(lambda  h: {"lh": b, "rh": r}[h])

colors = pd.DataFrame(dict(hemi=hemi_colors,
                           network=network_colors))
sns.clustermap(df.corr(), row_colors=colors, col_colors=colors, cmap=cmap)
```
```
<ipython-input-58-3fa8192a7ea5> in <module>()
      2                row_colors=colors,
      3                col_colors=colors,
----> 4                cmap=cmap)

/Users/mwaskom/Code/seaborn/seaborn/matrix.py in clustermap(data, pivot_kws, method, metric, z_score, standard_scale, figsize, cbar_kws, row_cluster, col_cluster, row_linkage, col_linkage, row_colors, col_colors, row_colors_ratio, col_colors_ratio, mask, **kwargs)
    969                         row_linkage=row_linkage, col_linkage=col_linkage,
    970                         mask=mask,
--> 971                         **kwargs)

/Users/mwaskom/Code/seaborn/seaborn/matrix.py in plot(self, metric, method, colorbar_kws, row_cluster, col_cluster, row_linkage, col_linkage, mask, **kws)
    875             yind = np.arange(self.data2d.shape[0])
    876 
--> 877         self.plot_colors(xind, yind, **kws)
    878         self.plot_matrix(colorbar_kws, mask, xind, yind, **kws)
    879         return self

/Users/mwaskom/Code/seaborn/seaborn/matrix.py in plot_colors(self, xind, yind, **kws)
    829         if self.row_colors is not None:
    830             matrix, cmap = self.color_list_to_matrix_and_cmap(
--> 831                 self.row_colors, yind, axis=0)
    832             heatmap(matrix, cmap=cmap, cbar=False, ax=self.ax_row_colors,
    833                     xticklabels=False, yticklabels=False,

/Users/mwaskom/Code/seaborn/seaborn/matrix.py in color_list_to_matrix_and_cmap(colors, ind, axis)
    763         if isinstance(colors, pd.DataFrame):
    764             colors = colors.T.values
--> 765             all_colors = set(itertools.chain(*colors))
    766             n, m = colors.shape
    767         elif any(issubclass(type(x), list) for x in colors):

TypeError: unhashable type: 'list'
```

jrderuiter:
Just pushed some changes that should address your issue with the colors as a series of lists and the wrong label order in the row color annotation.

I agree with you that the placing of the colorbar is currently not ideal. However, placing the colorbar on the right does solve any issues with dropping dendrograms and annotations, as there will always be space on the right (disregarding labels of course). A possible solution would entail more accurate positioning of the colorbar, which somehow takes the length of the xlabels into account. Would that be the preferred solution?

Concerning the size of the annotations, I think it would be best to calculate default ratios (based on the aspect ratio (and figure size?)) if none are supplied by the user, but use the given ratios if supplied as arguments. This would prevent having to fiddle upfront, whilst still allowing manual adjustment if required (which I think is useful). Ideally the defaults should give entries in both row/column colours the same size.

mwaskom:
Sorry to be absent from this, I've had a very busy week.

> Just pushed some changes that should address your issue with the colors as a series of lists and the wrong label order in the row color annotation.

Great. Just a note that when that kind of behavior slips through, it's a good indication that the tests are insufficient. So it would be good to add a test that would have caught the bad color labeling, so that future changes don't reintroduce the issue.

> I agree with you that the placing of the colorbar is currently not ideal. However, placing the colorbar on the right does solve any issues with dropping dendrograms and annotations, as there will always be space on the right (disregarding labels of course). A possible solution would entail more accurate positioning of the colorbar, which somehow takes the length of the xlabels into account. Would that be the preferred solution?

Yep, I can see why you moved it, but putting it on top of the row labels is not that much better of a solution. I think the answer is going to have to be something like:

- Set up the grid of axes
- Draw the plot
- Change the figure size so that there will be space for the labels and colorbar
- Move the colorbar axes.

I'm not sure I see any other way to get it to work.

It might also be helpful to add a parameter that will put the colorbar either off to the side vertically or at the bottom horizontally. If I'm making a plot with relatively few rows, where I'm going to have labels, and a lot of columns, where I'm not going to have labels, it would be a more efficient use of space, to put the colorbar on the bottom of the plot. Though it will take some hard thinking to get all that logic to work.

I also don't like huge colorbars and think it should be limited in how much of the axes it can take up on the "quantitative" axis, but that's an easy fix.

> Concerning the size of the annotations, I think it would be best to calculate default ratios (based on the aspect ratio (and figure size?)) if none are supplied by the user, but use the given ratios if supplied as arguments. This would prevent having to fiddle upfront, whilst still allowing manual adjustment if required (which I think is useful). Ideally the defaults should give entries in both row/column colours the same size.

Yep I think this is good. I'd make the default keyword value `"auto"` unless you can think of something that will more clearly communicate how that logic is going to work.

jrderuiter:
> Great. Just a note that when that kind of behavior slips through, it's a good indication that the tests are insufficient. So it would be good to add a test that would have caught the bad color labeling, so that future changes don't reintroduce the issue.

I've indeed been meaning to add some extra tests to cover these cases, so that will happen in the near future.

I have some ideas about the sizes/aspect ratio of the figure and the figure elements. I'll experiment a little bit and push an update if everything seems to be working ok.

sauliusl:
What is the status of this? 
I'd like to '+1' this feature to get merged.

seaborn/matrix.py
```diff
@@ -522,45 +533,22 @@ def __init__(self, data, pivot_kws=None, z_score=None, standard_scale=None,
...
-                self.gs[nrows - 1, ncols - 2])
-        if self.col_colors is not None:
-            self.ax_col_colors = self.fig.add_subplot(
-                self.gs[nrows - 2, ncols - 1])
-
-        self.ax_heatmap = self.fig.add_subplot(self.gs[nrows - 1, ncols - 1])
-
-        # colorbar for scale to left corner
-        self.cax = self.fig.add_subplot(self.gs[0, 0])
+        self.ax_row_dendrogram, self.ax_col_dendrogram, \
```
sauliusl: I personally would think the `self.ax_row_dendrogram` (and similar) assignments should be done inside the `self.setup_axes()` function: `setup_axes` name kind of implies this, and if we ever wanted to change something, we would only need to change it in one place. The code will look cleaner too.

seaborn/matrix.py
```diff
@@ -656,29 +644,93 @@ def standard_scale(data2d, axis=1):
         else:
             return standardized.T
 
-    def dim_ratios(self, side_colors, axis, figsize, side_colors_ratio=0.05):
-        """Get the proportions of the figure taken up by each axes
+    def setup_axes(self, fig, figsize,
```
sauliusl: How about making this protected (i.e. `_setup_axes`) ?

seaborn/matrix.py
```diff
@@ -656,29 +644,93 @@ def standard_scale(data2d, axis=1):
...
+        # Setup heatmap axis.
+        ax_heatmap = fig.add_subplot(gs[nrows - 1, ncols - 3])
+
+        # Setup colorbar axis.
+        ax_cbar = fig.add_subplot(gs[nrows - 1, ncols - 1])
+
+        return ax_row_dendrogram, ax_col_dendrogram, \
+            ax_row_colors, ax_col_colors, ax_heatmap, ax_cbar
+
+    def _ax_index_dendrogram(self, side_colors, cluster):
```
sauliusl: This could be local to `setup_axes()`

jrderuiter:
I got a bit stuck on the sizes/ratio issue and have not had enough time recently to look into it unfortunately. I'll take another stab at it this weekend as I should have some time then!

mwaskom:
There's been a couple of related issues recently that might be good to address while you're thinking hard about this part of the code, but no worries if they feel like they'd extend the scope of the work beyond what you have time for right now.

jrderuiter:
Which related issues are you specifically referring to? If they are related it hopefully wouldn't be too much extra work to take along.

mwaskom:
https://github.com/mwaskom/seaborn/issues/483 is probably most closely related, but I think would be somewhat annoying to fix. https://github.com/mwaskom/seaborn/issues/482 and https://github.com/mwaskom/seaborn/issues/471 should probably be addressed together and might make sense to do here.

johanneskoester:
Hi,
any updates on this? Maybe the current state could be merged, because I don't think it has more issues than the version that is right now provided by seaborn, and it is more configurable.

Thanks,
Johannes

jrderuiter:
Sorry, I haven't had much time to look into this recently. I would be interested in merging these changes into Seaborn, as they are definitely useful (at least for me) for creating heatmaps with annotated row/col colors.

The problem I had was

ChelseaCHENX:
I found it works to adjust the colorbar subplot, eg
```
g = sns.clustermap(df_p,  method='ward', metric='euclidean', cmap='coolwarm', figsize=(8,4),
                      cbar_kws={"shrink": 0.5})
cbar = g.cax
cbar.set_position([0.125, 0.1, 0.1, 0.88])
```

PR #446, opened 2015-02-11
Fix broken reference in dataset_exploration.ipynb

moraxy:
Replaced `:link:` with `:ref:` in [tutorial/dataset_exploration](http://web.stanford.edu/~mwaskom/software/seaborn/tutorial/dataset_exploration.html)

Old | New
-----|-------
Before | After


mwaskom:
Thanks!

PR #454, opened 2015-02-18
WIP: Add docs, examples re: mappable functions

phobson:
Hey -- here's that documentation about writing custom mappable functions. Curious what you think about a) its content, b) its placement in the existing `FacetGrid` notebook (well, I guess that's everything).

Also, you're working some weird voodoo on the notebooks with the `In[null]` and whatnot. I need to figure that out.

### DIFF
--- a/doc/tutorial/axis_grids.ipynb
+++ b/doc/tutorial/axis_grids.ipynb
@@ -1,7 +1,7 @@
 {
  "metadata": {
   "name": "",
-  "signature": "sha256:adea9f4bacdb1010758c941798de7ec13f983a0878e29b9c71ddaa84f0de0888"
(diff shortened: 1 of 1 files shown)


mwaskom:
Cool!

> Also, you're working some weird voodoo on the notebooks with the In[null] and whatnot. I need to figure that out.

I think that's just what happens when you do `make notebooks`.

mwaskom:
Thanks @phobson! I made a few changes, mostly to simplify the example functions and build up the examples a bit more incrementally.

I also couldn't figure out how to trigger an error that motivated the point about popping `color`, etc. from the kwargs, so I took that part out as I thought it was confusing. Let me know if you can reproduce that though and I can add it in with a more specific example.

PR #436, opened 2015-01-31
WIP: add gridspec_kws to FacetGrid

phobson:
Just a quick idea to add a `gridspec_kws` to FactGrid.

Needs:
- [ ] testing
- [ ] docstring
- [ ] documentation

PR #462, opened 2015-03-05
Add gridspec_kws to FacetGrid

phobson:
I force-pushed #436 and now I can't reopen that issue. 

Anyway, I thought about this some more and here's a stab at handling old versions of MPL and instances when `col_wrap` is used. the TODO list still applies:

- [x] testing
- [x] docstring
- [x] documentation
- [x] refine the warning messages
- [x] clear up PEP8/linting stuff

Here's a rough example:
```python
from io import StringIO

import pandas
import seaborn


datafile = StringIO("""\
location,date,parameter,matrix,units,res
SHP-99-31C,2007-10-17 00:00:00,Iron,WG,ug/L,44000.0
SHP-99-31C,2007-10-17 00:00:00,Arsenic,WG,ug/L,292.1
SHP-99-31C,2009-10-01 00:00:00,Oxidation Reduction Potential,WG,mV,-84.0
SHP-99-31C,2009-10-01 00:00:00,pH,WG,SU,6.57
SHP-99-31C,2009-10-01 00:00:00,Arsenic,WG,ug/L,223.5
SHP-99-31C,2009-10-01 00:00:00,Iron,WG,ug/L,26000.0
SHP-99-31C,2010-10-13 00:00:00,Arsenic,WG,ug/L,239.4
SHP-99-31C,2010-10-13 00:00:00,Iron,WG,ug/L,22000.0
SHP-99-31C,2010-10-13 00:00:00,pH,WG,SU,6.46
SHP-99-31C,2010-10-13 00:00:00,Oxidation Reduction Potential,WG,mV,-80.0
SHP-99-31C,2011-10-05 00:00:00,Iron,WG,ug/L,22000.0
SHP-99-31C,2011-10-05 00:00:00,Arsenic,WG,ug/L,244.0
SHP-99-31C,2011-10-05 00:00:00,Oxidation Reduction Potential,WG,mV,-59.2
SHP-99-31C,2011-10-05 00:00:00,pH,WG,SU,6.5
SHP-99-31C,2012-10-18 00:00:00,Arsenic,WG,ug/L,206.4
SHP-99-31C,2012-10-18 00:00:00,Iron,WG,ug/L,17000.0
SHP-99-31C,2012-10-18 00:00:00,pH,WG,SU,6.75
SHP-99-31C,2012-10-18 00:00:00,Oxidation Reduction Potential,WG,mV,-117.1
SHP-99-31C,2013-10-23 00:00:00,Oxidation Reduction Potential,WG,mV,-95.7
SHP-99-31C,2013-10-23 00:00:00,pH,WG,SU,6.7
""")
df = pandas.read_csv(datafile, parse_dates=['date'])
seaborn.set_style('ticks')
fg = seaborn.FacetGrid(
    data=df, row='units', row_order=['ug/L', 'mV', 'SU'],
    hue='parameter', hue_kws={"marker": ["o", "s", '^', 'd']},
    sharey=False, size=2, aspect=4, margin_titles=True,
    gridspec_kws={'height_ratios': [3.5, 1, 1]}  # <---- right here
)
fg.map(plt.plot, 'date', 'res', linestyle='--', marker='o', clip_on=False)
fg.axes.flat[0].set_yscale('log')
fg.axes.flat[-1].set_ylim(bottom=4.75, top=9)
```
test

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1,5 +1,8 @@
 from __future__ import division
 from itertools import product
+from distutils.version import LooseVersion
+import warnings
+
 import numpy as np
 import pandas as pd
 import matplotlib as mpl
@@ -174,7 +177,8 @@ class FacetGrid(Grid):
                  sharex=True, sharey=True, size=3, aspect=1, palette=None,
                  row_order=None, col_order=None, hue_order=None, hue_kws=None,
                  dropna=True, legend_out=True, despine=True,
-                 margin_titles=False, xlim=None, ylim=None, subplot_kws=None):
+                 margin_titles=False, xlim=None, ylim=None, subplot_kws=None,
+                 gridspec_kws=None):
         """Initialize the plot figure and FacetGrid object.
 
         Parameters
@@ -185,7 +189,7 @@ class FacetGrid(Grid):
         row, col, hue : strings, optional
             Variable (column) names to subset the data for the facets.
         col_wrap : int, optional
-            Wrap the column variable at this width. Incompatible with `row`.
+            Wrap the column variable at this width. Incompatible with ``row``.
         share{x, y}: booleans, optional
             Lock the limits of the vertical andn horizontal axes across the
             facets.
@@ -215,7 +219,12 @@ class FacetGrid(Grid):
         {x, y}lim: tuples, optional
             Limits for each of the axes on each facet when share{x, y} is True.
         subplot_kws : dict, optional
-            Dictionary of keyword arguments.
+            Dictionary of keyword arguments passed to matplotlib subplot(s)
+            methods.
+        gridspec_kws : dict, optional
+            Dictionary of keyword arguments passed to matplotlib's ``gridspec``
+            module (via ``plt.subplots``). Requires matplotlib >= 1.4 and is
+            ignored if ``col_wrap`` is not ``None``.
 
         Returns
         -------
@@ -229,6 +238,10 @@ class FacetGrid(Grid):
         factorplot : Combines pointplot, barplot, or boxplot and a FacetGrid
 
         """
+
+        MPL_GRIDSPEC_VERSION = LooseVersion('1.4')
+        OLD_MPL = LooseVersion(mpl.__version__) < MPL_GRIDSPEC_VERSION
+
         # Compute the grid shape
         ncol = 1 if col is None else len(data[col].unique())
         nrow = 1 if row is None else len(data[row].unique())
@@ -251,6 +264,7 @@ class FacetGrid(Grid):
 
         # Build the subplot keyword dictionary
         subplot_kws = {} if subplot_kws is None else subplot_kws.copy()
+        gridspec_kws = {} if gridspec_kws is None else gridspec_kws.copy()
         if xlim is not None:
             subplot_kws["xlim"] = xlim
         if ylim is not None:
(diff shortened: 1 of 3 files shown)


phobson:
I think the new IPython is breaking some of your testing infrastructure:
https://travis-ci.org/phobson/seaborn/jobs/53240853#L751

mwaskom:
Hm, can you force an install of IPython 2.x in the travs build until I can sort that out?

phobson:
Trying that now. 



—

seaborn/axisgrid.py
```diff
@@ -258,14 +263,24 @@ def __init__(self, data, row=None, col=None, hue=None, col_wrap=None,
...
-            fig, axes = plt.subplots(nrow, ncol, figsize=figsize,
-                                     squeeze=False,
-                                     sharex=sharex, sharey=sharey,
-                                     subplot_kw=subplot_kws)
+            kwargs = dict(figsize=figsize, squeeze=False,
+                          sharex=sharex, sharey=sharey,
+                          subplot_kw=subplot_kws,
+                          gridspec_kw=gridspec_kws)
+            mpl_version = utils.check_mpl_version()
+            if mpl_version < MPL_GRIDSPEC_VERSION:
```
mwaskom: Should this use `diutils.LooseVersion` for robustness?

seaborn/axisgrid.py
```diff
@@ -258,14 +263,24 @@ def __init__(self, data, row=None, col=None, hue=None, col_wrap=None,
...
+                msg = "gridspec module only available in mpl >= {}"
+                warnings.warn(msg.format(MPL_GRIDSPEC_VERSION))
+                _ = kwargs.pop('gridspec_kw')
+
+            fig, axes = plt.subplots(nrow, ncol, **kwargs)
             self.axes = axes
 
         else:
             # If wrapping the col variable we need to make the grid ourselves
+            if len(gridspec_kws) > 0:
```
mwaskom: `if gridspec_kws` is more idiomatic, right?

seaborn/utils.py
```diff
@@ -414,3 +414,9 @@ def axes_ticklabels_overlap(ax):
     """
     return (axis_ticklabels_overlap(ax.get_xticklabels()),
             axis_ticklabels_overlap(ax.get_yticklabels()))
+
+
+def check_mpl_version():
+    import matplotlib as mpl
+    major, minor, point = tuple(map(float, mpl.__version__.split('.')))
```
mwaskom: Ah, as above I would just use distutils.LooseVersion here, then the comparisons will just work.

seaborn/utils.py
```diff
@@ -414,3 +414,9 @@ def axes_ticklabels_overlap(ax):
     """
     return (axis_ticklabels_overlap(ax.get_xticklabels()),
             axis_ticklabels_overlap(ax.get_yticklabels()))
+
+
+def check_mpl_version():
+    import matplotlib as mpl
```
mwaskom: Any reason this import is delayed? matplotlib is a mandatory dependency :)

seaborn/utils.py
phobson: Ahh -- I didn't even know about that. Thank for the tip.

seaborn/axisgrid.py
phobson: Yup. I'll fix it.

seaborn/utils.py
mwaskom: It's good because it knows about things like `.dev`, `.rc1` as minor version, etc, which otherwise would, I think, break here.

mwaskom:
Here's an example that works nicely with the existing datasets:

```python
titanic = sns.load_dataset("titanic")
titanic = titanic.sort("deck")
g = sns.FacetGrid(titanic, col="class", sharex=False,
                 gridspec_kws={"width_ratios": [5, 3, 3]})
g.map(sns.boxplot, "deck", "age")
```

titanic_gridspec

mwaskom:
I'd drop that example into the tutorial docs right after the "The size of the figure is set by providing the height of the facets and the aspect ratio" cell.

seaborn/tests/test_axisgrid.py
```diff
@@ -432,7 +459,7 @@ def test_set_axis_labels(self):
         npt.assert_array_equal(got_x, xlab)
         npt.assert_array_equal(got_y, ylab)
 
-    def test_subplot_kws(self):
+    def test_axis_lims(self):
```
phobson: I think this test was previously misnamed -- ok to rename it with this PR?

mwaskom: The name predates `subplot_kws` as an actual argument, but yes your change is more clear. :+1:

phobson:
Not sure. Failures were due to that warnings about the MPL version were getting raised too soon (before  `gridspec_kws` was even used), so the default warning filter was swallowing the legitimate warnings.

Fixed that by resetting the warnings, which might be hacky (I'm honestly not sure what the preferred  testing protocol is there).

Latest tests pass. I suggest we let this marinate a bit and then I'll squash everything.

phobson:
Oh hell I got an errant file mixed up in all this. I'll clean that up tonight. **do not merge**

mwaskom:
Sounds good, I'll hold off.

I think it's good to have the warnings but not a *huge* deal if they're not tested, so no need to hack together a complicated infrastructure about it.

phobson:
Ok, reverted that errant file (the fact that tests still passed bodes well for another PR in the works).

I think this is good to go for whatever additional review you think it needs.

mwaskom:
Thanks @phobson !

PR #463, opened 2015-03-05
Use IPython 2.X for testing

phobson:
There seems to be problems with IPython 3 and the testing workflow right now.

https://github.com/mwaskom/seaborn/pull/462#issuecomment-77439356


mwaskom:
Thanks!

Issue #448, opened 2015-02-11
factorplot/barplot's x_order can't be a pandas Series


mbatchkarov:
When plotting the following data frame without an explicit `x_order` argument things look great.

```
df = pd.DataFrame({'name': list('abcde'), 'score' : [3,2,1,5,4]})
sns.factorplot(x='name', y='score', data=df, kind='bar');
```

If I wanted to order the bars by height, the labels of the x axis in the following bar chart are wrong (should be `c b a e d`, because `c` has a value of `1` in the data frame and a value of `3` in the figure).

```
order = df.name[df.score.argsort()]
sns.barplot(x=df.name, y=df.score, x_order=order)
```

![wrong](https://cloud.githubusercontent.com/assets/937114/6153634/e525e20c-b21d-11e4-987c-17db81a39c75.png)

If I convert the label to a python `list` or a numpy `ndarray` things also look correct:

```
order = df.name[df.score.argsort()].tolist()
sns.factorplot(x='name', y='score', data=df, kind='bar', x_order=order);
```

```
order = df.name[df.score.argsort()].values
sns.factorplot(x='name', y='score', data=df, kind='bar', x_order=order);
```

The same happens if I use `sns.barplot` directly. The docs says `x_order` should be list-like, and I think `pd.Series` falls in that category.

mwaskom:
This code is currently getting overhauled to use the same underlying implementation as the new boxplot/violinplot code so I think this issue will disappear (but keep an eye out for the pull request and bug me there if that isn't true).

mwaskom:
Closing #448 via #466

Issue #445, opened 2015-02-11
Strange errors message for invalid input to the new boxplot and violinplot


shoyer:
I'm running the dev version of Seaborn to test out the boxplot/violinplot refactor.

I think this functions could use more input validation -- they aren't giving sensible error message when the `y` or `hue` arguments are invalid. For example:
```python
tips = sns.load_dataset("tips")
sns.boxplot("day", "tip", "sexy", tips)
```
```
KeyError: 'Thur'
```

I'm also seeing some even stranger error messages that I think have the same source, though I haven't been able to reproduce them using a dataset I can share:
```python
sns.violinplot('x', 'not a real var', data=df.dropna())
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-169-2fbc318d67a2> in <module>()
----> 1 sns.boxplot('x', 'not a real var', data=df.dropna())

/Users/shoyer/dev/seaborn/seaborn/distributions.pyc in boxplot(x, y, hue, data, order, hue_order, orient, color, palette, saturation, width, fliersize, linewidth, whis, notch, ax, **kwargs)
   1161     plotter = _BoxPlotter(x, y, hue, data, order, hue_order,
   1162                           orient, color, palette, saturation,
-> 1163                           width, fliersize, linewidth)
   1164 
   1165     if ax is None:

/Users/shoyer/dev/seaborn/seaborn/distributions.pyc in __init__(self, x, y, hue, data, order, hue_order, orient, color, palette, saturation, width, fliersize, linewidth)
     30                  width, fliersize, linewidth):
     31 
---> 32         self.establish_variables(x, y, hue, data, orient, order, hue_order)
     33         self.establish_colors(color, palette, saturation)
     34 

/Users/shoyer/dev/seaborn/seaborn/distributions.pyc in establish_variables(self, x, y, hue, data, orient, order, hue_order)
    202 
    203                 # Group the numeric data
--> 204                 grouped_vals = vals.groupby(groups)
    205                 plot_data = [grouped_vals.get_group(g) for g in order]
    206                 plot_data = [d.values for d in plot_data]

/Users/shoyer/dev/pandas/pandas/core/generic.pyc in groupby(self, by, axis, level, as_index, sort, group_keys, squeeze)
   2892         axis = self._get_axis_number(axis)
   2893         return groupby(self, by=by, axis=axis, level=level, as_index=as_index,
[...]

Issue #435, opened 2015-01-28
Violinplot with groups and hue fails displaying legend


ahjulstad:
Violinplot tries to create the legend using the first violin only. If this violin does not have both 'hue' present (with split=True), legend creation fails.
```
df = pd.DataFrame({'a': np.arange(10), 'b': ['f',]*9+['g']*1, 'c': ['x',]*5 + ['y',]*5})
sns.violinplot('c', 'a', hue='b', data=df, split=True)
```
gives
```

  File "c:\users\ahju\documents\kamera\seaborn\seaborn\distributions.py", line 1293, in violinplot
    plotter.plot(ax)

  File "c:\users\ahju\documents\kamera\seaborn\seaborn\distributions.py", line 982, in plot
    self.draw_violins(ax)

  File "c:\users\ahju\documents\kamera\seaborn\seaborn\distributions.py", line 764, in draw_violins
    self.add_legend_data(ax, support[0], 0,

IndexError: index 0 is out of bounds for axis 0 with size 0

```

mwaskom:
Closing #435 via #466

Issue #425, opened 2015-01-19
Enhance factorplot with distribution plots


mwaskom:
Previously `kind="box"` worked in `factorplot`. That's been hidden for the merge of #410, but I want to do a larger overhaul that brings in violin/stripplots as well, and that allows you to pass a list of kinds to be composed.

mwaskom:
Closing #425 via #466

PR #466, opened 2015-03-09
Unify categorical plots

mwaskom:
This is a follow-on to #410. The majority of the changes are in terms of the implementation of things, but there are a few important new features and changes.

In short, the categorical distribution plots (`boxplot`, `violinplot` and `stripplot`) have been unified with the categorical estimation plots (`barplot`, `pointplot`, and `countplot`). From the user perspective, that means they all share a basic API. From the developer perspective, it means that as many common operations were abstracted out into the the underlying `_CategoricalPlotter` and `_CategoricalStatPlotter` classes. That means adding additional plot kinds in the future should be easier and more robust.

## Demonstration of major new features

The main highlight in terms of new functionality is that `barplot` and `pointplot` can now draw horizontally:

```python
sns.barplot(x="tip", y="day", data=tips)
```
seaborn-barplot-4

Additionally, `factorplot` can use any of the categorical plot kinds:

```python
sns.factorplot(x="age", y="embark_town", hue="sex", row="class",
               data=titanic[titanic.embark_town.notnull()],
               size=2, aspect=3.5, palette="Set3",
               kind="violin", split=True, cut=0, bw=.2)
```

seaborn-factorplot-6

## API changes and new features

I will update this PR later with a catalog of other changes.

- The major API change in `{bar,point,factor}_plot` is that the `x_order` parameter has been renamed to `order` to reflect the fact that the categorical variable can be on the x or y axis.

- The ability to draw counts by passing only an `x` variable to `barplot` has been removed, but the new `countplot` function has been added to retain this functionality.

- In all places within the categorical plots, the representation and order of categorical levels is inferred from the data with an attempt to use pandas Categorical information if it exists. Otherwise, the default is now to use the order of levels as they appear in the dataframe rather than sorting by default. However, this behavior has not *yet* been changed in `FacetGrid`. That will likely be a different PR.

- Two parameters have been removed from the estimation plots (with attempts to catch and warn): `hline` and `dropna`

- This PR adds a compatibility layer for some of the API breaks in #410 

- Added the `scale` parameter to pointplot, which scales the point size and linewidths by a single factor.

- Added the `errcolor` parameter to `barplot` and also now pass extra `**kwargs` down to `plt.bar`.

- Added the `ax` property to `FacetGrid`, which will give access to the main Axes in the case where there are no `row` or `col` facets.

- Added a more informative exception when string variables are named that don't appear in `data`.

- Fixed how the proxy artist used for generating the `violinplot` and `boxplot` legends is drawn to avoid errors when the first category has no data for a hue level.

- There are longer narrative explanations in the docstrings for all of the categorical plotting functions and copious examples for that will be executed and shown in the online API docs.

## Linked issues:

This PR addresses several issues: #361, #416, #425, #435, #445, #448 

## Outstanding issues

- [x] Update the release notes
- [x] Update and reorganize the API docs homepage
- [x] `countplot` needs some docstring examples
- [ ] Update the tutorials (may get punted on)


mwaskom:
It would be great if people could pull this branch and throw all their weird, real-life data at it to check corner cases I may have missed with the test data/my own generally pretty clean experimental data.

phobson:
Just want to say that this + the fact that factor plot takes `row|col|hue` arguments is 

# Huge

phobson:
What's your the timeline for this? I can maybe take a crack at the tutorials (separate branch & PR) while testing. But that'll definitely have to wait for later in the week spilling over into next week.

mwaskom:
The tutorial changes I have in mind are broader than just this PR -- now that I have an infrastructure for examples in the API docs, I want to move much of the nitty gritty stuff to there and make the tutorials more high-level and integrative. But I don't quite have a vision for what that will look like yet, so I'm not gonna hold up this PR for it.

mwaskom:
Something I am considering is changing `hue=` in the categorical plots to `by=`, or possibly something different, to better differentiate it from what `FacetGrid` is doing with `hue`, which in almost all cases (except `pointplot`) is different.

seaborn/utils.py
```diff
@@ -414,3 +418,37 @@ def axes_ticklabels_overlap(ax):
...
+        from the ``values`` object.
+
+    Returns
+    -------
+    order : list
+        Ordered list of category levels
+
+    """
+    if order is None:
+        if hasattr(values, "categories"):
```
JanSchulz: Just fore reference: this only supports pandas >=0.15, in former versions `Categorical.levels` have to be used...

PR #474, opened 2015-03-17
More flexible tsplot

ebothmann:
Hi there!

By these small modifications, `tsplot` covers a wider range of applications:
- an error estimator function can be passed (the default is `utils.ci`, as before)
- bootstrapping can be disabled by passing `n_boot = 0`

Cheers,
Eno


mwaskom:

PR #492, opened 2015-03-26
Plot joint plot with Gaussian fits to histograms

pwolfram:
This is a minor enhancement that added ability to plot a joint plot with Gaussians fitted to each of the x and y histograms.  This is helpful in assessing normalcy of the spread of a particle cloud.

A trivially simple example can be found at 

http://nbviewer.ipython.org/gist/pwolfram/d0250d84aec03de1045d

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -481,7 +481,7 @@ def jointplot(x, y, data=None, kind="scatter", stat_func=stats.pearsonr,
         Data or names of variables in `data`.
     data : DataFrame, optional
         DataFrame when `x` and `y` are variable names.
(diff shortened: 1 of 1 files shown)


mwaskom:
It's not clear that this add much over adding `marginal_kws={"fit": stats.norm}` to the call (or using `JointGrid.plot_marginals` directly. It's also somewhat of a deviation in that the other kinds apply to the representation of the joint distribution, while this is just a variant on the scatter kind with a different marginal plot. I originally expected it to fit a multivariate normal or something.

pwolfram:
Good call.  I concur.  It seems that fitting a multivariate normal would be more useful and worthy of a pull request.  However, it probably shouldn't be implemented within the joint plot framework.

PR #496, opened 2015-04-01
Heatmap mouseover

### LINKED ISSUE #495: corrplot does not show cell coordinates with nbagg backend
I was trying the gallery plot from [here](http://stanford.edu/~mwaskom/software/seaborn/examples/many_pairwise_correlations.html)
and noticed that while using the nbagg backend, that the x and y displayer just stays empty, so not even the cursor coordinate is shown.

The jointplot from [here](http://stanford.edu/~mwaskom/software/seaborn/examples/joint_kde.html) works fine, so it seems that there's is no general nbagg problem.

IPython version: 3.0.0
Python 3.4
Seaborn: 0.5.1

tacaswell:
Closes #495 for real.

### DIFF
--- a/seaborn/linearmodels.py
+++ b/seaborn/linearmodels.py
@@ -1051,26 +1051,26 @@ def symmatplot(mat, p_mat=None, names=None, cmap="Greys", cmap_range=None,
         ynames = names if annot else names[1:]
         ax.set_yticklabels(ynames)
 
-    _xfun = mouseover_factory(xnames)
-    _yfun = mouseover_factory(ynames)
+        _xfun = mouseover_factory(xnames)
+        _yfun = mouseover_factory(ynames)
 
-    def _zfun(x, y):
-        shp = mat.shape
-        x, y = x+.5, y+.5
-        if 0 < x < shp[1]-.5 and 0 < y < shp[0]:
-            return str(mat[int(y), int(x)])
-        else:
-            return ''
+        def _zfun(x, y):
+            shp = mat.shape
+            x, y = x+.5, y+.5
+            if 0 < x < shp[1]-.5 and 0 < y < shp[0]:
+                return str(mat[int(y), int(x)])
+            else:
+                return ''
 
-    def mouse_over(x, y):
-        if y < x:
-            return ''
-        x_str = _xfun(x)
-        y_str = _yfun(y)
-        z_str = _zfun(x, y)
-        return "x={} y={} z={}".format(x_str, y_str, z_str)
+        def mouse_over(x, y):
+            if y < x:
+                return ''
+            x_str = _xfun(x)
+            y_str = _yfun(y)
+            z_str = _zfun(x, y)
+            return "x={} y={} z={}".format(x_str, y_str, z_str)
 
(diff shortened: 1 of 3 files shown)


mwaskom:
This isn't something that can be handled on your end? Seaborn is likely not the only place people are going to want to use categorical labels on the axes of a `pcolormesh` plot...

michaelaye:
This works for me but could the implementation of showing the z value here not clash with https://github.com/matplotlib/matplotlib/pull/3989 ?

tacaswell:
mpl does not capture enough semantic information to be able to reliably tell when the user is plotting categorical data or not.  That knowledge is captured at the seaborn level (by which functions are being used) so this formatting should live downstream.  Eventually the z-data should move upstream, but reporting the x and y position as labels should not.

@michaelaye If you are using a version of mpl with matplotlib/matplotlib#3989 you will get this plus the artist specific string on the end.  The brains to modulate if the `z` string is included should probably be in seaborn based on what version of mpl it is built on top of.  

I thought about just doing the x and y values, but figured that you would also want the z value _now_ rather than some time this summer :wink: .

mwaskom:
> mpl does not capture enough semantic information to be able to reliably tell when the user is plotting categorical data or not. That knowledge is captured at the seaborn level (by which functions are being used) so this formatting should live downstream. Eventually the z-data should move upstream, but reporting the x and y position as labels should not.

Perhaps not automatically, but as an option in `pcolormesh` (`xcellnames`, `ycellnames`, e.g.) or a more general categorical axes object that the mouse stuff knows how to do lookups on. Each seem like they would be useful. I am unenthusiastic about adding a lot of random shim code all over seaborn to handle matplotlib interaction stuff that I don't use and am not at all knowledgeable about.

mwaskom:
Or alternatively there is already the `pcolormesh(X, Y, C)` signature, which could be generalized to treat the coordinate vectors as categorical when they cannot be converted to numerics.

tacaswell:
Similarly, I am unenthusiastic about mpl absorbing lots of shims to deal with domain-specific assumptions. I view `seaborn` as the domain specific code base where those details live.

I am some what surprised you don't use any of the interaction.  That is leaving a tremendous amount of the mpl's power on the table.

mwaskom:
Then I suppose we are at an impasse.

mdboom:
I'll apologize for my hasty solution in matplotlib/matplotlib#4302 that perhaps "gave hope" to this being a simple bug.  I think @tacaswell's descriptions make it clear why this isn't so easy and that, at least currently, seaborn has the semantic knowledge to do this in the right way but matplotlib doesn't.

To give some context, matplotlib I think historically has been burned by "assuming too much" in some of its interfaces, and while we can make changes on the matplotlib end to handle this other third-party tools may not agree.

I would be nice to step back and find a way forward here...  Perhaps as a way of comparing apples to apples, we already have a sense of what this looks like to solve it on the seaborn end in this PR.  What would this look like to solve on the matplotlib end -- and how, specifically, would it limit matplotlib's API flexibility in the future?  I think an implementation might clarify those points.

Issue #375, opened 2014-11-20
What to do with missing values in heatmaps


sdvillal:
Running this snippet:

```python
import seaborn as sns
import numpy as np
import pandas as pd

df = pd.DataFrame(data={'a': [1, 1, 1],
                        'b': [2, np.nan, 2],
                        'c': [3, 3, np.nan]})

sns.heatmap(df)
```

We get something like this:

![missing-get-small](https://cloud.githubusercontent.com/assets/642252/5127852/43a01d08-70d7-11e4-8798-930b3e1763ed.png)

Missing values are happily assigned "the color of the minimum" in the color bar. I wonder if this is a good default or if missing values should be treated differently. I would go for "treating them differently by default" and then the question would be how (maybe having a *missing_color* parameter and issuing a warning if the color is in the colormap/colorbar).

jm-contreras:
+1

JWarmenhoven:
+1 
I agree. Masking works fine and should perhaps be default for missings. There is indeed an issue with masking when used in combination with annotations as jm-contreras mentioned.

mwaskom:
Having looked into this a bit, the issue seems to be that `plt.pcolormesh` doesn't treat `np.nan` as "missing", and instead assigns those cells the "under" value in the colormap. E.g.:

```python
cmap = sns.cubehelix_palette(as_cmap=True, light=.9)
sns.heatmap(flights_missing, cmap=cmap)
```
![pwz xybmyopqlbsupejin2a3wcfmallix4rdkmsuogvuyrjxctelklsfzgxs5luruzskir1ero7jeldxmqusvix x8cv eomwcuqwaaaabjru5erkjggg](https://cloud.githubusercontent.com/assets/315810/7334558/d56fecc4-eb4b-11e4-98ac-ab6a05942a11.png)

```python
cmap.set_under(".5")
sns.heatmap(flights_missing, cmap=cmap)
```
![kz5akpwicj4jak8 jvl5cfqxhv7ruanyc7gncz lvqsmozz7flxsgilqooyt ymc 3utmskmtflnwzingf dxwozo6tocxypckqytysuus1evm7jikdretuyrjxctelklsfzgxs5luruzskir1kf8h 1lnqm15jhcaaaaasuvork5cyii](https://cloud.githubusercontent.com/assets/315810/7334561/e7910cc6-eb4b-11e4-984c-16de18bb3c21.png)

[...]

PR #497, opened 2015-04-02
Skip annotating masked values in heatmaps

JWarmenhoven:
There is an issue with heatmap annotations when when there are masked values in the data. Formatting fails for these masked values. 
Discussed in #375.

http://nbviewer.ipython.org/github/JWarmenhoven/ipython_notebooks/blob/master/Seaborn_heatmap_issue.ipynb

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -148,11 +148,12 @@ class _HeatMapper(object):
         xpos, ypos = np.meshgrid(ax.get_xticks(), ax.get_yticks())
         for x, y, val, color in zip(xpos.flat, ypos.flat,
                                     mesh.get_array(), mesh.get_facecolors()):
-            _, l, _ = colorsys.rgb_to_hls(*color[:3])
-            text_color = ".15" if l > .5 else "w"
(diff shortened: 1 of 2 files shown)


mwaskom:
Looks good thanks! Do you think you can add a test? Basically just add a function right below this one: https://github.com/JWarmenhoven/seaborn/blob/patch-1/seaborn/tests/test_matrix.py#L203 that is similar but tests the absence of annotations for masked values?

JWarmenhoven:
I am not really sure how I should go about.

mwaskom:
Well a) we want to make sure we exercise the code when heatmap is used with both a mask and annotations, to make sure it doesn't crash (as I believe it did before). Then I would at least check that the number of text objects in the figure is the same as the number of non-masked entries in the array. Best of course would be making sure that the values are the same (as in the linked test).

JWarmenhoven:
Well...that was interesting. I have a thing or two to learn about this whole workflow.

mwaskom:
Looks perfect, thanks!

PR #500, opened 2015-04-07
Allow combining logx with other regplot fitting options.

anntzer:
A quick attempt at allowing the combination of the `logx` option with other `regplot` fitting options (`logistic`, `lowess` and `robust`).


mwaskom:
Thanks, this will need tests, though.

mwaskom:
Closing as stalled (though also does not feel like the best way to approach the ideal behavior).

PR #507, opened 2015-04-09
Don't crash tsplot on TimeSeries of length <2.

### LINKED ISSUE #506: Empty argument to tsplot leads to UnboundLocalError
Arguably a not-so-important edge case, so I'm just mentioning it:
```
$ python -c 'from seaborn import tsplot; tsplot([])'
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.4/site-packages/seaborn/timeseries.py", line 255, in tsplot
    ax.set_xlim(x.min(), x.max())
UnboundLocalError: local variable 'x' referenced before assignment
```
(seaborn 0.5.1)

anntzer:
Fixes #506.

### DIFF
--- a/seaborn/timeseries.py
+++ b/seaborn/timeseries.py
@@ -252,10 +252,12 @@ def tsplot(data, time=None, unit=None, condition=None, value=None,
         ax.plot(x, central_data, color=color, label=label, **kwargs)
 
     # Pad the sides of the plot only when not interpolating
-    ax.set_xlim(x.min(), x.max())
-    x_diff = x[1] - x[0]
-    if not interpolate:
-        ax.set_xlim(x.min() - x_diff, x.max() + x_diff)
+    if len(data):
+        if interpolate or len(data) == 1:
+            ax.set_xlim(x.min(), x.max())
+        else:
+            x_diff = x[1] - x[0]
+            ax.set_xlim(x.min() - x_diff, x.max() + x_diff)
 
     # Add the plot labels
     if xlabel is not None:


mwaskom:
I am wary to merge any changes to `tsplot` as it currently does not have any tests and so there is no way to know if this is going to break things. Basically the function needs an entire refactor before it can be improved.

anntzer:
I would not call this an "improvement"; I just wish that `tsplot` does not randomly crash my program when given some problematic input.  Would you accept the patch if I make the fix really obvious (though ugly), e.g.
```
try:
    x
except NameError:
    pass
else:
    if interpolate or len(x) == 1:
        ax.set_xlim(x.min(), x.max())
    else:
        x_diff = x[1] - x[0]
        ax.set_xlim(x.min() - x_diff, x.max() + x_diff)
```
The issue is that `x` may not be defined or may not have two elements (this is trivially reproducible) so we just check whether the following code will run without issues...

mwaskom:
> I just wish that tsplot does not randomly crash my program when given some problematic input. 

Maybe I don't fully understand, but it sounds like that logic could be handled in your program? If the input is problematic there's only so much tsplot can do... I totally agree that the errors that eventually get raised are often very *unhelpful*, but It's a function to plot timeseries data, so I'm not sure I'd expect it to work when it gets an input with a single timepoint....

mwaskom:
Anyway the broader point is that `tsplot` has a lot of flexibility in how it can be invoked so even if this change works for your usecase I would be worried that it might cause issues for others. Probably not, but I don't want to totally break things in the service of making it less annoying (but technically functional). Since there are no automated tests, there's no way to be sure right now.

Unfortunately I don't have time right now to really test it out manually, but the first thing to do would be to make sure all the examples in the timeseries tutorial run. This used to happen automatically, but things changed in IPython faster than I could keep up with and I had to disable that with the hope that I would get around to refactoring and bringing the test coverage up to the standards of (most of) the rest of the package, but it hasn't happened yet.

anntzer:
If there is no timepoint, then the obvious thing to do is to do nothing (why could it matter? because I'm using `tsplot` to overlay on top of another plot).  If there is a single timepoint, I agree that the best choice is not so clear, but I'd be inclined to follow matplotlib's approach here, which is to do nothing by default but plot a point if e.g. `marker` is `"o"` (so e.g. if `err_style` is `unit_points`).
Anyways, not a big deal.

mwaskom:
> If there is no timepoint, then the obvious thing to do is to do nothing (why could it matter? because I'm using tsplot to overlay on top of another plot).

I don't fully agree, for two reasons. The first is that seaborn functions are more complicated than matplotlib functions, and including special cases in all the statistics to handle malformed input is likely to be overly complex and prone to issues (whereas directly mapping an empty vector of data to an empty vector of plot elements is more straightforward). The second is that it's really not obvious to me that the best behavior when a function gets bad input is to silently do nothing. In your case that might be fine, because you know and accept what might be causing the "bad" input, but in a lot of other cases, passing 0d data to a timeseries plotting function is a sign that something has gone awry upstream. It seems safer for seaborn to fail in that case, and delegate logic that says "0d datas is fine here" to the caller.

mwaskom:
But like I said, I agree the actual error that gets raised here is garbage.

mwaskom:

PR #522, opened 2015-04-24
Fix kwarg issue with plot_marginals

cyrusmaher:
Current version requires that the user pass `vertical=False` for the plotting to work correctly.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1220,6 +1220,7 @@ class JointGrid(object):
             Returns `self`.
 
         """
+        kwargs["vertical"] = False
         plt.sca(self.ax_marg_x)
         func(self.x, **kwargs)
 


mwaskom:
How did you actually run into this problem? I guess the assumption here is that the `vertical` keyword argument is `False` by default -- are you using a custom function or did you run into this issue with something in seaborn?

cyrusmaher:
I was using a custom function. What would you think about a PR for a
groupby option for jointplot? This would map group to hue across the three
panels (matching up bins between histograms)...

mwaskom:
Got it, ok this looks good.

cyrusmaher:
I would propose to do this through the option: kind="grouped".

mwaskom:
Specifically for `jointplot` you mean?

mwaskom:
This issue has come up a few times and I've never found a satisfying approach. My current thinking is to add `hue` parameters to more of the lower-level functions and let `JointGrid` leverage that rather than including all that logic in `JointGrid` itself (or in `jointplot`, which I think should be as thin a layer around `JointGrid` as possible).

cyrusmaher:
That's a good idea!! How can I help?

PR #524, opened 2015-04-27
Check if mask is empty before attempting rearange

zorro4:
If mask is empty, rearranging fails

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -796,7 +796,8 @@ class ClusterGrid(Grid):
 
     def plot_matrix(self, colorbar_kws, mask, xind, yind, **kws):
         data2d = self.data2d.iloc[yind, xind]
-        mask = np.asarray(mask, np.bool)[yind][:, xind]
+        if mask is not None:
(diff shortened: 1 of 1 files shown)


mwaskom:
Ah I guess this is even more complicated. The `mask` should be converted to a `DataFrame` with the same index and columns as the main data matrix. It might make sense for this to happen elsewhere in the `ClusterGrid` object, like during set up.

mwaskom:
> It might make sense for this to happen elsewhere in the ClusterGrid object, like during set up.

In general I think we should:

- if both `data` and `mask` are DataFrames, check that they have the same index/column values, otherwise raise
- if `mask` is an array, check that it has the same shape as `data`, otherwise raise
- otherwise make sure `data` and `mask` are both dataframes with the same index/column values before plotting.

This should happen early, probably in the constructor (where these kinds of checks usually live). The tricky thing is that this logic really belongs in `heatmap`, but that would be too late for the indexing to work properly in `ClusterGrid`.

mwaskom:
Thanks for catching this, I was able to spend a little time working on #523 and think I've now covered all the options for passing a mask.

PR #526, opened 2015-04-27
NF: get_dataset_names (overdo: test_load_datasets)

yarikoptic:
Was exploring functionality and realized that couldn't figure out what datasets you have available for demos.
I guess ideally you should just provide some .json or whatever file which would list all the .csv with descriptions or may be even proper dataset descriptors if you really like to.

### DIFF
--- a/seaborn/tests/test_utils.py
+++ b/seaborn/tests/test_utils.py
@@ -12,6 +12,14 @@ from nose.tools import assert_equal, raises
 from distutils.version import LooseVersion
 pandas_has_categoricals = LooseVersion(pd.__version__) >= "0.15"
 
+from pandas.util.testing import network
+from ..utils import get_dataset_names, load_dataset
+
+try:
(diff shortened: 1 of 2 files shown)


seaborn/tests/test_utils.py
```diff
@@ -281,3 +289,30 @@ def test_category_order():
 
         out = utils.categorical_order(x, ["b", "a"])
         nt.assert_equal(out, ["b", "a"])
+
+
+@network(url="https://github.com/mwaskom/seaborn-data")
```
mwaskom: Does this decorator allow the test to get skipped when there's no network connection?

yarikoptic: Yeap... that is its purpose + also tag the test with network tag ask you could skip it altogether (on the phone atm, so man nosetests for arg, smth like -X "not network")

mwaskom:
Thanks!

Issue #536, opened 2015-05-03
boxplot overwrites user-provided flierprops and sym kwargs


cgevans:
If a call to boxplot has kwargs with `flierprops` or `sym` in it, boxplot ignores these and overwrites them. This prevents the modification of flier styles: see [here](http://stackoverflow.com/questions/29647145/setting-flier-outlier-style-in-seaborn-boxplot-is-ignored/30009353#30009353) for example.

It seems like a reasonable solution to this would be to only set sym if sym isn't already provided, and to set flierprops by updating the base dict in boxplot with the user-provided dict. This would be relatively easy to implement, and I can do so soon.

mwaskom:
Yep, sure, that seems reasonable.

mwaskom:
I vaguely remember there being some issue with `flierprops` being broken in matplotlib, but I think that might have been early in the 1.4 series and thus not too worrisome. But good to keep in mind.

cgevans:
I believe that was the [annoying bug in 1.4.0](https://github.com/matplotlib/matplotlib/issues/3459/) that was fixed in 1.4.1.

mwaskom:
Closed with #538

PR #537, opened 2015-05-03
Modified boxplot to take into account flierprops and other props kwargs.

cgevans:
Fixes at least the props side of #536, by updating artists' properties after restyling if the props kwargs are there. Thus even if a props kwarg is present, it will only override seaborn's styling if that's actually what the user intends.

This does not fix the `sym` kwarg. The best way I could see to do so would involve importing `_process_plot_format` from `matplotlib.axes._base`, and I'm not sure if it's worth it, as flyerprops allows the same settings to be made in a more reliable way.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -449,7 +449,7 @@ class _BoxPlotter(_CategoricalPlotter):
                                          widths=self.width,
                                          **kws)
                 color = self.colors[i]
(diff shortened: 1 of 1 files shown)


cgevans:
I forgot to make a branch for this, and the change runs into [pep8 issue 231](https://github.com/jcrocholl/pep8/issues/231), so I'll close this and make a better version.

PR #538, opened 2015-05-03
Modified boxplot to take into account flierprops and other props kwargs. 

cgevans:
Fixes at least the props side of #536, by updating artists' properties after restyling if the props kwargs are there. Thus even if a props kwarg is present, it will only override seaborn's styling if that's actually what the user intends.

This does not fix the sym kwarg. The best way I could see to do so would involve importing _process_plot_format from matplotlib.axes._base, and I'm not sure if it's worth it, as flyerprops allows the same settings to be made in a more reliable way.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -449,7 +449,7 @@ class _BoxPlotter(_CategoricalPlotter):
                                          widths=self.width,
                                          **kws)
                 color = self.colors[i]
-                self.restyle_boxplot(artist_dict, color)
+                self.restyle_boxplot(artist_dict, color, kws)
             else:
                 # Draw nested groups of boxes
                 offsets = self.hue_offsets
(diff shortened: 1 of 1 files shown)


mwaskom:
Looks like this actually does fix the "sym" issue, more or less:

```python
sns.boxplot(data=tips, flierprops={"marker": "o"})
```
ngnrnzgaaaabjru5erkjggg

mwaskom:
Thanks!

cgevans:
Ah, by sym issue I meant that `sym=` itself won't work. One of the reasons I don't find this too concerning is that `flierprops={"marker": "o"}` and similar allow the same and more. `sym=` really seems to just be there for convenience in matplotlib.

phobson:
very cool @cgevans. thanks for this.

SilasK:
Hello, 
For me the solution works for cycles but not for other markers such as "+" or "_". Is there a way to implement this?

phobson:
@SilasK you need to set the `markeredgewidth` property to something > 0

Issue #545, opened 2015-05-09
RuntimeError: Cannot get window extent w/o renderer


tundo91:
I'm trying to plot a *seaborn* *clustermap* (it doesn't work with the *heatmap* too) with the following, no NaNs admitted:

```python
    import numpy as np
    import pandas as pd
    import seaborn as sns
    import matplotlib.pyplot as plt
    
    def plotClusterMap():
        a = pd.DataFrame(np.matrix('1 2; 3 4'))
        print a
        fig = plt.figure()
        sns.clustermap(a)
        plt.show()
```

```a``` is well formed:

       0  1
    0  1  2
    1  3  4

Console output:

    Traceback (most recent call last):
      File "main.py", line 78, in <module>
        main()
      File "main.py", line 72, in main
        heatmapPlotter.plotClusterMap()
      File "/Users/username/code.py", line 12, in plotClusterMap
        sns.clustermap(a)
      File "/Library/Python/2.7/site-packages/seaborn/matrix.py", line 895, in clustermap
        **kwargs)
      File "/Library/Python/2.7/site-packages/seaborn/matrix.py", line 813, in plot
        self.plot_matrix(colorbar_kws, mask, **kws)
      File "/Library/Python/2.7/site-packages/seaborn/matrix.py", line 803, in plot_matrix
        cbar_kws=colorbar_kws, mask=mask, **kws)
      File "/Library/Python/2.7/site-packages/seaborn/matrix.py", line 292, in heatmap
        plotter.plot(ax, cbar_ax, kwargs)
      File "/Library/Python/2.7/site-packages/seaborn/matrix.py", line 177, in plot
        if axis_ticklabels_overlap(xtl):
      File "/Library/Python/2.7/site-packages/seaborn/utils.py", line 374, in axis_ticklabels_overlap
        bboxes = [l.get_window_extent() for l in labels]
      File "/usr/local/lib/python2.7/site-packages/matplotlib/text.py", line 796, in get_window_extent
        raise RuntimeError('Cannot get window extent w/o renderer')
    RuntimeError: Cannot get window extent w/o renderer

* Mac OSX: 10.10.3
* Seaborn: 0.5.1
* Matplotlib: 1.3.1 (just downgraded from 1.4)

tundo91:
Python 3.4.2 and numpy 1.9.2 don't solve the error :(

tundo91:
[...]

Issue #408, opened 2014-12-24
seaborn install error (RuntimeError: Cannot get window extent w/o renderer)


deeplearnx:
I've be struggling to get seaborn running all night. I've installed it using (ana)conda. When running this example (http://stanford.edu/~mwaskom/software/seaborn/examples/network_correlations.html) I get the following. I haven't been able to find any solutions. I've tried reinstalling Anaconda as well as installing matplotlib and seaborn separately to no avail...

Traceback (most recent call last):
  File "heatmap.py", line 9, in <module>
    sns.heatmap(corrmat, vmax=.8, linewidths=0, square=True)
  File "/Users/ssingh/anaconda/lib/python2.7/site-packages/seaborn/matrix.py", line 292, in heatmap
    plotter.plot(ax, cbar_ax, kwargs)
  File "/Users/ssingh/anaconda/lib/python2.7/site-packages/seaborn/matrix.py", line 177, in plot
    if axis_ticklabels_overlap(xtl):
  File "/Users/ssingh/anaconda/lib/python2.7/site-packages/seaborn/utils.py", line 374, in axis_ticklabels_overlap
    bboxes = [l.get_window_extent() for l in labels]
  File "/Users/ssingh/anaconda/lib/python2.7/site-packages/matplotlib/text.py", line 739, in get_window_extent
    raise RuntimeError('Cannot get window extent w/o renderer')
RuntimeError: Cannot get window extent w/o renderer

cel4:
Sorry to jump in here. There seems to be a problem with the `MacOSX` backend. I have seen it on SO http://stackoverflow.com/questions/27868865/runtimeerror-cannot-get-window-extent-w-o-renderer-with-seaborn?noredirect=1#comment44142319_27868865 and can reproduce it.
Switching the backend is a workaround. I don't understand this issue, but I think it needs fixing. Maybe we should cross-link it to `matplotlib`?

mwaskom:
I guess it could catch the error and just set the ticklabels outward so they are guaranteed not to overlap, but that seems kinda lame.

cel4:
Still weird, that it works with other packages, but not with `MacOSX`. Seems suspicious.

mwaskom:
Seems similar (though not exactly the same error) as #231. I guess I just have a lower bar for surprise at matplotlib backend jankiness than you :)

cel4:
This could be related as well, not quite sure: https://github.com/matplotlib/matplotlib/issues/2654

[...]

Issue #231, opened 2014-07-03
Known matplotlib issue with tight_layout, "RuntimeError: CGContextRef is NULL", on the first tutorial example


jona-sassenhagen:
When running the tutorial on Mac OS X, the very first example won't work (see copy paste below).

It's a known issue in matplotlib:
https://github.com/matplotlib/matplotlib/issues/2654
Is it possible to either fix it, or at least warn Mac users about it? Might lose some potential customers otherwise.
It works when using another backend.

Code:

```
Python 3.4.1 (v3.4.1:c0e311e010fc, May 18 2014, 00:54:21) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> import pandas as pd
>>> import seaborn as sns
>>> import matplotlib as mpl
>>> import matplotlib.pyplot as plt
>>> np.random.seed(sum(map(ord, "linear_quantitative")))
>>> tips = sns.load_dataset("tips")
>>> sns.lmplot("total_bill", "tip", tips);
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/seaborn/linearmodels.py", line 724, in lmplot
    legend=legend, legend_out=legend_out)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/seaborn/axisgrid.py", line 215, in __init__
    fig.tight_layout()
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/matplotlib/figure.py", line 1606, in tight_layout
    rect=rect)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/matplotlib/tight_layout.py", line 352, in get_tight_layout_figure
    pad=pad, h_pad=h_pad, w_pad=w_pad)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/matplotlib/tight_layout.py", line 129, in auto_adjust_subplotpars
    tight_bbox_raw = union([ax.get_tightbbox(renderer) for ax in subplots])
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/matplotlib/tight_layout.py", line 129, in <listcomp>
[...]

PR #546, opened 2015-05-09
Two fixes to avoid matplotlib osx backend issues

mwaskom:
This changes some things in `FacetGrid` and `heatmap` to avoid two common and very annoying issues caused by the the matplotlib osx backend.

I believe this is the root matplotlib issue: https://github.com/matplotlib/matplotlib/issues/2654

This has come up a few times in seaborn: #545, #408, #231, and elsewhere.

The fix to `FacetGrid` should, I believe, simply make that work. The fix to `heatmap` is a compromise where the exception that is raised when trying to determine if the ticklabels overlap gets caught and the labels are not rotated. That means that the plots might not look as good for everyone, but this feels better than the plot simply not working.

I haven't tested this extensively, but I seem to be able to plot using the OSX backend without running into these common issues.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -137,3 +137,4 @@ Utility functions
     set_hls_values
     ci_to_errsize
     axlabel
+    load_dataset
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -4,11 +4,12 @@ Installing and getting started
 ------------------------------
 
 To install the released version of seaborn, you can use ``pip`` (i.e. ``pip install seaborn``). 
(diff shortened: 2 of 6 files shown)


mwaskom:
Unfortunately this causes issues with the `Grid.add_legend` method. Not sure what the best workaround is.

mwaskom:
Hm, also there is some weird issue such that the legend does not immediately appear when the gui window pops up, but it does show up if the window is resized.

mwaskom:
Closing in favor of #611

PR #553, opened 2015-05-11
add lowess bootstrap/ci functionality

### LINKED ISSUE #552: regplot: support confidence intervals with lowess model
Current `regplot` behavior when `lowess=True` is to ignore the confidence interval (`ci`) and bootstrap (`n_boot`) keyword arguments (example below):

``` python
sns.regplot('obs', 'mod', data=data, lowess=True, ci=95, n_boot=1000)
```

Any thoughts on making it possible to bootstrap the lowess model?

jhamman:
I gave #552 a quick try.  Comments, suggestions, and critiques are welcome.  

Fitting the `lowess` model is pretty slow when `n_boots` is large but I guess that is expected.  

[Example ipython notebook](http://nbviewer.ipython.org/urls/dl.dropbox.com/s/22n3ksv2cny7pd9/seaborn_bootstrap_lowess_example.ipynb/%3Fdl%3D0)

Fixes #552

### DIFF
--- a/seaborn/linearmodels.py
+++ b/seaborn/linearmodels.py
@@ -195,8 +195,7 @@ class _RegressionPlotter(_LinearPlotter):
             yhat, yhat_boots = self.fit_statsmodels(grid, GLM,
                                                     family=Binomial())
         elif self.lowess:
-            ci = None
-            grid, yhat = self.fit_lowess()
+            grid, yhat, yhat_boots = self.fit_lowess()
         elif self.robust:
             from statsmodels.robust.robust_linear_model import RLM
             yhat, yhat_boots = self.fit_statsmodels(grid, RLM)
@@ -255,8 +254,20 @@ class _RegressionPlotter(_LinearPlotter):
     def fit_lowess(self):
         """Fit a locally-weighted regression, which returns its own grid."""
         from statsmodels.nonparametric.smoothers_lowess import lowess
-        grid, yhat = lowess(self.y, self.x).T
-        return grid, yhat
+
+        def _wrap_lowess(X, y, return_grid=False):
+            grid, yhat = lowess(y, X).T
+            if not return_grid:
+                return yhat
+            else:
+                return grid, yhat
+
+        grid, yhat = _wrap_lowess(self.x, self.y, True)
+
+        yhat_boots = algo.bootstrap(self.x, self.y, func=_wrap_lowess,
+                                    n_boot=self.n_boot, units=self.units)
+
+        return grid, yhat, yhat_boots
 
     def fit_logx(self, grid):
         """Fit the model in log-space."""
--- a/seaborn/tests/test_linearmodels.py
+++ b/seaborn/tests/test_linearmodels.py
@@ -213,6 +213,10 @@ class TestRegressionPlotter(object):
         _, boots_smod = p.fit_statsmodels(self.grid, smlm.OLS)
         npt.assert_equal(boots_smod.shape, (self.n_boot, self.grid.size))
 
+        # Lowess
+        grid, _, boots_lowess = p.fit_lowess()
+        npt.assert_equal(boots_lowess.shape, (self.n_boot, grid.size))
+
     @skipif(_no_statsmodels)
     def test_regress_without_bootstrap(self):
 
@@ -368,7 +372,6 @@ class TestRegressionPlotter(object):
         grid, yhat, err_bands = p.fit_regression(x_range=(-3, 3))
 
         nt.assert_equal(len(grid), len(yhat))
-        nt.assert_is(err_bands, None)
 
     def test_regression_options(self):
 


mwaskom:
Yeah I'd originally implemented this but the fact that it ran incredibly slowly in my initial testing made me think that it was best not to use it.

I think newer statsmodels might do analytical confidence intervals for the lowess smoother? Is that true?

jhamman:
I don't see the analytical confidence intervals in [`statsmodels.nonparametric`](http://statsmodels.sourceforge.net/devel/nonparametric.html) but I'm not super familiar with that package.  

I do see that they have converted most of the [lowess module to cython](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/nonparametric/_smoothers_lowess.pyx) so maybe its not as slow as it once was?

Setting `ci=None` or `n_boots=0` retains the previous, faster, functionality.  Any thoughts on how make that the default behavior but still allow for confidence intervals when desired?

mwaskom:
> Setting `ci=None` or `n_boots=0` retains the previous, faster, functionality. Any thoughts on how make that the default behavior but still allow for confidence intervals when desired?

Right, my worry is that it's not obvious to playing with `regplot` that using `lowess=True` is suddenly going to make everyone a LOT slower, and there's no clean way to have not bootstrapping be a default only for the lowess kind.

I think my preference would actually be to have a dedicated `lowessplot` function, which could then not show CIs by default (but make them available) and also expose the other lowess parameters with a good API. This is a much larger undertaking though, and I won't have the bandwidth to supervise it until the summer.

Also I could have sworn I'd seen someone make a statsmodels lowess plot with a non-bootstrapped confidence interval somewhere, but I'm not sure.

jhamman:
I think that all sounds reasonable. Let's pick this back up this summer, for now my changes work for my needs but it would be nice to rework the API a bit.

jhamman:
@mwaskom - I'm just pinging back on this one.  

I haven't seen a non-bootstrap confidence interval for the lowess, have you?

mwaskom:
Have you tried asking on their github/mailing list? Could be buried in a PR or the sandbox.

jhamman:
I asked on the mailing list this afternoon.  I'll let you know what I hear back.  

A few years back they discussed this topic (https://github.com/statsmodels/statsmodels/pull/266).  @josef-pkt had [this comment](https://github.com/statsmodels/statsmodels/pull/266#issuecomment-5694974):
>aside: for some nonparametric methods, bootstrap confidence intervals are recommended, and with a fast lowess this will be much easier to do.

mwaskom:
Is "a fast lowess" what they currently have? I think the current version might be ok for statistical modeling, but it's slowwww for visualization.

jhamman:
Yes, "a fast lowess" is what they currently have.  He was referring to the cython rewrite of the function that is currently in statsmodels.  

I agree that slow is not ideal for visualization.  Are you thinking that if there is not a analytic solution that , you'll close this PR?

jhamman:
You may follow the `statsmodels` mailing list but here is the answer I got from them: https://groups.google.com/forum/#!topic/pystatsmodels/8i4JoJjIsFA

In short, it sounds like it would require some work to make it happen.

mwaskom:
Seems like the trick using `delta` in the slender means post is relevant. statsmodels has it but I gather it's 0 by default? Can you work up some timings using this PR with delta set using the heuristic given in that article to get a sense for how much it would buy us?

jhamman:
`delta` is zero by default. Timings are below and using the heuristic given in the article, we get a bit better than a 2x speedup.

```python
In [1]: from statsmodels.nonparametric.smoothers_lowess import lowess

In [2]: import seaborn as sns

In [3]: tips = sns.load_dataset("tips")

In [4]: delta = (tips.total_bill.max() - tips.total_bill.min()) * 0.01

In [5]: %timeit -n100 lowess(tips.tip, tips.total_bill, delta=0)
100 loops, best of 3: 76.7 ms per loop

In [6]: %timeit -n100 lowess(tips.tip, tips.total_bill, delta=delta)
100 loops, best of 3: 33.3 ms per loop

In [7]: %timeit -n100 lowess(tips.tip, tips.total_bill, delta=delta, frac=0.3)
100 loops, best of 3: 31.8 ms per loop
```

mwaskom:
Hm. 30ms for a single fit still means 30s for 1000 bootstraps (the default), right?

jhamman:
Correct.  The default in [`_RegressionPlotter`](https://github.com/mwaskom/seaborn/blob/master/seaborn/linearmodels.py#L75) for `n_boots` is 1000. 

Just throwing out some ideas here of how to move forward:

1. reduce the default number of bootstrap iterations when `lowess` is being used.
1. add a warning when bootstrapping `lowess` explaining that it is slow.
1. use the implementation in this PR (using the `delta = 0.01*range` approach discussed above), leaving the slow `lowess` bootstrap as the non-default behavior for `regplot`
1. wait for someone to speed up `lowess` in `statsmodels`, close this PR but leave #552 open.

mwaskom:
I don't think I understand the third option.

jhamman:
updated for clarity.

jhamman:
Closing as stale.

Demetrio92:
Looking through the codebase it easy to see how to use the same tools to simply enable this bootstrap functionality. 

```python
import numpy as np
import seaborn as sns
from statsmodels.nonparametric.smoothers_lowess import lowess


def regplot_lowess_ci(data, x, y, ci_level, n_boot, **kwargs):
    x_ = data[x].to_numpy()
    y_ = data[y].to_numpy()
    x_grid = np.linspace(start=x_.min(), stop=x_.max(), num=1000)

    def reg_func(_x, _y):
        return lowess(exog=_x, endog=_y, xvals=x_grid)

    beta_boots = sns.algorithms.bootstrap(
        x_, y_,
        func=reg_func,
        n_boot=n_boot,
    )
    err_bands = sns.utils.ci(beta_boots, ci_level, axis=0)
    y_plt = reg_func(x_, y_)

    ax = sns.lineplot(x=x_grid, y=y_plt, **kwargs)
    sns.scatterplot(x=x_, y=y_, ax=ax, **kwargs)
    ax.fill_between(x_grid, *err_bands, alpha=.15, **kwargs)
    return ax


mpg_df = sns.load_dataset('mpg')
ax = regplot_lowess_ci(mpg_df, x='mpg', y='acceleration', ci_level=99, n_boot=100)
ax.figure.show()
```
image

Demetrio92:
Then, it can also be easily used with `facet_grid` to replicate `lmplot`
```python
mpg_df.eval('heavy = weight>2803', inplace=True)
grid = sns.FacetGrid(mpg_df, col='heavy', sharex=False)
grid.map_dataframe(regplot_lowess_ci, x='mpg', y='acceleration', ci_level=99, n_boot=100)
grid.figure.show()
```
image

Issue #560, opened 2015-05-13
despine with trim=True fails on horizontal categorical plots


mwaskom:
```python
planets = sns.load_dataset("planets")
ax = sns.boxplot(x="orbital_period", y="method",
                 data=planets[planets.orbital_period < 1000])
sns.despine(trim=True)
```
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-103-cd9b605db889> in <module>()
      2 ax = sns.boxplot(x="orbital_period", y="method",
      3                  data=planets[planets.orbital_period < 1000])
----> 4 sns.despine(trim=True)

/Users/mwaskom/Code/seaborn/seaborn/utils.pyc in despine(fig, ax, top, right, left, bottom, offset, trim)
    218 
    219             yticks = ax_i.get_yticks()
--> 220             firsttick = np.compress(yticks >= ax_i.get_ylim()[0], yticks)[0]
    221             lasttick = np.compress(yticks <= ax_i.get_ylim()[-1], yticks)[-1]
    222             ax_i.spines['left'].set_bounds(firsttick, lasttick)

IndexError: index 0 is out of bounds for axis 0 with size 0
```

@phobson I think you originally wrote this code; any idea what's up here?

phobson:
If you take out the despine, you get an error thrown in the `matplotlib.cbook.bxp_stats` functions, which I *also* wrote. Pop into a debugger there and it fails on:

```
[snip]
C:\Miniconda3\envs\cvc\lib\site-packages\matplotlib\cbook.py in boxplot_stats(X, whis, bootstrap, labels)
   1994 
   1995         # arithmetic mean
-> 1996         stats['mean'] = np.mean(x)
   1997 
   1998         # medians and quartiles
[snip]
```

So then:
```
ipdb> print(x)
['Radial Velocity']
```

So I think somewhere along the line, `sns.boxplot` passed the labels and not the data under the label.

phobson:
Should I be debugging in from master?

mwaskom:
> If you take out the despine, you get an error thrown in the matplotlib.cbook.bxp_stats functions, which I also wrote.

Hmm, I can't reproduce *that*:

```python
planets = sns.load_dataset("planets")
ax = sns.boxplot(x="orbital_period", y="method",
                 data=planets[planets.orbital_period < 1000])
```
[...]

PR #561, opened 2015-05-14
Fix bug in boxplot where all observations in some group were missing

mwaskom:
Cf #560 (doesn't close the root issue there though)


mwaskom:
Ah dammit how did those commits end up in here.

PR #562, opened 2015-05-14
Fix bug in boxplot where some level had only missing data

mwaskom:
Cf #560 (don't close the root issue there though)


mwaskom:
@phobson does this branch fix all the weirdness you were seeing in #560 too?

PR #596, opened 2015-06-12
ENH: allow local cacheing of datasets

jakevdp:
This PR introduces a couple extra arguments to the ``load_dataset`` utility; it allows datasets to optionally be cached locally for use without a network connection.
I think the doc string is pretty self explanatory, but let me know if it needs clarification!

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -2,14 +2,13 @@
 from __future__ import print_function, division
 import colorsys
 import warnings
+import os
 
 import numpy as np
 from scipy import stats
 import pandas as pd
 import matplotlib.colors as mplcol
 import matplotlib.pyplot as plt
-
-import os
 
 from distutils.version import LooseVersion
 pandas_has_categoricals = LooseVersion(pd.__version__) >= "0.15"
@@ -393,7 +392,7 @@ def get_data_home(data_home=None):
     return data_home
 
 
-def load_dataset(name, cache=False, data_home=None, **kws):
+def load_dataset(name, cache=True, data_home=None, **kws):
     """Load a dataset from the online repository (requires internet).
 
     Parameters
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -386,14 +386,14 @@ def get_data_home(data_home=None):
     """
     if data_home is None:
         data_home = os.environ.get('SEABORN_DATA',
-                                   os.path.join('~', 'seaborn_data'))
+                                   os.path.join('~', 'seaborn-data'))
     data_home = os.path.expanduser(data_home)
     if not os.path.exists(data_home):
         os.makedirs(data_home)
     return data_home
 
 
-def load_dataset(name, cache=False, data_home=None, quiet=True, **kws):
+def load_dataset(name, cache=False, data_home=None, **kws):
     """Load a dataset from the online repository (requires internet).
 
     Parameters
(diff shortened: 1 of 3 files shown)


jakevdp:
Note we could probably remove the ``quietprint`` stuff; I mostly added it there to confirm the function was working as expected.

mwaskom:
Oh cool, I was hoping someone would come along and do this.

Though see my note in the `seaborn-data` repo about whether the best way forward is a more general purpose PyData repository that includes but is not limited to particular datasets that show off seaborn functionality.

jakevdp:
Responded in https://github.com/mwaskom/seaborn-data/issues/1

Regardless of the long-term solution, I think a cacheing system like this would be nice in the short-term!

seaborn/utils.py
```diff
@@ -380,13 +402,36 @@ def load_dataset(name, **kws):
         Name of the dataset (`name`.csv on
         https://github.com/mwaskom/seaborn-data).  You can obtain list of
         available datasets using :func:`get_dataset_names`
+    cache : boolean, optional
+        If True, then cache data locally and use the cache on subsequent calls
+    data_home : string, optional
+        The directory in which to cache data. By default, uses ~/seaborn_data/
+    quiet : boolean, default
```
mwaskom: Is "default" here a typo or does this mean something different? I've never seen this convention.

jakevdp: Typo – should be "optional"

seaborn/utils.py
```diff
@@ -380,13 +402,36 @@ def load_dataset(name, **kws):
         Name of the dataset (`name`.csv on
         https://github.com/mwaskom/seaborn-data).  You can obtain list of
         available datasets using :func:`get_dataset_names`
+    cache : boolean, optional
+        If True, then cache data locally and use the cache on subsequent calls
+    data_home : string, optional
+        The directory in which to cache data. By default, uses ~/seaborn_data/
```
mwaskom: The only thing I don't like about this kind of cache system is that I hate when my home directory gets cluttered with project-specific data directories. But I guess it's the best place to put it?

mwaskom:
Yep this is great, it will speed up the doctests a lot and will make it easier to hack offline.

seaborn/utils.py
jakevdp: We could make it a hidden file by default...
I chose this because it's similar to how scikit-learn approaches the problem.

mwaskom: No I agree, it's the convention. Just saying I wish the convention were different! :)

I think hidden directory isn't a great answer -- the seaborn datasets aren't currently very large but in general best not to hide large files I think.

jakevdp: Note that the way this is currently structured, you could point ``SEABORN_DATA`` to the ``seaborn-data`` git repository and it would do everything locally without ever doing any data caches!

mwaskom: Oops I missed the environment variable route (well I saw it, I just didn't have it in mind) that's a good solution.

jakevdp: or just clone the repo in your home directory and forget about the environment variable :smile:

jakevdp:
Fixed the typo, and removed the debugging stuff.

One last thing: do you think cacheing should be enabled by default?

mwaskom:
Yep, I think it's more often useful than problematic.

Can you also add a line in `doc/releases/v0.6.0.txt` mentioning this change?

jakevdp:
One last thing: unit tests would be nice, but I don't quite understand the decorator magic you have going on at https://github.com/mwaskom/seaborn/blob/master/seaborn/tests/test_utils.py#L320

Any thoughts on that?

mwaskom:
That's actually @yarikoptic's contribution, though I guess the decorator itself is a pandas function. I think the way it works is that it tries to connect to that URL and raises SkipTest instead of some other exception if it can't get a network connection.

jakevdp:
Test failure seems unrelated to this PR, I believe. https://travis-ci.org/mwaskom/seaborn/jobs/66587092

mwaskom:
Thanks Jake!

PR #606, opened 2015-06-22
End caps on barplot confidence intervals

jat255:
This is something I have been using on my own plots, and I figured it might be of use to others. 

Added a simple switch to `barplot` that allows for perpendicular lines to be drawn on the confidence intervals (which could be nicer for display, depending on your preference).

This changes the default behavior from this:
001-without_perp

To this:
001-with_perp

Since it's a totally optional argument, and I don't think it breaks anything, hopefully it's not too painful to incorporate :smiley:

- Josh

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1244,44 +1244,38 @@ class _CategoricalStatPlotter(_CategoricalPlotter):
                       at_group,
                       confint,
                       colors,
-                      draw_perp=False,
-                      thickness=1.8,
+                      conf_lw=1,
+                      capsize=0,
                       **kws):
 
-        kws.setdefault("lw", mpl.rcParams["lines.linewidth"] * thickness)
+        kws.setdefault("lw", mpl.rcParams["lines.linewidth"] * conf_lw)
 
         for at, (ci_low, ci_high), color in zip(at_group,
                                                 confint,
                                                 colors):
             if self.orient == "v":
                 ax.plot([at, at], [ci_low, ci_high], color=color, **kws)
-                if draw_perp:
-                    ax.plot([at - 0.05, at + 0.05],
-                            [ci_low, ci_low],
-                            color=color,
-                            **kws)
-                    ax.plot([at - 0.05, at + 0.05],
-                            [ci_high, ci_high],
-                            color=color,
-                            **kws)
+                ax.plot([at - capsize / 2, at + capsize / 2],
+                        [ci_low, ci_low], color=color, **kws)
+                ax.plot([at - capsize / 2, at + capsize / 2],
+                        [ci_high, ci_high], color=color, **kws)
+
             else:
                 ax.plot([ci_low, ci_high], [at, at], color=color, **kws)
-                if draw_perp:
-                    ax.plot([ci_low, ci_low],
-                            [at - 0.05, at + 0.05],
-                            color=color,
-                            **kws)
-                    ax.plot([ci_high, ci_high],
-                            [at - 0.05, at + 0.05],
-                            color=color,
-                            **kws)
+                ax.plot([ci_low, ci_low],
+                        [at - capsize / 2, at + capsize / 2],
+                        color=color, **kws)
+                ax.plot([ci_high, ci_high],
+                        [at - capsize / 2, at + capsize / 2],
+                        color=color, **kws)
 
 
 class _BarPlotter(_CategoricalStatPlotter):
     """Show point estimates and confidence intervals with bars."""
     def __init__(self, x, y, hue, data, order, hue_order,
                  estimator, ci, n_boot, units,
-                 orient, color, palette, saturation, errcolor, draw_perp):
+                 orient, color, palette, saturation, errcolor, conf_lw,
+                 capsize):
         """Initialize the plotter."""
         self.establish_variables(x, y, hue, data, orient,
                                  order, hue_order, units)
@@ -1289,7 +1283,8 @@ class _BarPlotter(_CategoricalStatPlotter):
         self.estimate_statistic(estimator, ci, n_boot)
 
         self.errcolor = errcolor
-        self.draw_perp = draw_perp
+        self.conf_lw = conf_lw
+        self.capsize = capsize
 
     def draw_bars(self, ax, kws):
         """Draw the bars onto `ax`."""
@@ -1308,7 +1303,8 @@ class _BarPlotter(_CategoricalStatPlotter):
             self.draw_confints(ax, barpos,
                                self.confint,
                                errcolors,
-                               draw_perp=self.draw_perp)
+                               self.conf_lw,
+                               self.capsize)
 
         else:
 
@@ -1327,7 +1323,8 @@ class _BarPlotter(_CategoricalStatPlotter):
                     self.draw_confints(ax, offpos,
                                        confint,
                                        errcolors,
-                                       draw_perp=self.draw_perp)
+                                       self.conf_lw,
+                                       self.capsize)
 
     def plot(self, ax, bar_kws):
         """Make the plot."""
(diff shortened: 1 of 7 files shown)


mwaskom:
Cool, thanks. This isn't my preferred style but other people like it so they'll be happy.

Test failures are because `countplot` uses `_BarPlotter` (but doesn't draw confidence intervals) so you just need to add that argument.

Code comments coming.

seaborn/categorical.py
```diff
@@ -1230,24 +1230,49 @@ def estimate_statistic(self, estimator, ci, n_boot):
...
                                                self.value_label)
 
-    def draw_confints(self, ax, at_group, confint, colors, **kws):
+    def draw_confints(self,
+                      ax,
+                      at_group,
+                      confint,
+                      colors,
+                      draw_perp=False,
+                      thickness=1.8,
```
mwaskom: It looks like this is being exposed here but not to the user? Is that intentional?

seaborn/categorical.py
```diff
@@ -1230,24 +1230,49 @@ def estimate_statistic(self, estimator, ci, n_boot):
             self.value_label = "{}({})".format(estimator.__name__,
                                                self.value_label)
 
-    def draw_confints(self, ax, at_group, confint, colors, **kws):
+    def draw_confints(self,
+                      ax,
+                      at_group,
+                      confint,
+                      colors,
+                      draw_perp=False,
```
mwaskom: IMO this parameter should be called `capsize` for consistency with matplotlib and should take a scalar value defaulting to 0. Then you can draw lines from `at - (capsize / 2)` to `at + (capsize / 2)`. Does that make sense?

seaborn/categorical.py
```diff
@@ -1270,7 +1296,10 @@ def draw_bars(self, ax, kws):
 
             # Draw the confidence intervals
             errcolors = [self.errcolor] * len(barpos)
-            self.draw_confints(ax, barpos, self.confint, errcolors)
+            self.draw_confints(ax, barpos,
+                               self.confint,
+                               errcolors,
+                               draw_perp=self.draw_perp)
```
mwaskom: Might as well keep things as either all args or all kwargs since this is internal.

seaborn/categorical.py
```diff
@@ -1511,6 +1543,11 @@ def plot(self, ax):
         ``1`` if you want the plot colors to perfectly match the input color
         spec.\
     """),
+    draw_perp=dedent("""\
```
mwaskom: See the `capsize` comment above.

seaborn/categorical.py
```diff
@@ -2155,7 +2192,7 @@ def stripplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
 def barplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
             estimator=np.mean, ci=95, n_boot=1000, units=None,
             orient=None, color=None, palette=None, saturation=.75,
-            errcolor=".26", ax=None, **kwargs):
+            errcolor=".26", ax=None, draw_perp=False, **kwargs):
```
mwaskom: Please move this parameter in front of `ax`

mwaskom:
Only other thing is that for consistency sake this should probably be added to `pointplot` as well?

jat255:
Thanks for the comments. I'll review these in detail in the next few days and push another commit with the corrections.

mwaskom:
Any progress here? 0.6 release is coming up and it would be good to have this in it.

seaborn/categorical.py
jat255: I had been using it internally, but have switched it to be exposed to the user

seaborn/categorical.py
jat255: I like your method better. I've implemented this in the latest commit

seaborn/categorical.py
jat255: Done.

seaborn/categorical.py
jat255: Done.

seaborn/categorical.py
jat255: Fixed.

seaborn/categorical.py
```diff
@@ -1230,24 +1230,46 @@ def estimate_statistic(self, estimator, ci, n_boot):
...
+                      ax,
+                      at_group,
+                      confint,
+                      colors,
+                      conf_lw=1,
+                      capsize=0,
+                      **kws):
+
+        kws.setdefault("lw", mpl.rcParams["lines.linewidth"] * conf_lw)
+        kws.pop('lw')
```
jat255: I had to add this because when plotting with `pointplot`, another `lw` is explicitly specified, so I removed it from the keyword list. I guess with the way this is, there is no reason to have line 1242 at all?

seaborn/categorical.py
```diff
@@ -2525,12 +2575,12 @@ def countplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
...
         raise TypeError("Cannot pass values for both `x` and `y`")
     else:
-        raise TypeError("Must pass valus for either `x` or `y`")
+        raise TypeError("Must pass values for either `x` or `y`")
 
     plotter = _BarPlotter(x, y, hue, data, order, hue_order,
                           estimator, ci, n_boot, units,
                           orient, color, palette, saturation,
-                          errcolor)
+                          errcolor, None, None)
```
jat255: Hopefully this is sufficient for the tests to pass... If not, please let me know how this should be done

mwaskom:
I think you might need to rebase on master to get the tests running again.

jat255:
> I think you might need to rebase on master to get the tests running again.

I think I did this correctly? I usually merge, rather than rebasing. Hopefully I haven't mucked things up.

mwaskom:
Hm, looks like you maybe had a slightly stale version of `master` locally?

jat255:
Yeah, something unusual happened... I pulled in all the changes from master first. Would it be best to reset to bb61668?

mwaskom:
I think it might be ok, I'll see what happens when I get home tonight and pull it down to my main seaborn development environment.

mwaskom:
Looks like any weirdness is in doc files anyway so shouldn't affect the build.

cachitas:
There's also a style used by GraphPad Prism software (see figure attached from their docs) in barplots where only the upper half of the error bar is drawn. Do you think it is possible to add this option, since you are already enhancing this feature?
image

mwaskom:
> Do you think it is possible to add this option, since you are already enhancing this feature?

No, sorry, I think that "dynamite plot" style falls somewhere between "mostly unhelpful" and "actively misleading". Also give that seaborn computes bootstrapped CIs and not parametric standard errors, showing only one side of the CI would not make sense.

mtresch:
What is the status of the End caps on bar plots?

dcgemperline:
Is there anyway I can help get this into master ? It's a pretty nice feature, and only requires a couple of changes. If all you need is a rebase to the current dev and a new pull request submitted I think I can help.

jat255:
Not sure what the current status is. This was good to go at the time, I
think, but I sort of lost interest. If someone wants to fork it and
resubmit, that's fine by me.

On Wed, Mar 30, 2016 at 1:48 PM, David <notifications@github.com> wrote:

> Is there anyway I can help get this into master ? It's a pretty nice
> feature, and only requires a couple of changes. If all you need is a rebase
> to the current dev and a new pull request submitted I think I can help.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/mwaskom/seaborn/pull/606#issuecomment-203548941>
>



-- 
Joshua Taillon
Materials Science and Engineering
University of Maryland, College Park
www.joshuataillon.com

dcgemperline:
Looks like there are some issues with make test after integrating your changes. I am working on a fix right now. I had to change some of the test case calls to match the added arguments.

dcgemperline:
I started a new pull request that should pass the automated testing. I reworked some of the way it was implemented so there was less hard coded default values.

https://github.com/mwaskom/seaborn/pull/898

mwaskom:
Closed with #898

Issue #509, opened 2015-04-10
Multiple calls to FacetGrid.set_titles() create misaligned margin titles


ebellm:
I'd like to use multiple calls to FacetGrid.map() to layer multiple outputs.  However, the right margin titles become misaligned, because each call to map() triggers an additional set_titles() call, and it appears the inferred bounding box coordinates are changing.

Here's a minimal working example from v 0.5.1:

```
import numpy as np
import seaborn as sns
import matplotlib.pyplot as plt
sns.set(style="darkgrid")

tips = sns.load_dataset("tips")
g = sns.FacetGrid(tips, row="sex", col="time", margin_titles=True)
bins = np.linspace(0, 60, 13)
g.map(plt.hist, "total_bill", color="steelblue", bins=bins, lw=0)
g.map(plt.hist, "total_bill", color="green", bins=bins, lw=0)
```

The following also reproduces the problem:

```
g2 = sns.FacetGrid(tips, row="sex", col="time", margin_titles=True)
g2.map(plt.hist, "total_bill", color="steelblue", bins=bins, lw=0)
g2.set_titles()
```

![factorgrid_titles](https://cloud.githubusercontent.com/assets/733983/7098598/9540697c-df97-11e4-81e8-d56d90ff37be.png)

mwaskom:
I think this might be somewhat unavoidable, because the margin titles are basially a hack that are just adding matplotlib figure texts (matplotlib doesn't support title object off the side of the Axes). I'm not sure exactly how it could be handled in seaborn, but if you want to avoid this you could remove all the `figure.texts` objects before the final call to `FacetGrid.map`, I think that will work.

mwaskom:
See #440 for an example.

ebellm:
Ah, great!  Calling `plt.setp(g.fig.texts, text="")` before the last `map` call does indeed work around the problem.  Thanks!

shoyer:
Just encountered this issue and @ebellm's hack fixed it for me. Perhaps we could add it to the docs?

mwaskom:
Added two notes on this in #586, hopefully that will help dispel confusion. (Alternatively someone could try to get official support for titling an Axes off the right margin into matplotlib itself, then this issue would go away).

mwaskom:
See #612, which changes how the margin titles are drawn. This improves a few things, including adding a solid background to the text object to hide any texts behind it. So this hack is hopefully no longer necessary.

parchd-1:
[...]

PR #612, opened 2015-06-28
Draw margin titles with ax.annotate instead of f.text

### LINKED ISSUE #303: Factorplot margin_titles end up outside/overlapping with the legend

mwaskom:
This should keep the titles tighter to the axes when the figure
layout changes, such as adding a legend. This PR therefore fixes #303.

Also added a background to the text, which should prevent the issue
where adding multiple titles "stacks" the row titles (#509).

Also changed the default `savefig` `bbox_inches` to "tight" in Grid objects, so these texts won't get clipped when saving.

### DIFF
--- a/doc/releases/v0.6.0.txt
+++ b/doc/releases/v0.6.0.txt
@@ -79,7 +79,7 @@ Other additions and changes
 
 - The various property dictionaries that can be passed to ``plt.boxplot`` are now applied after the seaborn restyling to allow for full customizability.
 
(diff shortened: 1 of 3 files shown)


Teodormz:
plt.setp(g.fig.texts, text="") doesn't work anymore, and I can't figure out a different way to remove all the row titles.

Teodormz:
fast way. This works but is slow
for cax in g.axes.flat:
    if not cax.has_data():
        cax.texts[0].set_text("");

PR #621, opened 2015-07-02
Compute linkage correctly when fastcluster is not installed

marcelm:
To reproduce the problem this PR tries to fix, create a virtualenv without fastcluster and run [the code in this clustermap example](http://stanford.edu/~mwaskom/software/seaborn/examples/structured_heatmap.html). The rows and columns will be sorted in a different way than what is shown on that page.

When fastcluster is not installed, `_calculate_linkage_scipy` is used to
compute linkages. It contains an erroneous call to squareform().
distance.pdist() already returns a condensed distance matrix that is
required as input for the linkage() function. Converting it to a matrix
with squareform() is possible and linkage() will accept the resulting
matrix, but consider it to be a list of vectors of observations and run
distance.pdist() again on it. Thus, we get a linkage based on distances
between distances.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -522,8 +522,7 @@ class _DendrogramPlotter(object):
             UserWarning('This will be slow... (gentle suggestion: '
                         '"pip install fastcluster")')
 
-        pairwise_dists = distance.squareform(
-            distance.pdist(self.array, metric=self.metric))
+        pairwise_dists = distance.pdist(self.array, metric=self.metric)
         linkage = hierarchy.linkage(pairwise_dists, method=self.method)
         del pairwise_dists
         return linkage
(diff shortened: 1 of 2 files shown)


marcelm:
scipy/scipy#2614 also has some details.

mwaskom:
cc @olgabot

mwaskom:
Looks like the test failures are meaningful but I don't know enough about the linkage/dendrogram stuff to figure out what's going on.

marcelm:
I’ve looked into this again and it - thankfully - wasn’t so difficult. The tests were also assuming at two positions that one can/should feed squareform()ed matrices into linkage.hierarchy. Removing those two extraneous calls to squareform() makes the tests pass.

mwaskom:
@olgabot are you :+1: on merge?

mwaskom:
Thanks!

marcelm:
Thanks for merging!

PR #629, opened 2015-07-06
Factor Heatmaps

lukauskas:
This pull request implements the ability to treat heatmap data as factors. 

Namely, the current implementation draws heatmaps as follows:

``` python
import numpy as np; np.random.seed(0)
integer_data = np.round(np.random.rand(10, 12) * 2)
ax = sns.heatmap(integer_data)
```

01-current

This is visually nice, but the colorbar on the right hand side is a bit confusing as the values are purely in the set {0, 1, 2}.

This PR adds the `as_factors` option to `seaborn.heatmap`, which (1) treats values as qualitative, rather sequential (2) prints a legend, not colorbar.

``` python
ax = sns.heatmap(integer_data, as_factors=True)
```

02-as-factors

Boolean data is set to default to `as_factors=True` if not specified (but can be overriden with providing False explicitly):

``` python
boolean_data = np.random.rand(10, 12) > 0.5
ax = sns.heatmap(boolean_data)
```

03-boolean

This also allows one to plot non-numeric data. For instance, we can plot string data directly (defaults to `as_factor=True` too):

``` python
import string
_to_letters = np.vectorize(lambda x: string.uppercase[int(x)])
data_as_letters = _to_letters(integer_data)
sns.heatmap(data_as_letters, annot=True)
```

04-string-data

By default the factors are ordered using python's `sorted()`, but the sorting order (and colors) can be overriden directly:

``` python
sns.heatmap(data_as_letters, annot=True, 
                      as_factors=['C', 'A', 'B'], cmap=['red', 'purple', 'blue'])
```

05-string-data-uglier

Well, that's pretty much it.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -163,23 +163,42 @@ class _HeatMapper(object):
         data = data.ix[::-1]
         mask = mask.ix[::-1]
 
+        plot_data = np.ma.masked_where(np.asarray(mask), plot_data)
+
         if as_factors is None:
-            if data.values.dtype in [np.dtype(bool), np.dtype(str), np.dtype(object)]:
+            print data.values.dtype
+
+            if data.values.dtype in [np.dtype(bool), np.dtype(object)] \
+                    or np.issubdtype(data.values.dtype, str):
                 as_factors = True
             else:
                 as_factors = False
+            print as_factors
 
         if as_factors:
-            unique_values = sorted(pd.Series(np.ravel(plot_data)).unique())
+            unique_values = sorted(pd.Series(np.ravel(plot_data)).dropna().unique())
             # Transform data into numeric one by mapping each unique value to an int
             unique_values_map = {val: i for i, val in enumerate(unique_values)}
-            vector_get = np.vectorize(unique_values_map.get)
+
+            def _get_func(x):
+                try:
+                    return unique_values_map[x]
+                except KeyError as e:
+                    nan = x is None
+                    try:
+                        nan |= np.isnan(x)
+                    except TypeError:
+                        pass
+                    if nan:
+                        return np.nan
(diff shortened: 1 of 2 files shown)


mwaskom:
Yes this is cool. I won't get a chance to look closely until probably later in the week, but in the meantime can you fix the failing pep8 check?

I think also the R-centric "factor" terminology should be converted to "categorical", which is the term used in seaborn/pandas for the same kind of object.

sauliusl:
Thanks for the quick reply, guys, appreciate that. 

I just pushed a commit that should fix the PEP8 warnings (most of them were just reduction of line widths to 80 chars as I am used to 100 char layout).

I agree with the categorical terminology too, will change that in the next few days.
Another thing worth considering is maybe moving `heatmap(as_categorical=True)` to `categorical_heatmap` (or something) purely because quite a lot of code at its current state has the `if as_categorical` blocks all over the place, which is kind of a code smell, I have to admit. Not sure which representation is better from user's perspective.

mwaskom:
I think `categorical` would be better than `as_categorical`

mwaskom:
I don't think from a user-perspective there is a need for two functions since they basically have the same interface and parameters (right?) but it's possible there is cause for more refactoring on the internals.

mwaskom:
> By default the factors are ordered using python's sorted(), but the sorting order (and colors) can be overriden directly:

This should use `seaborn.utils.category_order` to be consistent with the rest of the package.

mmagnuski:
BTW - don't you think that in the boolean heatmap True should be white? You know, it goes back to Plato - the intuition that truth is like light. :wink:

jrderuiter:
Any timeline on this?

mwaskom:
Closing as stalled.

PR #632, opened 2015-07-06
factorplot: Extend input check when kind="count"

kyleam:
When kind is "count" for factorplot, raise ValueError if both x and y
are given.  Otherwise, x_ and _y will not be defined, leading to an
UnboundLocalError.


mwaskom:
Looks good, just needs a little fix for the pep8 issue

kyleam:
Thanks.  Updated with a shorter message.

PR #642, opened 2015-07-09
Add font.size to plotting context

### LINKED ISSUE #635: font.size should be part of rcmod.plotting_context?
When adding text to a figure with `plt.text`, it would be nice if this scaled with seaborn's `font_scale`. Would this break something else?

petebachant:
Resolves #635. 

`font.size` was not being set by `rcmod`, which meant that text generated by `plt.text` and, e.g. tick labels in scientific notation were not scaled by `font_scale`. This is addressed by using matplotlib's default `font.size: 12` and scaling accordingly.

### DIFF
--- a/seaborn/rcmod.py
+++ b/seaborn/rcmod.py
@@ -44,6 +44,7 @@ _style_keys = (
 _context_keys = (
     "figure.figsize",
 
+    "font.size",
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks!

Issue #583, opened 2015-06-05
Getting min-max ranges with factorplot


jackjackk:
If I use a 100% confidence interval, do I get min-max ranges over the "units" dimension with factorplot?
In this trivial case, is there a combination of parameters I could pass to factorplot to avoid the bootstrap and speed up the procedure?
Thanks

mwaskom:
No, a 100% CI would give you the minimum and maximum of the bootstrap distribution. It's not straightforwardly related to what you want to show, thought it sounds like you might be better suited with a boxplot?

mbatchkarov:
This is something that I'd like to see in `factorplot`, too. Since I'm not sure what the proper names for what I'm talking about are, I'll make up some terminology:
 - *CI of the mean* is what `factorplot` shows currently with `kind='bar'`, which is estimated via bootstrapping
 - *CI of the data* is what the OP is asking for, percentiles of the original data. This is what `boxplot`/`violinplot` show.

I want the latter. A box/violin plot would do, but they are much harder to read as the natural visual grouping by treatment is lost. Compare the two plots below: in the box plot it is not at all obvious that the two leftmost treatments are the same (Add). Also, I find it a bit confusing that the error bars represent one thing if I use `kind='box'` and another if I use `kind='bar'`.


![bar](https://cloud.githubusercontent.com/assets/937114/8575529/5ee6cdf6-2594-11e5-898e-e9af3bca7b51.png) ![box](https://cloud.githubusercontent.com/assets/937114/8575531/5f05daf2-2594-11e5-8932-4e3fd41e9bc7.png)

Would it make sense to have something like `err_style='ci_mean_bars'` vs `err_style='ci_data_band'` for `factorplot`?

I don't think what I want is possible currently. The closest to a solution I've got is 

    sns.factorplot(..., estimator=lambda foo: foo)

This is broken because:
 - the new estimator return a list, not a scalar, so [this](https://github.com/mwaskom/seaborn/blob/8d46524a8514dcab228c9f78a1f9b2be645bfd40/seaborn/algorithms.py#L76-77) would fail
 - would still bootstrap, so kind of slow
 - not sure if estimating *CI of the data* through bootstrapping makes sense

Issue #450, opened 2015-02-14
Extend tsplot to plot observation quantiles?


shoyer:
We often like to plot time-series almost exactly like this example from tsplot:
![image](https://cloud.githubusercontent.com/assets/1217238/6197735/be036856-b39f-11e4-840b-0278b934bd71.png)

...but with the plot showing quantiles of the measurement distribution, not CI bands for the mean/median.

We could call this `err_style='unit_band'`, which would make sense, though controlling it with the argument `ci` would be a little strange.

mwaskom:
I think this is basically what #56 is. `tsplot` could use a lot of work...

jetheurer:
Is this supported now?

shoyer:
@jetheurer  Nope... this still needs someone to put in the work :).

jetheurer:
@shoyer Do you have a example code of what you like to do?  I'd be interested in building off of it.

mwaskom:
Unfortunately, the tsplot internals are poorly implemented and untested, so any enhancement is probably going to have to wait until it can be refactored. 



—

shoyer:
@jetheurer Not that I can share publicly, I'm afraid...

All the computation is handled by the `pandas.DataFrame.quantile` method, though.

mwaskom:
Closing with deprecation of `tsplot`.

Issue #56, opened 2013-12-27
Add prediction interval error styles to tsplot


mwaskom:

mwaskom:
Closing with tsplot deprecation

PR #643, opened 2015-07-12
adds ability to visualise distribution of data

mbatchkarov:
Hi,

this PR is a discussion of #56, #450 and #583, and possibly related to #474. I've only implemented a proof-of-concept in the context of `tsplot`. The same could be done for `factorplot`, but it would require significantly more work and possibly an API change. The point of this PR is to demonstrate visually what the issue is and to start a discussion on whether a change is really needed (I am not convinced myself).

Let's generate time series with high variance and one with low variance. Both will have a very tight confidence interval of the mean.

``` python
import pandas as pd
import seaborn as sns
import numpy as np


samples_per_time = 500
time = np.repeat(np.arange(0, np.pi, 0.05), samples_per_time)
num_samples = len(time)
y = np.sin(time) + np.random.normal(0, .01, size=time.shape)
y1 = np.cos(time) + np.random.normal(0, .2, size=time.shape)
units = list(range(samples_per_time)) * (num_samples // samples_per_time)

df = pd.DataFrame(dict(
    cond=['low var'] * num_samples + ['high var'] * num_samples,
    unit=units * 2,
    value=np.hstack((y, y1)),
    time=np.hstack((time, time))))
```

The current version of `seaborn` (0.6) does not make it obvious that the variances are different.

``` python
sns.tsplot(df, time='time', value='value', unit='unit', 
           condition='cond', ci=[30, 68, 95])
```

errorbars

With the proposed change, one can write a custom bootstrap-like function:

``` python
def my_bootstrap(*args, **kwargs):
    return np.vstack(args) # can also bootstrap here if the sample is small

sns.tsplot(df, time='time', value='value', unit='unit', 
           ci=[30, 68, 95], condition='cond', bootstrap_func=my_bootstrap)
```

Now the error bars show the empirical distribution of the data. It is immediately obvious that the variances are different.
errorbars2

I do not know how many users (other than myself) would prefer to see the second figure instead of the first. Of course, both have their uses, I just find the latter more informative for exploratory work

In terms of implementation, passing in the bootstrap function as a parameter (similarly to how `estimator` is passed in) is the least invasive approach. If we think the feature would be commonly used, it may become an extra keyword, as I suggested in #583.

### DIFF
--- a/seaborn/timeseries.py
+++ b/seaborn/timeseries.py
@@ -17,7 +17,7 @@ from .palettes import color_palette
 def tsplot(data, time=None, unit=None, condition=None, value=None,
            err_style="ci_band", ci=68, interpolate=True, color=None,
            estimator=np.mean, n_boot=5000, err_palette=None, err_kws=None,
-           legend=True, ax=None, **kwargs):
+           legend=True, ax=None, bootstrap_func=algo.bootstrap, **kwargs):
     """Plot one or more timeseries with flexible representation of uncertainty.
 
     This function is intended to be used with data where observations are
@@ -278,7 +278,7 @@ def tsplot(data, time=None, unit=None, condition=None, value=None,
         x = df_c.columns.values.astype(np.float)
 
         # Bootstrap the data for confidence intervals
-        boot_data = algo.bootstrap(df_c.values, n_boot=n_boot,
+        boot_data = bootstrap_func(df_c.values, n_boot=n_boot,
                                    axis=0, func=estimator)
         cis = [utils.ci(boot_data, v, axis=0) for v in ci]
         central_data = estimator(df_c.values, axis=0)


mwaskom:
I am not in favor of this. In this example, the `bootstrap_func` is not doing a bootstrap, and the `ci` is not a confidence interval.

shoyer:
We discussed this (with a different API) in https://github.com/mwaskom/seaborn/issues/450

twiecki:
I was missing this. maybe the solution to @mwaskom's point is to rename the args.

ariddell:
This would, indeed, be a really useful feature. The current behavior is a little unexpected given how well seaborn handles full distributions elsewhere (e.g. kdeplot)

mwaskom:
Perhaps. As stated elsewhere, I'm not really in favor of any changes to `tsplot` until it can be refactored and better tested. The general idea here would be nice to have, but this particular proposal is a hack.

It also isn't that hard to make this kind of plot directly in matplotlib, so it doesn't feel urgent.

ariddell:
@mwaskom I'll definitely defer to you. Having wonderful libraries like seaborn has made me a bit lazy about gaining fluency with the matplotlib plumbing.

Do you have any pointers on how the plot should be done (i.e., "directly in matplotlib"). I didn't find the relevant example given in the pandas documentation terribly appealing: http://pandas.pydata.org/pandas-docs/stable/visualization.html#plotting-directly-with-matplotlib

mwaskom:
Closing with tsplot deprecation (but see also #1208)

Issue #649, opened 2015-07-16
zscore in clustermap is wrong


michael-imbeault:
The zscore is defined as (x - mean) / standard deviation (https://en.wikipedia.org/wiki/Standard_score), yet in z_score in clustermap, it is defined as:

z_scored = (z_scored - z_scored.mean()) / z_scored.var()

instead it should be:

z_scored = (z_scored - z_scored.mean()) / z_scored.std()

This caused quite a bit of hair pulling finding the cause of differences between seaborn and plots obtained with other tools.

Thanks!
Michael

olgabot:
Mea culpa - thanks for catching it. I'll fix it right away.

PR #650, opened 2015-07-16
Divide by standard dev instead of variance for z-score

olgabot:
Fixes https://github.com/mwaskom/seaborn/issues/649


mwaskom:
Thanks for the quick fix @olgabot

PR #661, opened 2015-07-30
Letter Value Plots

drewokane:
This pull request adds the `_LVPlotter` class, `lettervalueplot` function, documentation, and usage examples.

# TODO
-----------
- [x] PEP8 errors
- [x] Rename plotting function `lvplot`
- [x] Median lines should stay inside boxes
- [x] Change default scale to "exponential"
- [x] Rename `box_widths` to `scale`
- [x] Put default values into function signature
- [x] Remove deprecation warnings
- [x] Rename `p` to `outlier_prop`
- [x] Change `outlier_prop` to take float, [0, 1].
- [x] Fix documentation
- [x] Change outlier plotting symbol to `d` for internal consistency
- [x] Make width functions class level functions
- [x] Write tests

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1507,10 +1507,17 @@ class _LVPlotter(_CategoricalPlotter):
         upper_out = vals[np.where(vals > edges[1])[0]]
         return np.concatenate((lower_out, upper_out))
 
+    def _width_functions(width_func):
+        # Dictionary of functions for computing the width of the boxes
+        width_functions = {'linear' : lambda h, i, k: (i + 1.) / k,
+                           'exponential' : lambda h, i, k: 2**(-k+i-1),
+                           'area' : lambda h, i, k: (1 - 2**(-k+i-2)) / h}
+        return width_functions[width_func]
+
     def _lvplot(self, box_data, positions,
                 color=[255. / 256., 185. / 256., 0.],
                 vert='v', widths=1, k_depth='proportion',
-                ax=None, p=None, box_widths='linear',
+                ax=None, p=None, box_widths='exponential',
                 **kws):
 
         x = positions[0]
@@ -1530,14 +1537,9 @@ class _LVPlotter(_CategoricalPlotter):
             # letter-value plot
             box_ends, k = self._lv_box_ends(box_data, k_depth=k_depth, p=p)
 
-            # Dictionary of functions for computing the width of the boxes
-            width_functions = {'linear' : lambda h, i, k: (i + 1.) / k,
-                               'exponential' : lambda h, i, k: 2**(-k+i-1),
-                               'area' : lambda h, i, k: (1 - 2**(-k+i-2)) / h}
-
             # Anonymous functions for calculating the width and height
             # of the letter value boxes
-            width = width_functions[box_widths]
+            width = self._width_functions(box_widths)
             height = lambda b: b[1] - b[0]
 
             # Functions to construct the letter value boxes
@@ -1623,6 +1625,7 @@ class _LVPlotter(_CategoricalPlotter):
                                           widths=self.width,
                                           k_depth=self.k_depth,
                                           ax=ax,
+                                          box_widths=self.box_widths,
                                           **kws)
 
             else:
@@ -1654,6 +1657,7 @@ class _LVPlotter(_CategoricalPlotter):
                                               widths=self.nested_width,
                                               k_depth=self.k_depth,
                                               ax=ax,
+                                              box_widths=self.box_widths,
                                               **kws)
 
     def plot(self, ax, boxplot_kws):
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1788,7 +1788,7 @@ _categorical_docs = dict(
     factorplot : Combine categorical plots and a class:`FacetGrid`.\
     """),
     lettervalueplot=dedent("""\
-    lv
+    lettervalueplot : An extension of the boxplot for long-tailed and large data sets.
     """),
     )
 
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1561,7 +1561,6 @@ class _LVPlotter(_CategoricalPlotter):
 
             # Calculate the medians
             y = np.median(box_data)
-            w = (widths + .1)
 
             # Calculate the outliers and plot
             outliers = self._lv_outliers(box_data, k)
@@ -1571,7 +1570,7 @@ class _LVPlotter(_CategoricalPlotter):
                             for i, b in enumerate(zip(box_ends, w_area))]
 
                 # Plot the medians
-                ax.plot([x -  w / 2, x + w / 2], [y, y], c='k', alpha=.45, **kws)
+                ax.plot([x -  widths / 2, x + widths / 2], [y, y], c='k', alpha=.45, **kws)
 
                 ax.scatter(np.repeat(x, len(outliers)), outliers,
                            marker=r"$\ast$", c=color)
@@ -1580,7 +1579,7 @@ class _LVPlotter(_CategoricalPlotter):
                             for i, b in enumerate(zip(box_ends, w_area))]
 
                 # Plot the medians
-                ax.plot([y, y], [x -  w / 2, x + w / 2], c='k', alpha=.45, **kws)
+                ax.plot([y, y], [x -  widths / 2, x + widths / 2], c='k', alpha=.45, **kws)
 
                 ax.scatter(outliers, np.repeat(x, len(outliers)),
                            marker=r"$\ast$", c=color)
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1494,8 +1494,8 @@ class _LVPlotter(_CategoricalPlotter):
         except ValueError:
             k = 1
         # If the number happens to be less than 0, set k to 0
-        if k < 0.:
-            k = 0
+        if k < 1.:
+            k = 1
         # Calculate the upper box ends
         upper = [100*(1 - 0.5**(i+2)) for i in range(k, -1, -1)]
         # Calculate the lower box ends
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1464,7 +1464,7 @@ class _LVPlotter(_CategoricalPlotter):
         self.linewidth = linewidth
 
         if scale is None:
-            scale = 'linear'
+            scale = 'exponential'
         self.scale = scale
 
         self.outlier_prop = outlier_prop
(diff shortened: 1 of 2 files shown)


mwaskom:
If you want to work out the pep8 errors locally you can install `pip install pep8==1.5 ` which will be the version on travis. The `pep8` command got a little stricter and some things in seaborn fail, so just running the latest version won't be that informative about the new code.

mwaskom:
I think the function could be called `lvplot` like it is in R. A little confusing with `lmplot`, but `lettervalueplot` is such a mouthful.

seaborn/categorical.py
```diff
@@ -1438,6 +1440,229 @@ def plot(self, ax):
         if self.orient == "h":
             ax.invert_yaxis()
 
+class _LVPlotter(_CategoricalPlotter):
+
+    def __init__(self, x, y, hue, data, order, hue_order,
+                 orient, color, palette, saturation,
+                 width, k_depth, linewidth, box_widths):
+
+        if width is None:
```
mwaskom: These default values should be specified in the function signature for `lvplot`.

seaborn/categorical.py
```diff
@@ -2870,3 +3098,193 @@ def factorplot(x=None, y=None, hue=None, data=None, row=None, col=None,
         <seaborn.axisgrid.FacetGrid object at 0x...>
 
     """).format(**_categorical_docs)
+
+def lettervalueplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
+            orient=None, color=None, palette=None, saturation=None,
+            width=None, k_depth=None, linewidth=None, box_widths=None,
+            ax=None, **kwargs):
+
+    # Try to handle broken backwards-compatability
```
mwaskom: You don't need this section, it's only relevant to handle the changes from < 0.6 in `boxplot` and `violinplot`, but nobody would have old `lvplot` code using these parameters.

seaborn/categorical.py
```diff
@@ -2870,3 +3098,193 @@ def factorplot(x=None, y=None, hue=None, data=None, row=None, col=None,
...
+        All methods are detailed in Wickham's paper. Each makes different
+        assumptions about the number of outliers and leverages different
+        statistical properties.
+    {linewidth}
+    box_widths : "linear" | "exponential" | "area"
+        Method to use for the width of the letter value boxes. All give similar
+        results visually. "linear" reduces the width by a constant linear factor,
+        "exponential" uses the proportion of data not covered, "area" is
+        proportional to the percentage of data covered.
+    p : float, optional
```
mwaskom: I think `outlier_prop` might be a better name for this argument, just `p` isn't that informative.

seaborn/categorical.py
```diff
@@ -2870,3 +3098,193 @@ def factorplot(x=None, y=None, hue=None, data=None, row=None, col=None,
...
+        Proportion of data believed to be outliers. Is used in conjuction with
+        k_depth to determine the number of percentiles to draw. Defaults to 8
+        outliers.
+    {ax_in}
+    kwargs : key, value mappings
+        Other keyword arguments are passed through to ``plt.boxplot`` at draw
```
mwaskom: To `plt.boxplot`?

mwaskom:
This looks good on a very brief first pass, but it will be very important to get tests on the numerical internals.

seaborn/categorical.py
```diff
@@ -1438,6 +1440,229 @@ def plot(self, ax):
...
+                           marker=r"$\ast$", c=color)
+            else:
+                boxes = [horz_perc_box(x, b[0], i, k, b[1])
+                            for i, b in enumerate(zip(box_ends, w_area))]
+
+                # Plot the medians
+                ax.plot([y, y], [x -  w / 2, x + w / 2], c='k', alpha=.45, **kws)
+
+                ax.scatter(outliers, np.repeat(x, len(outliers)),
+                           marker=r"$\ast$", c=color)
```
mwaskom: Why not just a normal `"o"`?

seaborn/categorical.py
```diff
@@ -1438,6 +1440,229 @@ def plot(self, ax):
...
+                               'exponential' : lambda h, i, k: 2**(-k+i-1),
+                               'area' : lambda h, i, k: (1 - 2**(-k+i-2)) / h}
+
+            # Anonymous functions for calculating the width and height
+            # of the letter value boxes
+            width = width_functions[box_widths]
+            height = lambda b: b[1] - b[0]
+
+            # Functions to construct the letter value boxes
+            def vert_perc_box(x, b, i, k, w):
```
mwaskom: Might as well make these class-level functions so you can test them.

seaborn/categorical.py
drewokane: I didn't really have a preference as far as marker for outliers. Matplotlib uses '+' markers, so I could try to be consistent with Matplotlib or go with 'o'. Thoughts?

mwaskom: I liked the circles you had in the example plots I just think it will be faster to draw those with `"o"` than with a custom latex marker, right?

mwaskom: Of you could just `"d"` to be consistent with seaborn boxplots i guess...

mwaskom:
Also before I forget this should get added as a plot kind in `factorplot` too eventually.

mwaskom:
The median lines extend past the width of the boxes, so they overlap on nested plots. Is this intentional? I don't love the effect.

dcouraphf7aaaaaelftksuqmcc

mwaskom:
I find this kind of weird:

```python
df = pd.DataFrame({"normal": np.random.normal(3, 1, 10000),
                   "gamma": np.random.gamma(3, .5, 10000)})

sns.lettervalueplot(data=df)
sns.violinplot(data=df, scale="width")
```

az5eiuygm1 2aaaaaelftksuqmcc


I think I still don't quite understand how these things get drawn, but I would have expected the letter-value plots to look more like a discretized violinplot. Is that wrong?

drewokane:
Your intuition wouldn't be off in this case. The widths of the boxes are determined by which width function is chosen in `_lvplot`. My initial stab at this I liked the look of the "linear" widths, but if you use the "exponential" widths it actually looks like a discrete violin plot (as you mentioned). Perhaps changing the default to "exponential" will be better for consistency sake. (Also, I realized there is a bug, in that I am not passing the keyword argument to the plotter (embarassed))

```
df = pd.DataFrame({"normal": np.random.normal(3, 1, 10000),
                   "gamma": np.random.gamma(3, .5, 10000)})

sns.violinplot(data=df, scale="width")
sns.lettervalueplot(data=df, box_widths='exponential')
```
image

mwaskom:
Ah, OK. I had played around with that keyword argument, but couldn't seem to get the effect I was looking for. I guess that explains why :)

mwaskom:
Speaking of, I think the `box_widths` parameter might make more sense as `scale`, for better consistency with the `violinplot` paramter. The values aren't the same, but they have a similar effect. Also for some reason I kept expecting `box_widths` to take a numeric size value.

seaborn/categorical.py
drewokane: Robotic copy paste. The keywords are actually only going to `plt.plot`. I think I'll change this so it goes to `plt.plot` and `plt.scatter`.

seaborn/categorical.py
drewokane: Changed it to 'd' for consistency.

seaborn/categorical.py
drewokane: Good idea.

seaborn/categorical.py
drewokane: Good call. Another thought I had for this is that `outlier_prop` could take either (0, 1.] float or an actual number of suspected outliers, i.e. 6. Thoughts.

mwaskom: I think proportion makes more sense since in a lot of cases you won't have the same number of observations in each bin (and it allows the default the be agnostic to the number of observations as well). Why not inclusive of 0, aren't there situation where you don't want to identify outliers?

mwaskom:
Any update on tests?

drewokane:
I've added some basic tests so far and plan on adding tests for the numerical internals, e.g. testing the letter value calculations, etc. I'll push the latest edits I have today.

mwaskom:
Cool. By the way, I would generally advise writing the tests in the other direction. I've found that it's best to start directly testing the core numeric parts, then to pull back and test some of the plotting methods, and finally pull back further and add some high-level smoke tests for the function itself (what I assume you mean by "basic tests"). The reason is that this way lets the code coverage report be a useful guide to how complete the tests are. If you start with tests that run the function and just ask "are there some plots on the axes", it will look like the internals are "covered", but the plots very well may be wrong. It might be helpful to rename the basic tests so nose doesn't run them as you develop and use the coverage report to make sure you're really finding all the corners in the numeric code. Anyway, just some thoughts.

mwaskom:
Also can you add a TODO list in the original issue just so we can track the items that have come up in the comments?

drewokane:
I've completed writing the tests (making them Python 3 forward compatible) and the other tasks on the checklist. I've checked the tests (locally with nosetests) and the only failures I can see are the ones connected with other modules, not `TestLVPlotter`.

mwaskom:
The Python 3 errors are in `TestLVPlotter`. It's hard to tell with the Python 2 build as it didn't make it past the PEP8 check.

drewokane:
Yeah I checked Travis. I didn't realize `filter` and `map` were iterators in Python 3. Hopefully the modifications pass muster.

drewokane:
Unfortunately I can't duplicate the PEP8 error Travis is getting (and I made sure to install `pep8==1.5`).

seaborn/tests/test_categorical.py
```diff
@@ -2178,3 +2179,208 @@ def test_plot_colors(self):
...
+        pal = palettes.color_palette("deep", 2)
+        for patch, color in zip(ax.artists, pal * 2):
+            nt.assert_equal(patch.get_facecolor()[:3], color)
+
+        plt.close("all")
+
+    def test_draw_missing_boxes(self):
+
+        ax = cat.lvplot("g", "y", data=self.df,
+                         order=["a", "b", "c", "d"])
```
mwaskom: This line is over-indented

drewokane:
Build is passing on all versions, PEP8 is all done.

drewokane:
Any more comments, concerns, suggestions?

drewokane:
Not trying to be bothersome but I wanted to check in a see if you had any further comments or suggestions for the PR? I completely understand if you're swamped with other stuff.

mwaskom:
Yeah, sorry about letting this drop. This is a substantial enough addition that I've really wanted to make time to go through it in detail and make sure I understand how everything works. Unfortunately, I've been very busy the past few months (and will be for the foreseeable future) and just have not been able to devote a large enough block of time to get my head around everything here. But since you put so much effort into this, I really do feel bad about letting this hang in limbo. I'll merge it now so that it can start to get some real use, hopefully any residual issues will get identified and ironed out before the next release. If you could try to stay on top of anything that pops up, that would be very helpful.

stonebig:
Hi,

May  the documentation take Matplotlib 1.5 into account now ?
````
ax = sns.lvplot(x="size", y="tip", data=tips.sort("size"))
# or, if you are on Matplotlib 1.5+
ax = sns.lvplot(x="size", y="tip", data=tips.sort_values(by="size"))
````

drewokane:
@mwaskom Sure, I can stay on top of any issues that pop up. Thanks!

mwaskom:
> May the documentation take Matplotlib 1.5 into account now ?

This is a change in pandas, not matplotlib, but yes that example should be updated.

PR #665, opened 2015-08-06
DOC: correct some typos

gef756:
I corrected some typos I found, as well as one :ref: link that was not being parsed correctly.


mwaskom:
Thanks!

Issue #242, opened 2014-07-29
Seaborn should intelligently label datetime axes


shoyer:
I'm often interested in plotting data indexed with a `pandas.DatetimeIndex` object / `datetime64` array. It would be nice if seaborn could recognize when the values along an axis are given by dates and provide sensible labels (like pandas).

laurieskelly:
my god @shoyer, i hope you eventually let that poor subject out of the scanner...

clarkfitzg:
@clearf @shoyer `fig.autofmt_xdate()` is pretty handy. http://matplotlib.org/users/recipes.html#fixing-common-date-annoyances

foobarbecue:
Wow, this is still open? Seaborn timeseries plots don't support time axis labels?

mwaskom:
There, now it's closed. Official answer: seaborn is not going to support time axis labels in `tsplot`, and @foobarbecue is blocked from the repository for being obnoxious.

joseortiz3:
I think its not difficult to extract the methods pandas uses to make such nice date axes. I was looking at it yesterday, the machinery can be tracked down and extracted.

szeitlin:
@joseortiz3 It would be great if you would open a PR with a fix for this, or at least outline the steps in detail if it's really that straightforward, so someone else can contribute a fix.

mwaskom:
This is an issue about a deprecated function so I am locking it.

PR #669, opened 2015-08-07
Allow pandas objects to flow through tsplot

clearf:
I'm submitting this PR to try to kickstart the conversation around https://github.com/mwaskom/seaborn/issues/242. It's a surprisingly annoying error. 

My understanding is that, while matplotlib doesn't handle all instances of pandas objects correctly, it definitely does not handle processing dates passed merely as values correctly, going so far as to throw an overflow error if sns.tsplot is called while trying to represent the data as dates: 

    fig, ax = plt.subplots()
    ax.xaxis.set_major_locator(mdates.AutoDateLocator())
    ax.xaxis.set_major_formatter(mdates.DateFormatter('%Y.%m.%d'))
    ax.xaxis_date()
    fig.autofmt_xdate()
    sns.tsplot(data=ph, time='signal_date',  
                        value='cum_ret', unit='key', condition='ticker', 
                        estimator=np.mean, interpolate=True, ax=ax)
    plt.show() # Throws overflow error. 

I'm not sure if other things break with this PR, but I do know that at least one thing gets fixed. 

And relying on matplotlib to implement doesn't seem great either – the initial conversation on this was 3 years ago, and the last comment 6 months ago. 

What's the best way to move forward on this?


clearf:
@mwaskom Hi Michael, just wanted to make sure this PR was pointing to the correct repository, etc. Would love to get a conversation started about this issue. 

Thanks,
Chris

mwaskom:
Sorry, I have been very busy with actual responsibilities lately and have had basically no time for any real thought about seaborn.

My basic position on this is that I am apprehensive about making any changes to `tsplot` until it can be refactored and rigorously tested. With that said, your change is quite small, so if that's really all that it takes to make it work, I am open to it.

The problem is that it is unclear what this change might disrupt, due to the lack of unit tests. So it will be important to manually test the different code paths rigorously to check that everything works as expected. The best place to start there would be with this [old notebook](https://github.com/mwaskom/seaborn/blob/v0.2/examples/timeseries_plots.ipynb), which i the first thing that comes to mind in terms of a single example of all the various things `tsplot` can do. If you can modify the example datasets to use datetime data for the time variable and everything runs smoothly and looks good, then it will provide some confidence. (Sorry in advance if the notebook doesn't run out of the box with modern libraries, it's a bit old, but I *think* it should either work or just require a few small tweaks).

This may be more work than you were expecting, sorry. The internals of `tsplot` are very old and existed before I had any expectation that seaborn would be something anyone else would want to use, much less contribute to.

clearf:
Great, I'll take a look. Thanks.

eyadsibai:
It would be nice to have this pull request in master...

PR #670, opened 2015-08-10
BUG/TST: ax.legend needs to be explicitly called for patch labels.

wrobstory:
Pretty minor. If you pass a `label`, but don't explicitly call `ax.legend`, you won't get any type of legend labeling: 

https://gist.github.com/wrobstory/c990c94fc4598a440e7f

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -248,6 +248,9 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
         else:
             ax.set_xlabel(axlabel)
 
+    if label:
+        ax.legend()
+
     return ax
 
 
--- a/seaborn/tests/test_distributions.py
+++ b/seaborn/tests/test_distributions.py
@@ -235,3 +235,54 @@ class TestJointPlot(object):
     def teardown_class(cls):
         """Ensure that all figures are closed on exit."""
         plt.close("all")
+
+
+class TestDistPlot(object):
+
+    data = np.random.normal(0, 0.1, 1000)
+
+    def test_distplot_legend_render(self):
+
+        # No label should indicate no legend
+        g = dist.distplot(self.data)
+        nt.assert_is_none(g.legend_)
+        plt.close("all")
+
+        # Label should create a legend
+        g = dist.distplot(self.data, label="Foo")
+        nt.assert_is_not_none(g.legend_)
+        plt.close("all")
+
+        # Keywords should still create a legend
+        g = dist.distplot(self.data, kde_kws={"label": "Foo"})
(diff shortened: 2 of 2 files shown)


mwaskom:
Oh THIS is what you meant by having an open PR. I missed this.

wrobstory:
Hehehehehehe. I will rebase this on master tonight.

mwaskom:
Thanks! I think you just need to remove the teardown methods and then make the `TestDistPlot` class inherit from `PlotTestCase`.

mwaskom:
I'm a little concerned that this might cause problems with some of the higher level functions in terms of legends suddenly appearing where you don't want them, but it's hard to say no to a PR that has 25x more test code than new feature code.

wrobstory:
If you think it's going to cause problems in higher level functions, we can close it. Here's what my code looks like right now without it: 
```python
    sns.regplot(x="Air Temperature Average (degF)", y="Precipitation Increment (in)",
                data=snotel_monthly, x_bins=10, ax=ax_318, label="Average",
                color=sns.color_palette("husl")[-1], order=order)
    sns.regplot(x="Air Temperature Minimum (degF)", y="Precipitation Increment (in)",
                data=snotel_monthly, x_bins=10, ax=ax_318, label="Minimum",
                color=sns.color_palette("husl")[-2], order=order)
    sns.regplot(x="Air Temperature Maximum (degF)", y="Precipitation Increment (in)",
                data=snotel_monthly, x_bins=10, ax=ax_318, label="Maximum",
                color=sns.color_palette("husl")[-3], order=order)
    ax_318.legend()
```
My biggest motivation for this patch was that I had to hunt around a bit in the source to figure out why the labels weren't appearing, but once I figured it out it wasn't a big deal.

mwaskom:
It is just a matter of a (not fully thought out or obvious) distinction between which functions should act like normal matplotlib functions, where they accept a label kwarg but then wait for the user to make a legend, versus higher-level seaborn functions, which do a legend by default. I think the distinction probably comes down to whether the information used for the label is passed to one of seaborn's arguments (like a `hue` kwarg) or gets into the plot through kwargs passed straight through.

It's hard to know whether it's more surprising/disruptive to get a legend when you don't want one or to not get a legend when you do.

wrobstory:
> It's hard to know whether it's more surprising/disruptive to get a legend when you don't want one or to not get a legend when you do.

Interesting point: in MPL I generally find it harder to remove components of a chart than add them. Maybe in this case we just let it be.

mwaskom:
It's not obvious that you turn a legend off by doing `ax.legend_ = None`? :)

PR #682, opened 2015-08-21
changed handling of annot_kws so that standard values can be overwritten

### LINKED ISSUE #681: annot_kws of heatmap doesn't allow to change standard values
`seaborn.heatmap` provides a keyword argument `annot_kws` to change the text properties of the annotations. However, it doesn't allow to overwrite the standard values for `color`, `ha` and `va`. I agree that these only need to be changed in rare cases (especially the nice coloring feature), but I still I believe that the functionality should be there.

hitzg:
This fixes #681. I didn't add an extra test, I think the changes are small enough.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -201,8 +201,9 @@ class _HeatMapper(object):
                 _, l, _ = colorsys.rgb_to_hls(*color[:3])
                 text_color = ".15" if l > .5 else "w"
                 val = ("{:" + self.fmt + "}").format(val)
-                ax.text(x, y, val, color=text_color,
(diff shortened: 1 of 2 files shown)


mwaskom:
Seems reasonable. Can you add a test?

hitzg:
I just added a simple test.
Somehow I could not run `make lint`: `pyflakes` choked at the `-x`. (error message: `pyflakes: error: no such option: -x`)

mwaskom:
The linter uses a fork that allows you to ignore more benign warnings, you can install it with

`pip install https://github.com/dcramer/pyflakes/tarball/master`

if you'd like.

Anyway, looks good!

PR #683, opened 2015-08-22
Do not create heatmap labels, when labels are false

sauliusl:
This PR changes the behaviour of `xticklabels=False` (also `yticklabels=False`).
Instead of creating a list of empty tick labels (as it did before), it now just operates on an empty list.
This has the effect that `axis_ticklabels_overlap` does not do the expensive overlap calculation downstream. And thus the code runs faster.

For instance, try out the following code snippet:

```python
import numpy as np
import fastcluster
import pandas as pd
import seaborn
ROWS, COLS = 1000, 10
np.random.seed(10)
random_data = np.random.randn(ROWS, COLS)
random_data = pd.DataFrame(random_data, 
                           index=['Row {}'.format(i) for i in range(ROWS)], 
                           columns=['Column {}'.format(i) for i in range(COLS)])
row_linkage = fastcluster.complete(random_data)
col_linkage = fastcluster.complete(random_data.T)
seaborn.clustermap(random_data, 
                   row_linkage=row_linkage, 
                   col_linkage=col_linkage, yticklabels=False)
```

Naively timing the `seaborn.clustermap` command at the end of the code, I get 5.7s using the code in master, whereas the PR drops it to roughly 3.8s.  Even more time is gained when the number of rows being plotted is even larger. This is quite important as in these situations one would want to have `yticklabels` off anyway (as nobody would be able to read them anyway...)

I am attaching figures as well to show that the outputs look identical visually.

master:

PR:

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -111,19 +111,19 @@ class _HeatMapper(object):
         if isinstance(xticklabels, int) and xticklabels > 1:
             xtickevery = xticklabels
             xticklabels = _index_to_ticklabels(data.columns)
-        elif isinstance(xticklabels, bool) and xticklabels:
+        elif xticklabels is True:
             xticklabels = _index_to_ticklabels(data.columns)
-        elif isinstance(xticklabels, bool) and not xticklabels:
-            xticklabels = ['' for _ in range(data.shape[1])]
+        elif xticklabels is False:
+            xticklabels = []
 
         ytickevery = 1
         if isinstance(yticklabels, int) and yticklabels > 1:
             ytickevery = yticklabels
             yticklabels = _index_to_ticklabels(data.index)
-        elif isinstance(yticklabels, bool) and yticklabels:
+        elif yticklabels is True:
             yticklabels = _index_to_ticklabels(data.index)
-        elif isinstance(yticklabels, bool) and not yticklabels:
-            yticklabels = ['' for _ in range(data.shape[0])]
+        elif yticklabels is False:
+            yticklabels = []
         else:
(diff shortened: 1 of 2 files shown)


mwaskom:
Cool, thanks!

mwaskom:
Hm looks like this broke the buildbot: http://nipy.bic.berkeley.edu/builders/seaborn-py2.7-wheezy-sparc/builds/163

mwaskom:
```
======================================================================
FAIL: seaborn.tests.test_matrix.TestHeatmap.test_tickabels_off
----------------------------------------------------------------------
  File "/home/buildslave/nd-bb-slave-sparc64-wheezy/seaborn-py2_7-wheezy-sparc/build/venv/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/buildslave/nd-bb-slave-sparc64-wheezy/seaborn-py2_7-wheezy-sparc/build/venv/local/lib/python2.7/site-packages/seaborn-0.7.0.dev-py2.7.egg/seaborn/tests/test_matrix.py", line 189, in test_tickabels_off
    self.df_norm.shape[1])])
AssertionError: Lists differ: [] != ['', '', '', '', '', '', '', '...

Second list contains 8 additional elements.
First extra element 0:


- []
+ ['', '', '', '', '', '', '', '']
```

mwaskom:
Not sure why Travis passed but this didn't; any thoughts @sauliusl ?

sauliusl:
@mwaskom my best guess would be that the buildbot somehow missed commit 7d2fbfb. 
That commit fixes this particular test.

PR #685, opened 2015-08-26
Use `LineCollection` when plotting dendrograms.

sauliusl:
Instead of plotting the dendrogram lines one by one using standard `ax.plot()`, this PR makes use  of the `LineCollection` object from `matplotlib` to draw them all at once.

This has the effect of significantly speeding up the time needed to draw the clustermap.

For instance, the following script:
```python
import numpy as np
import fastcluster
import pandas as pd
import seaborn

ROWS, COLS = 10000, 10
np.random.seed(10)
random_data = np.random.randn(ROWS, COLS)
random_data[ROWS/5: 2*ROWS/5, :] *= 10
random_data = pd.DataFrame(random_data, 
                           index=['Row {}'.format(i) for i in range(ROWS)], 
                           columns=['Column {}'.format(i) for i in range(COLS)])
row_linkage = fastcluster.complete(random_data)
col_linkage = fastcluster.complete(random_data.T)
seaborn.clustermap(random_data,
                   row_linkage=row_linkage,
                   col_linkage=col_linkage, yticklabels=False)

plt.savefig('test.png')
plt.close()
```

Takes 0m35.323s on my computer when dendrogram is drawn as `LineCollection` (note that 10k rows are being plotted!)
Meanwhile code on master branch takes 0m53.243s on the same machine. Meaning we get 30% improvement with the new code.

If you look at the actual diff -- quite a few things had to change because of the new way of plotting.
Most notably, the axes do not autoscale when plotting using `LineCollection`, thus I had to manually put the scaling in. I used a similar approach to scaling that `scipy.cluster.hierarchy` use for their dendrogram plotting.

If we consider the `flights` example, that is given in the documentation:
```python
>>> import seaborn as sns; sns.set()
>>> flights = sns.load_dataset("flights")
>>> flights = flights.pivot("month", "year", "passengers")
>>> g = sns.clustermap(flights)
```
The plots generated are identical, thus I believe I did not screw anything up with this new scaling.

master branch:
flights master

my branch:
flights faster_dendrograms

The time improvement for this example is less severe, of course. 0m2.896s vs 0m2.987s on my machine (including the time to savefig). Nevertheless, it is still an improvement.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -603,13 +603,13 @@ class _DendrogramPlotter(object):
 
         if self.rotate and self.axis == 0:
             ax.yaxis.set_ticks_position('right')
-            ax.invert_yaxis()
 
             # Constants 10 and 1.05 come from _plot_dendrogram in scipy.cluster.hierarchy
             ax.set_ylim(0, number_of_leaves * 10)
             ax.set_xlim(0, max_dependent_coord * 1.05)
 
             ax.invert_xaxis()
+            ax.invert_yaxis()
         else:
             # Constants 10 and 1.05 come from _plot_dendrogram in scipy.cluster.hierarchy
             ax.set_xlim(0, number_of_leaves * 10)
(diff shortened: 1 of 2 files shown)


mwaskom:
Cool!

PR #686, opened 2015-08-26
Update linearmodels.py

dhimmel:
Fixes a bug in lmplot. When `sharex=True`, the `scatter` argument was being overwritten and evaluating as True. Therefore, when lmplot was called with `scatter=False` a scatterplot would be drawn unless `sharex=False`.


mwaskom:
Ah nice catch.

PR #690, opened 2015-08-28
Do not draw ticks at all if ticklabels are disabled

sauliusl:
This PR actually does something I should have done in PR #683 that is already merged.

After PR #685 was merged, I ran the profiler again and noticed that quite peculiarly most of the time is now spent setting `yticks` when plotting a large number of rows. I never thought this would be the bottleneck, but disabling the ticks completely, when no ticklabels are given, brings down the runtime of the same test script as given in #685 to 5.7s (from around 30s, and from around 60s before dendrogram fix...). 

What is even better, `fastcluster` is now the biggest bottleneck of that example, not plotting!
As it should be! Woo-hoo!

Not really sure how I missed this when pushing #683. Anyway, it is an easy fix, and all tests seem to pass, so have yet  another PR (sorry for such a pile-up)!

Again, the plots are identical:
master:
master

PR
pr

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -130,12 +130,22 @@ class _HeatMapper(object):
 
         # Get the positions and used label for the ticks
         nx, ny = data.T.shape
-        xstart, xend, xstep = 0, nx, xtickevery
-        self.xticks = np.arange(xstart, xend, xstep) + .5
-        self.xticklabels = xticklabels[xstart:xend:xstep]
-        ystart, yend, ystep = (ny - 1) % ytickevery, ny, ytickevery
-        self.yticks = np.arange(ystart, yend, ystep) + .5
-        self.yticklabels = yticklabels[ystart:yend:ystep]
+
+        if xticklabels == []:
+            self.xticks = []
+            self.xticklabels = []
+        else:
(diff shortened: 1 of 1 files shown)


seaborn/matrix.py
```diff
@@ -130,12 +130,22 @@ def __init__(self, data, vmin, vmax, cmap, center, robust, annot, fmt,
...
         # Get the positions and used label for the ticks
         nx, ny = data.T.shape
-        xstart, xend, xstep = 0, nx, xtickevery
-        self.xticks = np.arange(xstart, xend, xstep) + .5
-        self.xticklabels = xticklabels[xstart:xend:xstep]
-        ystart, yend, ystep = (ny - 1) % ytickevery, ny, ytickevery
-        self.yticks = np.arange(ystart, yend, ystep) + .5
-        self.yticklabels = yticklabels[ystart:yend:ystep]
+
+        if xticklabels == []:
```
mwaskom: Any reason for this explicit check for equality against an empty list instead of the more pythonic `if not xticklabels` ?

sauliusl: `if not xticklabels` fails if `xticklabels` is a numpy array. Something about truth value being ambiguous for numpy arrays.

mwaskom:
Thanks!

Issue #618, opened 2015-06-30
Fix random failure of logistic regression test


mwaskom:
Probably just poisson variation but it feels like this has been failing more often (weirdly, almost always on the Python 3.4 build).

```python
======================================================================
ERROR: seaborn.tests.test_linearmodels.TestRegressionPlotter.test_logistic_regression
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/testenv/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/testenv/lib/python3.4/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/home/travis/build/mwaskom/seaborn/seaborn/tests/test_linearmodels.py", line 347, in test_logistic_regression
    _, yhat, _ = p.fit_regression(x_range=(-3, 3))
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 194, in fit_regression
    family=Binomial())
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 250, in fit_statsmodels
    n_boot=self.n_boot, units=self.units)
  File "/home/travis/build/mwaskom/seaborn/seaborn/algorithms.py", line 76, in bootstrap
    boot_dist.append(func(*sample, **func_kwargs))
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 244, in <lambda>
    reg_func = lambda _x, _y: model(_y, _x, **kwargs).fit().predict(grid)
  File "/home/travis/miniconda/envs/testenv/lib/python3.4/site-packages/statsmodels/genmod/generalized_linear_model.py", line 719, in fit
    raise PerfectSeparationError(msg)
statsmodels.tools.sm_exceptions.PerfectSeparationError: Perfect separation detected, results not available
```

mwaskom:
Closed with #696 (probably)

PR #696, opened 2015-09-02
use teardown method in PlotTestCase for most tests

### LINKED ISSUE #694: Use teardown methods for tests
```
~/dev/seaborn/seaborn/tests $ grep "plt.close" *.py | wc
     241     482   10404
```
That's a lot of closing going on. This could be reduced through the use of [teardown methods](http://nose.readthedocs.org/en/latest/writing_tests.html).

Just define this class and have everything inherit from it.
```
class SeabornTestCase(object):
    def tearDown(self):
        plt.close('all')
```

@mwaskom let me know if you're interested and I can send a PR. Should be simple.

clarkfitzg:
Closes #694.

After this we have:
```
~/dev/seaborn/seaborn/tests $ grep "plt.close" *.py | wc
      98     196    4323
```
So `plt.close` shows up 98 times rather than 241. There are still so many because some tests call `plt.close` several times within one test method. Those can stay.

These tests are now more robust, since the `tearDown` method is called regardless of whether the test causes an unexpected exception.

### DIFF
--- a/seaborn/tests/__init__.py
+++ b/seaborn/tests/__init__.py
@@ -1,7 +1,11 @@
+import numpy as np
 from matplotlib.pyplot import close
 
 
 class PlotTestCase(object):
+
+    def setUp(self):
+        np.random.seed(49)
 
     def tearDown(self):
         close('all')
--- a/seaborn/tests/__init__.py
+++ b/seaborn/tests/__init__.py
@@ -1,0 +1,7 @@
+from matplotlib.pyplot import close
+
+
+class PlotTestCase(object):
+
+    def tearDown(self):
+        close('all')
--- a/seaborn/tests/test_axisgrid.py
+++ b/seaborn/tests/test_axisgrid.py
@@ -12,6 +12,7 @@ import numpy.testing as npt
 from numpy.testing.decorators import skipif
 import pandas.util.testing as tm
 
+from . import PlotTestCase
 from .. import axisgrid as ag
 from .. import rcmod
 from ..palettes import color_palette
@@ -25,7 +26,7 @@ rs = np.random.RandomState(0)
 old_matplotlib = LooseVersion(mpl.__version__) < "1.4"
 
 
-class TestFacetGrid(object):
+class TestFacetGrid(PlotTestCase):
 
     df = pd.DataFrame(dict(x=rs.normal(size=60),
                            y=rs.gamma(4, size=60),
@@ -38,21 +39,17 @@ class TestFacetGrid(object):
 
         g = ag.FacetGrid(self.df)
         nt.assert_is(g.data, self.df)
-        plt.close("all")
 
     def test_self_fig(self):
 
         g = ag.FacetGrid(self.df)
         nt.assert_is_instance(g.fig, plt.Figure)
-        plt.close("all")
 
     def test_self_axes(self):
 
         g = ag.FacetGrid(self.df, row="a", col="b", hue="c")
         for ax in g.axes.flat:
             nt.assert_is_instance(ax, plt.Axes)
-
-        plt.close("all")
 
     def test_axes_array_size(self):
 
@@ -73,8 +70,6 @@ class TestFacetGrid(object):
 
         for ax in g5.axes.flat:
             nt.assert_is_instance(ax, plt.Axes)
-
-        plt.close("all")
 
     def test_single_axes(self):
 
@@ -116,8 +111,6 @@ class TestFacetGrid(object):
         g_missing_wrap = ag.FacetGrid(df, col="d", col_wrap=4)
         nt.assert_equal(g_missing_wrap.axes.shape, (9,))
 
-        plt.close("all")
-
     def test_normal_axes(self):
 
         null = np.empty(0, object).flat
@@ -150,8 +143,6 @@ class TestFacetGrid(object):
         npt.assert_array_equal(g._not_left_axes, g.axes[:, 1:].flat)
         npt.assert_array_equal(g._inner_axes, g.axes[:-1, 1:].flat)
 
-        plt.close("all")
-
     def test_wrapped_axes(self):
 
         null = np.empty(0, object).flat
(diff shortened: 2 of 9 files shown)


mwaskom:
Thanks, looks like the test failure is just some pesky pep8 rules.

mwaskom:
> So plt.close shows up 98 times rather than 241. There are still so many because some tests call plt.close several times within one test method. Those can stay.

If those are bothering you, those tests can probably be modified to explicitly create a figure for each different plot and pass the axes along.

clarkfitzg:
Alternatively they could just be broken out into smaller tests. But it's not a big deal- they're fine to leave.

clarkfitzg:
Interesting error for only the Python 3.3 build. 

```
ERROR: seaborn.tests.test_linearmodels.TestRegressionPlotter.test_logistic_regression
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/testenv/lib/python3.3/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/home/travis/build/mwaskom/seaborn/seaborn/tests/test_linearmodels.py", line 348, in test_logistic_regression
    _, yhat, _ = p.fit_regression(x_range=(-3, 3))
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 194, in fit_regression
    family=Binomial())
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 250, in fit_statsmodels
    n_boot=self.n_boot, units=self.units)
  File "/home/travis/build/mwaskom/seaborn/seaborn/algorithms.py", line 76, in bootstrap
    boot_dist.append(func(*sample, **func_kwargs))
  File "/home/travis/build/mwaskom/seaborn/seaborn/linearmodels.py", line 244, in <lambda>
    reg_func = lambda _x, _y: model(_y, _x, **kwargs).fit().predict(grid)
  File "/home/travis/miniconda/envs/testenv/lib/python3.3/site-packages/statsmodels/genmod/generalized_linear_model.py", line 719, in fit
    raise PerfectSeparationError(msg)
statsmodels.tools.sm_exceptions.PerfectSeparationError: Perfect separation detected, results not available
----------------------------------------------------------------------
Ran 405 tests in 280.423s
FAILED (SKIP=6, errors=1)
```

mwaskom:
Just not enough random data: https://github.com/mwaskom/seaborn/issues/618

clarkfitzg:
Seems a little strange to me that it only occasionally fails when the random number generators are seeded. Bootstrapping uses random data too. Also unnittests run in basically a random order.

clarkfitzg:
So along the same lines as this PR, what if we had this:

```
class PlotTestCase(object):
    def setUp(self):
         np.random.seed(49)

    def tearDown(self):
        close('all')
```

I wonder if that would help with getting all the tests to run more consistently.

mwaskom:
> Seems a little strange to me that it only occasionally fails when the random number generators are seeded. Bootstrapping uses random data too.

Right, the bootstrapping isn't seeded, and I think that error happens when one of the bootstrap samples achieves perfect separation.

clarkfitzg:
Looking at the test that occasionally fails, it doesn't seed the generator.

clarkfitzg:
> that error happens when one of the bootstrap samples achieves perfect separation

Exactly, because logistic regression can't work for that case.

mwaskom:
Yes. I understand the problem, I just haven't gotten around to fixing it because it is always easier to just restart travis :)

clarkfitzg:
Latest commit will seed the generator with the same value before every test in classes inheriting from `PlotTestCase`. I'd expect that the tests involving random things will either pass or fail every time now.

clarkfitzg:
This is ready to go from my end. I believe it also closes #618.

mwaskom:
Thanks!

mwaskom:
Just ran the tests locally and saw this warning, which I hadn't noticed in the travis output:

```
/Users/mwaskom/anaconda/envs/pandas_dev/lib/python2.7/site-packages/matplotlib/pyplot.py:424: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
```

I don't recall seeing this before the change in this PR.

Is it possible you missed some fixtures?

clarkfitzg:
@mwaskom I get the same results. But this was here before this PR. You can verify by checking out a previous commit and running the tests:

```
git checkout 99e54e19
```

Running nose without `doctest` the warning doesn't come up:

```
(seaborn)testimage ~/dev/seaborn $ nosetests
/Users/clark.fitzgerald/anaconda/envs/seaborn/lib/python2.7/site-packages/pandas/core/indexing.py:115: SettingWithCopyWarni
ng:
A value is trying to be set on a copy of a slice from a DataFrame

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-co
py
  self._setitem_with_indexer(indexer, value)
................./Users/clark.fitzgerald/anaconda/envs/seaborn/lib/python2.7/site-packages/matplotlib/tight_layout.py:225:
UserWarning: tight_layout : falling back to Agg renderer
  warnings.warn("tight_layout : falling back to Agg renderer")
.........S.................................................................................................................
.............................................................................................................F.......F..SS.
.................F...F...................................................................................SSS
```

This leads me to believe that the large number of figures are from the doctests. So maybe you could look into [doctest fixtures](http://nose.readthedocs.org/en/latest/doc_tests/test_doctest_fixtures/doctest_fixtures.html) to do a similar thing as this PR. However, it doesn't seem to be hurting anything.

mwaskom:
Interesting. I doubt it is the doctests, though, because of the option I'm using for `plot_directive`:

```
        .. plot::
            :context: close-figs
```

mwaskom:
Smart to actually check against the old version, though :)

PR #697, opened 2015-09-03
Fixed an issue with factorplot when plotting data that contains different categories across columns

dimarkov:
Using the factorplot for plots with categorical data that contain different categories across columns (e.g. first column has categories 'A', 'B', 'C', and second column has categories 'A', 'B', 'D') would add all the categories to all subplots (e.g. all plots would have 'A', 'B', 'C', 'D' categories in spite of missing data). Hence, I have added a re-computation of the "order" variable (lines 793 - 796) for each subplot, if the sharex is set to false (Setting sharex = True already assumes that the data in all subplots contains same categories). I have also introduced a new private variable self._sharex.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -365,6 +365,7 @@ class FacetGrid(Grid):
         self._y_var = None
         self._dropna = dropna
         self._not_na = not_na
+        self._sharex = sharex
 
         # Make the axes look good
         fig.tight_layout()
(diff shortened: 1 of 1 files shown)


mwaskom:
Sorry, but this does not work. It will break any code that tries to use `FacetGrid` with a plotting function that does not accept an `order` parameter.

dimarkov:
Sorry I didn't realize that. I looked only at factorplot.

Would you accept the change if I find a way to make it work?
For example, the code could test first if kwargs contains the order
parameter before recomputing it.


On Thu, Sep 3, 2015 at 10:34 PM, Michael Waskom <notifications@github.com>
wrote:

> Closed #697 <https://github.com/mwaskom/seaborn/pull/697>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mwaskom/seaborn/pull/697#event-400516697>.
>



-- 
Anyone who has never made a mistake has never tried anything new.

mwaskom:
No, I want to keep that kind of thing out of `FacetGrid`.

PR #701, opened 2015-09-11
Fix for plotting series of only one point

wjaskowski:
If interpolate=true and len(x)=1 padding the sides of the plot might caused an out of bounds exception (note that after the fix, it still can when interpolate=false)


mwaskom:
Is a timeseries with one point like the sound of one hand clapping?

wjaskowski:
Well, if you ask whether that happens in practice, then my answer is "yes" - it happened to me many times before I realized what is the problem. The fix is simple and does not break anything.

mwaskom:

Issue #713, opened 2015-09-28
plt.scatter and seaborn.pointplot use different pluralizations for their markers parameter


cviner:
In the [`FacetGrid`](http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.FacetGrid.html) documentation's `hue_kws` marker-varying example, the keyword argument is erroneously provided in singular, instead of plural form, which makes the keyword argument not be recognized and therefore have no effect. That is, `hue_kws=dict(marker=["^", "v"]))` is used, but it should be `hue_kws=dict(markers=["^", "v"]))`; i.e. `marker` -> `markers`.

cviner:
Unfortunately, it seems that this is indeed correct for `plt.scatter` in the example, but does fail for Seaborn's [`pointplot`](http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.pointplot.html), since it uses `markers`, but not `marker`.

It might be worthwhile for these to be made consistent.

cviner:
The difference is [because `pointplot` accepts a list of markers itself, whereas `plt.scatter` only takes a single marker](https://github.com/mwaskom/seaborn/pull/714#issuecomment-143611975).

PR #714, opened 2015-09-28
Trivial typo fix in example (fix keyword pluralization)

cviner:
Resolves issue #713, by fixing the keyword argument for the specification of markers from [`FacetGrid`](http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.FacetGrid.html) via the `hue_kws` parameter. The keyword `marker` was erroneously provided in singular, instead of plural form (i.e. `markers`).

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -535,7 +535,7 @@ class FacetGrid(Grid):
 
             >>> g = sns.FacetGrid(tips, col="sex", hue="time", palette=pal,
             ...                   hue_order=["Dinner", "Lunch"],
-            ...                   hue_kws=dict(marker=["^", "v"]))
+            ...                   hue_kws=dict(markers=["^", "v"]))
             >>> g = (g.map(plt.scatter, "total_bill", "tip", **kws)
             ...      .add_legend())
 


cviner:
Unfortunately, it seems that this is indeed correct for `plt.scatter` in the example, but does fail for Seaborn's [`pointplot`](http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.pointplot.html), since it uses `markers`, but not `marker`.

Apologies for the invalid pull request.

mwaskom:
Ah, but the difference is, you can actually pass multiple markers to `pointplot`, hence the plural.

cviner:
But doesn't the example depict the passing of a list of two markers to `plt.scatter`?

cviner:
Yes, that does make sense and a good point. I guess my question now would be how to use `hue_kws` to pass markers per hue to `pointplot`. Should I nest lists?

mwaskom:
The key name you pass to `FacetGrid` is just whatever the parameter is called in the function(s) you want to map. In `scatter` there is a `marker` kwarg, which takes a single value. In pointplot there is a `markers` kwarg because you can pass multiple markers. See the fourth example here: http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.pointplot.html#seaborn.pointplot

cviner:
Yes, this makes sense now. Sorry for my confusion with this.

Is there an easy way for me to set the markers per hue using `FacetGrid` with a `pointplot`?

mwaskom:
Well the easiest way is just to use `factorplot`, which combines `FacetGrid` and `pointplot` for you. Pass a list of markers to `markers`; it should have the same length as your palette.

cviner:
Yes, I realize that `factorplot` would likely be easier. I sort of like the separation, but also ran into a problem with `factorplot` but not `FacetGrid`, which was difficult to make a MWE to reproduce. I do plan to refactor to use `factorplot` at some point.

I tried doing that, but found that it seems to set all markers to the first list element. It should have the same length as the number of legend labels, not the palette though, correct (since my palette can have fewer elements and then it cycles)?

Thanks for your patience with this.

mwaskom:
If you're going to use `FacetGrid` directly you'll need to make sure you pass a `hue_order`. Otherwise the colors will correspond to the hue levels in the order they appear in each chunk of the DataFrame, which is not guaranteed to be the same across facets. The same is true of `order`, come to think of it. This is part of why `factorplot` exists. Anyway, provided that you're being careful and know the DataFrame is sorted properly, you could do, e.g.

```python
g = sns.FacetGrid(tips, hue="smoker", col="sex",
                  hue_kws={"markers": ["o", "v"]})
g.map(sns.pointplot, "day", "total_bill")
```

cviner:
I didn't realize that. That explains some of the problems I've been having. Perhaps I will switch to `factorplot` earlier than I originally planned. Thanks for these details!

mwaskom:
Keep in mind that `factorplot` returns the `FacetGrid` object it's plotting onto in case you need access to it for other reasons.

cviner:
Yes, I do indeed need to take advantage of that fact. Thanks for all your help.

Issue #719, opened 2015-09-30
FacetGrid gives UnicodeEncodeError when data contains utf-8 characters


carnby:
When labeling facets or labels in grids on the IPython notebook (using Python 2), an UnicodeEncodeError is raised.

Minimal example:

```
import seaborn as sns
import matplotlib.pyplot as plt
%matplotlib inline
data = sns.load_dataset('anscombe')
data['dataset_unicode'] = data.dataset.map(lambda x: u'ÁéÍóÚ {0}'.format(x))
# works
g = sns.FacetGrid(data, col='dataset')
g.map(plt.scatter, 'x', 'y')
# doesn't work
g = sns.FacetGrid(data, col='dataset_unicode')
g.map(plt.scatter, 'x', 'y')
```

This raises the following exception:
```
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-20-03ef1de0f907> in <module>()
     10 # doesn't work
     11 g = sns.FacetGrid(data, col='dataset_unicode', col_wrap=2)
---> 12 g.map(plt.scatter, 'x', 'y')

/home/egraells/envs/ipython/local/lib/python2.7/site-packages/seaborn/axisgrid.pyc in map(self, func, *args, **kwargs)
    727 
    728         # Finalize the annotations and layout
--> 729         self._finalize_grid(args[:2])
    730 
    731         return self

/home/egraells/envs/ipython/local/lib/python2.7/site-packages/seaborn/axisgrid.pyc in _finalize_grid(self, axlabels)
    817         """Finalize the annotations and layout."""
    818         self.set_axis_labels(*axlabels)
--> 819         self.set_titles()
    820         self.fig.tight_layout()
    821 

/home/egraells/envs/ipython/local/lib/python2.7/site-packages/seaborn/axisgrid.pyc in set_titles(self, template, row_template, col_template, **kwargs)
    959             for i, col_name in enumerate(self.col_names):
    960                 args.update(dict(col_name=col_name))
--> 961                 title = template.format(**args)
    962                 # Index the flat array so col_wrap works
    963                 self.axes.flat[i].set_title(title, **kwargs)

UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-4: ordinal not in range(128)
```

The same happens when using the hue parameter. For instance, this is the error (copy-paste from a real situation):
```
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
[...]

PR #724, opened 2015-10-07
Ensures compatibility of axisgrid labels with unicode objects.

carnby:
This PR makes the example from #719 to work without problems.

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -527,3 +527,29 @@ def categorical_order(values, order=None):
                     order = order
         order = filter(pd.notnull, order)
     return list(order)
+
+
+def to_unicode_or_bust(obj, encoding='utf-8'):
+    """
+    Ensures convertion of an object to unicode.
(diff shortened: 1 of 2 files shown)


seaborn/utils.py
```diff
@@ -527,3 +527,29 @@ def categorical_order(values, order=None):
                     order = order
         order = filter(pd.notnull, order)
     return list(order)
+
+
+def to_unicode_or_bust(obj, encoding='utf-8'):
```
mwaskom: "or_bust" suggests that an error will be raised if it fails, but I don't see that in the code. Maybe just "to_unicode" then?

carnby: Actually this is a "classic" function which I copied (the docstring credits the source). I didn't think much about the name.

But it will fail if the object includes invalid characters according to the specified encoding.

mwaskom:
Cloded with #885

PR #730, opened 2015-10-11
Fix typos and improve documentation consistency

tlnagy:
Fixed a couple typos and inconsistencies in the distribution docs

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -67,7 +67,7 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
         Color to plot everything but the fitted curve in.
     vertical : bool, optional
         If True, oberved values are on y-axis.
-    norm_hist : bool, otional
+    norm_hist : bool, optional
         If True, the histogram height shows a density rather than a count.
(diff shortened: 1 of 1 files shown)


seaborn/distributions.py
```diff
@@ -465,18 +465,18 @@ def kdeplot(data, data2=None, shade=False, vertical=False, kernel="gau",
...
-    shade_lowest : bool
+    shade_lowest : bool, optional
         If True, shade the lowest contour of a bivariate KDE plot. Not
         relevant when drawing a univariate plot or when ``shade=False``.
         Setting this to ``False`` can be useful when you want multiple
         densities on the same Axes.
     ax : matplotlib axis, optional
         Axis to plot on, otherwise uses current axis.
-    kwargs : key, value pairings
+    kwargs : key, value pairs
```
mwaskom: If we're being pedantic, "pairs" suggests a tuple with two items.

tlnagy: I was trying to decide whether to use pairings, mappings, or pairs as all three where used in the docs previously, but I agree. Changing to pairings.

PR #764, opened 2015-11-14
implement err_style='std' and err_style='sem' for timeseries plots

astraw:
Hi,

We use the timeseries plots extensively here (thanks for that!) and it would be useful to plot the standard deviation and standard error of the mean using the same interface.

Would something like this be within-scope for seaborn?

### DIFF
--- a/seaborn/timeseries.py
+++ b/seaborn/timeseries.py
@@ -57,7 +57,8 @@ def tsplot(data, time=None, unit=None, condition=None, value=None,
         legend (unless legend is set to False).
     err_style : string or list of strings or None
         Names of ways to plot uncertainty across units from set of
-        {ci_band, ci_bars, boot_traces, boot_kde, unit_traces, unit_points}.
+        {ci_band, ci_bars, boot_traces, boot_kde, unit_traces, unit_points,
+        std, sem}.
         Can use one or more than one method.
     ci : float or list of floats in [0, 100]
         Confidence interval size(s). If a list, it will stack the error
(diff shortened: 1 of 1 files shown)


sdvillal:
This is also discussed in, for example, #643

mwaskom:
This doesn't really work with the API of `tsplot`. In principle, *what* is shown should be orthogonal to *how* it is shown. Currently "what" only includes bootstrapped estimates of the sampling distribution, which can be plotted ("how") with bands, bars, a density image, etc. The problem with this change is that it would be very reasonable for someone to want `tsplot` to show the standard deviation, but with bars instead of bands. So would you add `std_bars`? That quickly becomes a messy hack.

There are also many other places in seaborn where bootstrapped confidence intervals are drawn that one might also reasonably want standard deviations (or some other prediction interval). So ideally, one would only want to add that functionality to certain functions in a way that it could be similarly added to other ones, even if not simultaneously.

Anyway, the point is that this is a trickier issue, from an API perspective, than this PR appreciates.

Also, see my earlier comments about  not wanting to change much about `tsplot` until it can be refactored (although, to be honest, that's not something I am likely to have time for in the near future).

astraw:
It seems to me that the args to `tsplot()` already specify both what is plotted and how it looks. (And not just bootstrapped estimates of sampling distribution -- also the original data, for example.) 

So, would you be enthusiastic about a small refactor such that there is a `band` vs `bar` plotter (how) used by `ci`, `std`, or `sem` (what)?

In principle the how and what could be made more explicitly separate, but that would seem to induce API breakage because I think they are already mixed (ci_bars vs ci_bands). So, I would just add std_bars, std_bands, sem_bars, sem_bands for consistency. I wouldn't call that a "messy hack".

mwaskom:
I am inclined against adding messy complexity to allow plots that are easily drawn directly in matplotlib.

mwaskom:
Also I'd point out that if you want to see the sem, you should just plot the 68% CI

astraw:
If this is a no-go for seaborn, the messy complexity will then be in our code (perhaps a seaborn wrapper), as we really do want to switch between CIs and STD or SEM.

I also don't think it's a matter of easily drawing directly in matplotlib -- seaborn has nice styles and pandas groupby functionality that are not trivial.

(Regarding SEM, I would like be technically correct if I write "standard error of the mean" -- as far as I'm aware, the 68% confidence interval around the mean found using bootstrapping is only an estimate of this value.)

For reference, I pushed a new commit which implements std_bands, std_bars, sem_bands, and sem_bars. The bands and bars implementations are shared also with ci_bands and ci_bars, respectively. There could almost certainly be some redundancy taken out of the implementation. If this has a shot of making it in, let me know and I'll try slimming this down.

mwaskom:
Closing with tsplot deprecation (but see #1208)

Issue #774, opened 2015-11-18
Turning axes_style and plotting_context into contextdecorators


anntzer:
Python3's `contextmanager` decorator actually creates contextdecorators, i.e. they can be used as decorators to apply themselves to the whole decorated function.
I believe this same functionality can be easily added to seaborn's `axes_style` and `plotting_context` simply by adding a `__call__` method to the `_AxesStyle` and `_PlottingContext` classes (and would be useful to have).
Thoughts?

mwaskom:
So the idea would be that you would have in your codebase something like

```python
@sns.axes_style("ticks")
def my_plot(ax, ...):
    ...
```

and then any time you call `my_plot` it will make a plot with the `ticks` style?

anntzer:
Yes.

mwaskom:
Sounds cool! What does the `__call__` method have to do? Do you want to put together a PR?

anntzer:
Something like (untested)
```
def __call__(self, func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        with self: return func(*args, **kwargs)
    return wrapper
```
I have a few other things on my plate right now so the PR will have to wait if you want me to do it :-)

mwaskom:
Closed with #775

PR #775, opened 2015-11-21
AxesStyle & PlottingContext as contextdecorators.

anntzer:
cf #774.

### DIFF
--- a/seaborn/rcmod.py
+++ b/seaborn/rcmod.py
@@ -1,5 +1,6 @@
 """Functions that alter the matplotlib rc dictionary on the fly."""
 from distutils.version import LooseVersion
+import functools
 
 import numpy as np
 import matplotlib as mpl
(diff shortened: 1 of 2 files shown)


anntzer:
(also hijacking this discussion to inquire about whether there are plans for a future release soon to avoid warnings with matplotlib 1.5.0)

seaborn/rcmod.py
```diff
@@ -460,6 +433,35 @@ def set_context(context=None, font_scale=1, rc=None):
     mpl.rcParams.update(context_object)
 
 
+class _StyleOrContext(dict):
```
mwaskom: I think a better name for this would be `_RCAesthetics`; less constraining for the future.

seaborn/rcmod.py
```diff
@@ -460,6 +433,35 @@ def set_context(context=None, font_scale=1, rc=None):
...
+    def __enter__(self):
+        rc = mpl.rcParams
+        self._orig = {k: rc[k] for k in self._keys}
+        type(self)._set(self)
```
mwaskom: This seems overly complicated. Why not just `self._set()`? If you want, you can add a `_set` method to the base class that raises `NotImplementedError` to make it clear that derived classes need it. But I'm not aware of any reason you'd need to use `type` here.

seaborn/rcmod.py
anntzer: Sure, I don't really care :-)

seaborn/rcmod.py
anntzer: Because `self._set` would treat (on Py3) `_set` as a method and fill in the first argument with `self`, see for example
```
In [1]: def f(*args): print(args)

In [2]: class C:
    method = f
   ...: 

In [3]: C().method()
(<__main__.C object at 0x7f9538812d30>,)
```
Either I need to call via `type` (to avoid binding self to the first argument), or I can do `_set = staticmethod(set_keys)` further down -- either way is fine with me.

mwaskom: Ah I see, it's because you're calling a a non-class function stored as an attribute. Hm. I think that the `staticmethod` approach is a little bit more clear.

anntzer: Changed.  Apparently that was needed on Python2 anyways.

mwaskom:
OK look great, thanks!

Issue #756, opened 2015-11-07
Workaround for violinplot edge line color issue


tritemio:
Following up from the latest comments on #678.

A bug in matplotlib (https://github.com/matplotlib/matplotlib/issues/5423) causes erratic edge line color in violinplot in seaborn under python 3.

As a temporary workaround I call `patch_violinplot()` after a call to `seaborn.violinplot()`

```
def patch_violinplot():
    """Patch seaborn's violinplot in current axis to workaround matplotlib's bug ##5423."""
    from matplotlib.collections import PolyCollection
    ax = plt.gca()
    for art in ax.get_children():
        if isinstance(art, PolyCollection):
            art.set_edgecolor((0.3, 0.3, 0.3))
```

But it looks like seaborn can also avoid the issue using `facecolor` instead of `color` in (which is apparently a more "robust" keyword):

https://github.com/mwaskom/seaborn/blob/master/seaborn/categorical.py#L798

PR #782, opened 2015-11-25
Use facecolor instead color keyword argument to avoid random behaviour

sebp:
This is a workaround the random behavior with python 3.4 as described in bug #756


mwaskom:
Thanks!

PR #788, opened 2015-11-30
Force color of plot in grid diagonal

yoavram:
Allows setting the `color` kwarg in `map_diag`, which then overrides seaborn's attempt to set the colors via the palette.
Example:

``` py
#1
g = sns.PairGrid(data)
g.map_diag(plt.hist, color=color)
#2
g.map_diag(sns.distplot, color=color)
```

The first (`plt.hist`) was possible before by giving `facecolor=color`; the second (`sns.distplot`) wasn't possible AFAIK.

### DIFF
--- a/seaborn/tests/test_axisgrid.py
+++ b/seaborn/tests/test_axisgrid.py
@@ -781,6 +781,29 @@ class TestPairGrid(PlotTestCase):
             nt.assert_equal(len(ax.patches), 40)
 
     @skipif(old_matplotlib)
+    def test_map_diag_color(self):
+        color_set = {mpl.colors.cnames[x].lower()
+                     for x in ['red', 'white', 'black']}
+
+        g1 = ag.PairGrid(self.df)
+        g1.map_diag(plt.hist, color='red')
+
+        for ax in g1.diag_axes:
+            colors = [mpl.colors.rgb2hex(patch.get_facecolor()).lower()
+                      for patch in ax.patches]
+            for color in colors:
+                nt.assert_true(color in color_set, color)
+
+        g2 = ag.PairGrid(self.df)
+        g2.map_diag(kdeplot, color='red')
+
+        for ax in g2.diag_axes:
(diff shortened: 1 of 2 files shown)


seaborn/tests/test_axisgrid.py
```diff
@@ -781,6 +781,29 @@ def test_map_diag(self):
...
+            colors = [mpl.colors.rgb2hex(patch.get_facecolor()).lower()
+                      for patch in ax.patches]
+            for color in colors:
+                nt.assert_true(color in color_set, color)
+
+        g2 = ag.PairGrid(self.df)
+        g2.map_diag(kdeplot, color='red')
+
+        for ax in g2.diag_axes:
+            colors = [mpl.colors.rgb2hex(patch.get_facecolor()).lower()
```
mwaskom: `kdeplot` draws lines, so this isn't testing the right thing. (It passes because `ax.patches` is an empty list, I guess).

seaborn/tests/test_axisgrid.py
```diff
@@ -781,6 +781,29 @@ def test_map_diag(self):
...
+                     for x in ['red', 'white', 'black']}
+
+        g1 = ag.PairGrid(self.df)
+        g1.map_diag(plt.hist, color='red')
+
+        for ax in g1.diag_axes:
+            colors = [mpl.colors.rgb2hex(patch.get_facecolor()).lower()
+                      for patch in ax.patches]
+            for color in colors:
+                nt.assert_true(color in color_set, color)
```
mwaskom: I'm confused about why this isn't just testing if `color` is `red`. Why would it be black or white?

yoavram: Well I think that both of your questions are tied together; I must have tested for the wrong thing, and the patches have the line (which should be red) and other objects (which are black and white).
If you can suggest a better test for this let me know, I don't have experience testing plotting libraries.

PR #793, opened 2015-12-04
PERF: Speed up test for empty facet (trivial)

ischwabacher:
Using the cached `size` property on an `ndarray` must be faster than converting the array to a list, and it's more obvious what's going on to boot.

I can't imagine this makes much difference, but it might speed up plotting large datasets.


mwaskom:
Looks good, thanks!

PR #804, opened 2015-12-20
Add swarmplot function (beeswarm plot)

mwaskom:
Swarmplot draws a plot like `stripplot`, but the position of the points are adjusted along the categorical axis so that the points don't overlap. This provides a better representation of the distribution for small-medium sized data. This is also often called a "beeswarm plot". Naturally, the plot is based on seaborn's general categorical plot API, works as a`kind` in `factorplot`, etc.

Here is a simple example:

```python
sns.swarmplot(x="day", y="total_bill", data=tips)
```
image

And an example showing hue-nesting:

```python
sns.swarmplot(x="class", y="age", hue="survived", data=titanic, palette=[".75", "b"])
```
titanic_swarmplot

This PR also adds some general code for categorical scatterplots that will end up relieving a few issues with `stripplot`, but changes to `stripplot` will probably be done in a different PR.

### Things that would be nice to have but won't be in this PR

- More options for what to do with points that extend too far along the categorical axis
- Clever default point size to optimize the use of categorical axis space

### Todo

- [x] Update release notes
- [x] Update categorical tutorial
- [x] Add example gallery script
- [x] Add cross-references in other categorical plot docstrings


seaborn/categorical.py
```diff
@@ -1129,15 +1168,228 @@ def plot(self, ax, kws):
...
+        """Return a list of all swarm points that could overlap with target.
+
+        Assumes that we are working through a sorted storm.
+        """
+        _, y_i = xy_i
+        neighbors = []
+        for xy_j in swarm:
+            _, y_j = xy_j
+            if (y_i - y_j) < d:
+                neighbors.append(xy_j)
```
Carreau: If you assume sorted, (y increasing), you should be able to break out of the `for`as soon as `y_j > y_i +d` , which would save a bunch of loops ? No ?

seaborn/categorical.py
```diff
@@ -1129,15 +1168,228 @@ def plot(self, ax, kws):
...
+        # Set object attributes
+        self.split = split
+        self.width = .8
 
-    def plot(self, ax):
+    def overlap(self, xy_i, xy_j, d):
+        """Return True if two circles with the same diameter will overlap."""
+        x_i, y_i = xy_i
+        x_j, y_j = xy_j
+        return np.linalg.norm([x_i - x_j, y_i - y_j]) < d
```
Carreau: `(x_i - x_j)**2 + (y_i - y_j)**2 < d**2` will likely be much faster, `np.linalg.norm` seem overkill for a norm2 comparison. Unless I misunderstand the usage `np.linalg.norm`. Though you might be able, if you use numpy to vectorize `prune_candidates`, which might be faster.

seaborn/categorical.py
```diff
@@ -1129,15 +1168,228 @@ def plot(self, ax, kws):
...
+            if left_first:
+                new_candidates = [cl, cr]
+            else:
+                new_candidates = [cr, cl]
+            candidates.extend(new_candidates)
+            left_first = not left_first
+        return candidates
+
+    def prune_candidates(self, candidates, neighbors, d):
+        """Remove candidates from the list of they overlap with the swarm."""
```
Carreau: `of` -> `if`

seaborn/categorical.py
mwaskom: No, because `swarm` in this function is only the points that have already been plotted, so working through a sorted list means that you'll never actually encounter a `y_j` that is greater than the `y_i`.

Carreau: Hum, indeed, that's why there is no `abs()`. So loop `reversed(swarm)` and break at `y_j < y_i - d` maybe instead of comparing all ?

mwaskom: Nice idea

Carreau: Even wonderring if the following wouldn't be equivalent:

```
for j,xy_j in enumerate(swarm):
      _, y_j = xy_j
      if (y_i - y_j) < d:
          return swarm[j:]
```

Carreau: Or you se the idea.

mwaskom: Probably slower just because I would expect len(neighbors) << len(swarm) for most plots.

Carreau: I trust you with that, still trying to get the all thing in my head.

seaborn/tests/test_categorical.py
```diff
@@ -1532,6 +1594,167 @@ def test_three_strip_points(self):
...
+class TestSwarmPlotter(CategoricalFixture):
+
+    default_kws = dict(x=None, y=None, hue=None, data=None,
+                       order=None, hue_order=None, split=False,
+                       orient=None, color=None, palette=None)
+
+    def test_overlap(self):
+
+        p = cat._SwarmPlotter(**self.default_kws)
+        nt.assert_false(p.overlap((0, 0), (1, 1), np.sqrt(2)))
```
Carreau: Test failing because sqrt(2)**2 = 2.0000004 ... you can use (0,0), (3,4) which should make an exact distance of 5 betwwen the 2 circles center...

mwaskom: > http://www.smbc-comics.com/index.php?db=comics&id=2999

:)

Issue #763, opened 2015-11-14
[Feature Request] Show SD in seaborn.tsplot() instead of SEM


cachitas:
This issue follows a question posted previously [in stackoverflow](http://stackoverflow.com/questions/33696588/how-to-show-sd-in-seaborn-tsplot-instead-of-sem?noredirect=1#comment55169589_33696588).

Do you think it is possible to implement it in the future? In #450, you say:

> tsplot internals are poorly implemented and untested, so any enhancement is probably going to have to wait until it can be refactored

Could it be done having this feature in mind? How can I help?

mwaskom:
Closed with #1208

PR #814, opened 2016-01-03
[WIP] Refactoring tsplot

caichinger:
Hi,

to begin with, I want to thank you for developing such a great library! :)

Since I am interested in an extended version of `tsplot`, to visualize for example observational quantiles, I took a look at a number of issues:
- Extend tsplot to plot observation quantiles? #450
- adds ability to visualize distribution of data #643
- [Feature Request] Show SD in seaborn.tsplot() instead of SEM #763 
- implement err_style='std' and err_style='sem' for timeseries plots #764 
- ...

From the respective discussions and the current implementation of `tsplot` I understand the need for refactoring and providing tests to allow future developments.

I would like to contribute to your project and want to propose this PR as a **basis for a discussion** on how to possibly refactor and test `tsplot`.

Regarding the PR:
- For sake of simplicity, I created two new modules `timeseries_new` and `test_timeseries_new` instead of touching the original `timeseries` implementation since the old and new implementation are very different in their design. When we know how to proceed, the modules need to be renamed or their code being copied somewhere else.
- The code is the same as in the original `tsplot` (in particular there are no new features) but broken down into smaller pieces and formed into a class to facilitate testing (similar to other implementations in seaborn). While refactoring, I tried to stick to the design of the `categorical` module.
- The API and the function-interface is of course still the same but the main plotting task is now handled by the new `_TimeSeriesPlotter` class.
- `_TimeSeriesPlotter` has a number of methods to prepare and compute the data needed for plotting, e.g. `_init_from_df`, `_init_from_array`, `_set_up_color_palette`, `_compute_plot_data`, `_err_color`. The only public method is `plot` which carries out the work of drawing the figure.
- The subroutines for error-plotting such as `_plot_ci_band` are unchanged.
- Note that the majority of the tests deal with the refactored version but there are of course also some tests inspecting the produced figures (which are passed by both the old and new version...).

I am looking forward to hearing your opinion on this topic!


caichinger:
Builds work fine for
  - PYTHON=2.7 DEPS=modern
  - PYTHON=3.4 DEPS=modern

but there are two issues with
  - PYTHON=2.7 DEPS=minimal

1. AttributeError: 'DataFrame' object has no attribute 'sort_values': [pandas.DataFrame.sort_values](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.sort_values.html#pandas.DataFrame.sort_values) was only introduced in *version 0.17.0*. Which pandas version is used in this setting?
2. There is a problem with numpy.testing.assert_allclose but I do not understand why. Any hints?

mwaskom:
Closing with tsplot deprecation (#1217)

Issue #830, opened 2016-01-21
Annotation in heatmaps are not showing up


gkunter:
In my installation (seaborn 0.6.0, matplotlib 1.5.1 with Qt2 back end under Linux Mint, Python 3.5), the annotation feature of heatmaps doesn't work. When I run the "Annotated heatmap" example from the [project website](https://stanford.edu/~mwaskom/software/seaborn/examples/heatmap_annotation.html), only the cell in the lower left corner has an annotation (in a wrong color):

![image](https://cloud.githubusercontent.com/assets/12086562/12489905/04f63956-c073-11e5-82e6-4cd10927a5c6.png)

The same issue also occurs with other data sets:
```python
sns.heatmap(
    pd.DataFrame({
        'wanna': {'ACAD': 53, 'MAG': 490, 'NEWS': 249}, 
        'gonna': {'ACAD': 202, 'MAG': 1974, 'NEWS': 1207}}), 
    annot=True)
```

![image](https://cloud.githubusercontent.com/assets/12086562/12489884/e62750fa-c072-11e5-9d8c-b6c6649427e2.png)

mwaskom:
Duplicate of #619

Issue #619, opened 2015-07-01
Annotation not showing on heatmap


legaultmarc:
In version 0.6.0, annotations do not show up on the heatmap (for me at least). Here is a minimal example to reproduce the problem:

```python
import numpy as np
import seaborn
import matplotlib.pyplot as plt

data = np.random.random((5, 5))
seaborn.heatmap(data, annot=True)
plt.savefig("test.png")
````
Resulting in the following image:

![test](https://cloud.githubusercontent.com/assets/864275/8461869/981a5db4-1ffd-11e5-98ac-41562eccc080.png)

From what I understand, `annot=True` should write the matrix entries on the heatmap, like symmatplot used to do (and as shown in the documentation). Interestingly, there seems to be only one value in the bottom left corner.

**Comment by mwaskom** [comment]
I can't reproduce locally:

![test](https://cloud.githubusercontent.com/assets/315810/8462338/faea392a-1fe7-11e5-8150-fd7a7163746f.png)

Can you provide some more information about your environment (matplotlib version, backend, maybe OS, etc).

**Comment by legaultmarc** [comment]
Mac OS: 10.10.4
Python: 2.7.8
Numpy: 1.9.2
Matplotlib: 1.4.3, backend: MacOSX 
Seaborn: 0.7.dev and 0.6.0

This is probably platform specific, so I will see if I can fix it on my setup.

**Comment by mwaskom** [comment]
First thing I would try is using a different matplotlib backend...seaborn seems to find a number of bugs in the MacOSX one.

**Comment by legaultmarc** [comment]
Oh you are right, using "Agg" fixed it...
Thank you for your help.

**Comment by mwaskom** [comment]
Not really sure what the actual problem is, though. Relevant code is here: https://github.com/mwaskom/seaborn/blob/master/seaborn/matrix.py#L195

Strangely, it looks like only one cell is getting annotated (so it's not a downstream drawing problem), as `len(ax.texts)` is `1`.

**Comment by mwaskom** [comment]
Ah,

```python
mesh = plt.pcolormesh(data)
mesh.get_facecolors()
```
returns 

```
array([[ 0.29803922,  0.44705882,  0.69019608,  1.        ]])
```

with the MacOSX backend, but

```
array([[ 0.21776241,  0.21776241,  0.21776241,  1.        ],
       [ 0.57747023,  0.57747023,  0.57747023,  1.        ],
       [ 0.56455212,  0.56455212,  0.56455212,  1.        ],
[...]

PR #831, opened 2016-01-21
Update mesh colors before drawing annotations

gkunter:
This commit fixes Issue #830. Currently, ```mesh.get_facecolors()``` returns only the color code for one cell in the grid, but calling ```mesh.update_scalarmappable()``` changes this so that the color information is shown for all cells.

### DIFF
--- a/seaborn/tests/test_matrix.py
+++ b/seaborn/tests/test_matrix.py
@@ -249,7 +249,7 @@ class TestHeatmap(PlotTestCase):
 
         ax = mat.heatmap(self.df_norm, annot=True)
         mesh = ax.collections[0]
-        nt.assert_equal(len(mesh.get_facecolors()), 4 * 8)
+        nt.assert_equal(len(mesh.get_facecolors()), self.df_norm.size)
 
         plt.close("all")
 
--- a/seaborn/tests/test_matrix.py
+++ b/seaborn/tests/test_matrix.py
@@ -245,6 +245,14 @@ class TestHeatmap(PlotTestCase):
         for val, text in zip(df_masked[::-1].compressed(), ax.texts):
             nt.assert_equal("{:.1f}".format(val), text.get_text())
 
+    def test_heatmap_annotation_mesh_colors(self):
+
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks — but please add a test.

mwaskom:
Also given that this behavior varies across matplotlib backends, I suspect that really there is a patch that should be submitted upstream.

gkunter:
I've tried to come up with a test -- please bear with me if I did that wrong; I don't have much experience with unit testing yet.

With the qt4agg backende, the fix does resolve the cause of the problem that you've identified in #619, i.e. mesh.get_facecolors() returning only data for a single row. With the fix, the expected number of rows is returned.

seaborn/tests/test_matrix.py
```diff
@@ -245,6 +245,14 @@ def test_heatmap_annotation_with_mask(self):
         for val, text in zip(df_masked[::-1].compressed(), ax.texts):
             nt.assert_equal("{:.1f}".format(val), text.get_text())
 
+    def test_heatmap_annotation_mesh_colors(self):
+
+        ax = mat.heatmap(self.df_norm, annot=True)
+        mesh = ax.collections[0]
+        nt.assert_equal(len(mesh.get_facecolors()), 4 * 8)
```
mwaskom: I would change this to infer the needed size from the dataframe itself (i.e. `self.df_norm.size`). As written, this test would break (possibly in a confusing way) if `df_norm` is changed. Otherwise looks great.

gkunter:
Thanks for the comment, I've changed the test in the way you suggested.

mwaskom:
Oops, I guess `DataFrame` doesn't have a `size` method on whatever the minimal pandas version is. `df.values.size` should work, I think.

gkunter:
Changed that. ```df_norm.values.size``` does work with my pandas version (0.16.2), but I can't test older versions. Sorry for that!

mwaskom:
No worries, that's what Travis is for.

mwaskom:
All clear — thanks again!

PR #838, opened 2016-01-26
Change luminance calculations for Annotation color choice

BrianLandry:
### Adjustment to luminance calculation when annotating heatmaps

Currently when annotating heatmaps the color of the text annotation is either white or gray based upon a contrast calculation. This pull request contains an improved contrast calculation to address an issue when using the viridis heatmap.

Currently the luminance is calculated by conversion to HSL and extraction of the luminance term, which is simply [the average of the highest and lowest RGB values](http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/). This is then compared to a 50% cutoff to decide when to use white or gray text.

I am proposing adherence to the WC3 standards for [luminance](https://www.w3.org/TR/WCAG20/#relativeluminancedef) and [optimal contrast](https://www.w3.org/TR/WCAG20/#relativeluminancedef). Importantly these take into account how people view luminance and contrast of different colors.

Excerpt describing relative luminance calculation:

> For the sRGB colorspace, the relative luminance of a color is defined as L = 0.2126 * R + 0.7152 * G + 0.0722 * B where R, G and B are defined as:

> if RsRGB <= 0.03928 then R = RsRGB/12.92 else R = ((RsRGB+0.055)/1.055) ^ 2.4
> if GsRGB <= 0.03928 then G = GsRGB/12.92 else G = ((GsRGB+0.055)/1.055) ^ 2.4
> if BsRGB <= 0.03928 then B = BsRGB/12.92 else B = ((BsRGB+0.055)/1.055) ^ 2.4

> and RsRGB, GsRGB, and BsRGB are defined as:

> RsRGB = R8bit/255
> GsRGB = G8bit/255
> BsRGB = B8bit/255

Excerpt for luminance calculation:

> (L1 + 0.05) / (L2 + 0.05), where
> L1 is the relative luminance of the lighter of the colors, and
> L2 is the relative luminance of the darker of the colors.

To calculate the text color, either grey (luminance = 0.15) or white (luminance = 1), I calculate which has the maximal contrast to the background luminance:

````
".15" if ((l+0.05)/(0.15+0.05))>((1+0.05)/(l+0.05)) else "w"
````

This simplifies to 

````
text_color = ".15" if l > .408 else "w"
````

This fix was motivated by the following issue with the viridis heatmap (matplotlib 2.0 default) where 19 is difficult to read. 

image

This is addressed in this pull request.

image

To get a better undestranding of the change I ploted luminance of both viridis and RdBu_r compared to the cutoff. This shows the weird dip in viridis that gave rise to the previously described issue/

### Viridis

image

### RdBu_r

image

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -210,8 +210,12 @@ class _HeatMapper(object):
         for x, y, val, color in zip(xpos.flat, ypos.flat,
                                     mesh.get_array(), mesh.get_facecolors()):
             if val is not np.ma.masked:
-                _, l, _ = colorsys.rgb_to_hls(*color[:3])
-                text_color = ".15" if l > .5 else "w"
+                r_s, g_s, b_s = color[:3]
+                r = r_s/12.92 if r_s <= 0.03928 else ((r_s+0.055)/1.055) ** 2.4
+                g = g_s/12.92 if g_s <= 0.03928 else ((g_s+0.055)/1.055) ** 2.4
+                b = b_s/12.92 if b_s <= 0.03928 else ((b_s+0.055)/1.055) ** 2.4
+                l = 0.2126 * r + 0.7152 * g + 0.0722 * b
+                text_color = ".15" if l > .408 else "w"
                 val = ("{:" + self.fmt + "}").format(val)
                 text_kwargs = dict(color=text_color, ha="center", va="center")
                 text_kwargs.update(self.annot_kws)
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -1,7 +1,6 @@
 """Functions to visualize matrices of data."""
 import itertools
 
-import colorsys
 import matplotlib as mpl
 from matplotlib.collections import LineCollection
 import matplotlib.pyplot as plt
@@ -13,7 +12,7 @@ from scipy.cluster import hierarchy
 
 from .axisgrid import Grid
 from .palettes import cubehelix_palette
-from .utils import despine, axis_ticklabels_overlap
+from .utils import despine, axis_ticklabels_overlap, relative_luminance
 from .external.six.moves import range
 
 
@@ -210,11 +209,7 @@ class _HeatMapper(object):
         for x, y, val, color in zip(xpos.flat, ypos.flat,
                                     mesh.get_array(), mesh.get_facecolors()):
             if val is not np.ma.masked:
-                r_s, g_s, b_s = color[:3]
-                r = r_s/12.92 if r_s <= 0.03928 else ((r_s+0.055)/1.055) ** 2.4
-                g = g_s/12.92 if g_s <= 0.03928 else ((g_s+0.055)/1.055) ** 2.4
-                b = b_s/12.92 if b_s <= 0.03928 else ((b_s+0.055)/1.055) ** 2.4
-                l = 0.2126 * r + 0.7152 * g + 0.0722 * b
+                l = relative_luminance(color)
                 text_color = ".15" if l > .408 else "w"
                 val = ("{:" + self.fmt + "}").format(val)
                 text_kwargs = dict(color=text_color, ha="center", va="center")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mwaskom** [comment]
Please refactor out into a separate function and add tests.

**Comment by mwaskom** [comment]
Otherwise I like the general idea — using the `l` channel was just a simple hack.

**Comment by BrianLandry** [comment]
Good to hear, I'll refactor and unit test tomorrow hopefully. Thanks for the great package! I'm converting my lab to it.

**Comment by mwaskom** [comment]
I'd put a little function in `seaborn.utils`. Maybe be useful elsewhere...

**Comment by mwaskom** [comment]
Also I should say that I am a little skeptical of the plots you've made, particularly the one for `RdBu`, which behaves very strangely towards the blue end. Using colorspacious to convert to CAM02-UCS suggests a different luminance profile that better matches my subjective perception:

rdbu

Similarly, `viridis` has a linear luminance ramp by design, but that isn't apparent in your plot.

I don't think the conversion for this sort of application needs to be perfect, but the plots seem weird enough to warrant double-checking (or using a different conversion method).

**Comment by BrianLandry** [comment]
Good catch. My code generating those plots had a bug. I fixed it and now the RdBu_r does not have the non-intuitive dip. Viridis is still not a linear ramp, but that's probably because it was designed with another luminance calculation (code below).

Given that relative luminance behaves as expected now I would like to go forward with the W3C standards.

I refactored the luminance calculation into a utils function and unit tested it, let me know if you need anything else done.

```python
colorMap = cm.RdBu_r

x = [i/256.0 for i in range(0,256,8)]
c = range(0,256,8)
hls = [colorsys.rgb_to_hls(colorMap(i)[0],colorMap(i)[1],colorMap(i)[2])[1] for i in range(0,256,8)]
relative_luminance = [sns.utils.relative_luminance(colorMap(i)) for i in range(0,256,8)]

plt.scatter(x = x, y = hls, c=c, cmap=colorMap, marker = "^", label = "hsl")
plt.scatter(x = x, y = relative_luminance, c=c, cmap=colorMap, label = "Relative Luminance")
plt.legend(loc = "upper left")
plt.xlim((0,1));
plt.ylim((0,1));
```
output_1_0

```python
colorMap = cm.viridis

x = [i/256.0 for i in range(0,256,8)]
c = range(0,256,8)
hls = [colorsys.rgb_to_hls(colorMap(i)[0],colorMap(i)[1],colorMap(i)[2])[1] for i in range(0,256,8)]
relative_luminance = [sns.utils.relative_luminance(colorMap(i)) for i in range(0,256,8)]

plt.scatter(x = x, y = hls, c=c, cmap=colorMap, marker = "^", label = "hsl")
plt.scatter(x = x, y = relative_luminance, c=c, cmap=colorMap, label = "Relative Luminance")
plt.legend(loc = "upper left")
plt.xlim((0,1));
plt.ylim((0,1));
```
output_2_0

mwaskom:
I'd suggest vectorizing the function, e.g.:

```python
def relative_luminance(color):

    rgb = mpl.colors.colorConverter.to_rgba_array(color)[:, :3]
    rgb = np.where(rgb <= .03928, rgb / 12.92, ((rgb + .055) / 1.055) ** 2.4)
    lum = rgb.dot([.2126, .7152, .0722])
    try:
        return lum.item()
    except ValueError:
        return lum
```

This behaves as before for single inputs 

```python
relative_luminance("red")
0.2126
```

But can also operate on a vector of points 

```python
cmap = mpl.cm.RdBu
x = np.linspace(0, 1, 100)
y = relative_luminance(cmap(x))
plt.scatter(x, y, c=x, cmap=cmap, s=50)
```

Plus, the fixed parameters don't have to be copy-pasted, avoiding errors.

mwaskom:
Thanks!

BrianLandry:
No problem, thanks for the great package!

PR #845, opened 2016-01-29
Remove color_list from dendrogram call in tests

### LINKED ISSUE #844: hierarchy.dendrogram getting color_list argument in test_matrix
Commit #758 removed the `color_list` argument, which was removed in scipy 0.17.0, from `seaborn/matrix.py`.  However, the `color_list` argument is still present in `seaborn/tests/test_matrix.py` even in master, causing unit tests to fail with scipy 0.17.0.

toddrme2178:
Fixes #844


mwaskom:
Thanks!

PR #868, opened 2016-02-17
Ability to override histtype when plotting PairGrid histograms

nickgravish:
This change gives the ability to pass in a specific histtype to map_diag to override the default behavior. The default barstacked behavior is retained if no histtype is passed into the function. This feature gives added functionality to map_diag without having to write a custom hist function just to use a different histtype. 

As a side note I wonder if barstacked is the best default behavior, barstacked has the annoying behavior that when passing in an array of samples the histograms are normalized to the total sample size instead of each histogram being independently normalized. Thus distributions with larger numbers of samples are larger in the resulting histogram. For my use cases this seems backwards, since a point of normalization is to compare distribution shape independent of sample size. 

A simple working example of this

```python
import numpy as np
import seaborn as sns
import pandas as pd

# make some test data with a boolean label of unequal sample size
df = pd.DataFrame({'data1':np.random.randn(2000), 
                   'data2':np.random.randn(2000), 
                   'label': np.random.randn(2000) > 0.2})

# Default seaborn/barstacked behavior
g = sns.PairGrid(df, x_vars= ['data1', 'data2'], 
                 y_vars= ['data1', 'data2'], 
                 hue = "label", diag_sharey = False, 
                 palette = "Set1")
                 
g.map_diag(plt.hist, normed = True, bins = np.arange(-10,10,.5))
                     

# Updated custom histtype, and also illustrate how normalization should appear
g = sns.PairGrid(df, x_vars= ['data1', 'data2'], 
                 y_vars= ['data1', 'data2'], 
                 hue = "label", diag_sharey = False, 
                 palette = "Set1")
                 
g.map_diag(plt.hist, normed = True, 
                     alpha = 0.8, 
                     histtype = 'step',
                     linewidth = 3, 
                     bins = np.arange(-10,10,.5))
                     
```

Default seaborn/barstacked normalization behavior
image

Updated function to illustrate stepped histogram, and also highlight correct normalization of distributions with different sample sizes.
image

### DIFF
--- a/examples/custom_map_diag_hist.py
+++ b/examples/custom_map_diag_hist.py
@@ -12,9 +12,9 @@ sns.set(style="white", palette="muted", color_codes=True)
 rs = np.random.RandomState(10)
 
 # make some test data with a boolean label of unequal sample size
-df = pd.DataFrame({'data1':np.random.randn(2000),
-                   'data2':np.random.randn(2000),
-                   'label': np.random.randn(2000) > 0})
+df = pd.DataFrame({'data1':np.random.randn(20000),
+                   'data2':np.random.randn(20000),
+                   'label': np.random.randn(20000) > 0})
 
 # Default barstacked histogram
 g = sns.PairGrid(df, x_vars= ['data1', 'data2'],
--- a/examples/custom_map_diag_hist.py
+++ b/examples/custom_map_diag_hist.py
@@ -12,25 +12,27 @@ sns.set(style="white", palette="muted", color_codes=True)
 rs = np.random.RandomState(10)
 
 # make some test data with a boolean label of unequal sample size
-df = pd.DataFrame({'data1':np.random.randn(20000),
-                   'data2':np.random.randn(20000),
-                   'label': np.random.randn(20000) > 0})
+df = pd.DataFrame({'data1':np.random.randn(2000),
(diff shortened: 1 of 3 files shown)


mwaskom:
I don't feel strongly about stacked histograms. I don't fully understand the point about normalization, but that's just default matplotlib behavior so it doesn't matter.

Please fix the failing PEP8 check, and it would be best to add a test to cover the new behavior.

Additionally, if you could add or update an example to document this option in the docstring, it would be great.

nickgravish:
Thanks. I have fixed the error and added an example as well as a test. 

My point about the barstacked behavior is that it makes visual comparisons between distributions challenging, and even worse possibly misleading if you don't recognize the behavior (which was the problem I initially had). 

In the example I give above the distributions between the hue datasets are the same, but the barstacked histograms don't give a good visual representation of this because one distribution is stacked atop the other. To identify that they are the same I need to visually notice that one distribution is exactly half the height of the other. In the non-stacked cases (like the stepped example) the identical distributions lie atop each other and give a good visual indication that they are the same. I think "step" or "stepfilled" would be a better default behavior when there is labeled data. But I am just as happy to have the option to choose the histtype myself withi this PR. Thanks!

mwaskom:
Thanks!

PR #877, opened 2016-03-01
Support dataframes as col/row_colors in clustermap and use columns as annotation

jrderuiter:
Includes the proposed changes from #441 for supporting dataframe row_colors and col_colors in cluster map and using the column names as labels in the resulting plot. The current implementation allows supplying colors as dataframes, in which case columns as used as labels, or as a (named) series, in which case the name of the series is used as label (if the name is not None). 

The current implementation also matches the index of row_colors and col_colors to the data index/columns, to ensure that the order of the colors is coherent with the data. This may break current behaviour, but I personally feel that it is a welcome addition, as it ensures that your colors correspond properly with your data (this has bitten me before) and makes good use of the index/column annotation to do so. I do not currently check for missing entries, and the torgb code breaks on NaNs at the moment. My initial solution to NaNs with be to replace them with the background color (white?) effectively omitting them from the annotation.

I have not written any extra unit tests yet, but the existing tests all pass. If there is legitimate interest in the feature, I am happy to add the extra unit tests.

A basic example:

```{python}
import pandas as pd
import numpy as np
import seaborn as sis

np.random.seed(0)

# Generate data.
data = pd.DataFrame(np.random.randn(10, 10),
                    index=['F{}'.format(i) for i in range(10)],
                    columns=['S{}'.format(i) for i in range(10)])

# Generate colors.
col_colors = pd.DataFrame({'Annotation 1': (['red'] * 5) + ['blue'] * 5},
                          index=data.columns)

row_colors = pd.DataFrame({'Annotation 2': (['green'] * 5) + (['black'] * 5)},
                          index=data.index)

# Draw example.
sns.clustermap(data, col_colors=col_colors, row_colors=row_colors)
```

unknown

A shuffled example (where colors are shuffled):

```{python}
np.random.seed(0)

# Generate colors
col_colors = pd.DataFrame({'Annotation 1': (['red'] * 5) + ['blue'] * 5},
                          index=data.columns)

row_colors = pd.DataFrame({'Annotation 2': (['green'] * 5) + (['black'] * 5)},
                          index=data.index)

# Shuffle col_colors.
samples = list(data.columns)
np.random.shuffle(samples)
col_colors = col_colors.ix[samples]

# Shuffle row_colors
features = list(data.index)
np.random.shuffle(features)
row_colors = row_colors.ix[features]

# Shuffled example, without proper matching.
sns.clustermap(data, col_colors=col_colors.T.values, row_colors=row_colors.T.values)

# Shuffled example, now matched by index/column.
sns.clustermap(data, col_colors=col_colors, row_colors=row_colors)
```

unknown-1
unknown-2

And using the network example:

```{python}
import pandas as pd
import seaborn as sns
sns.set(font="monospace")

# Load the brain networks example dataset
df = sns.load_dataset("brain_networks", header=[0, 1, 2], index_col=0)

# Select a subset of the networks
used_networks = [1, 5, 6, 7, 8, 11, 12, 13, 16, 17]
used_columns = (df.columns.get_level_values("network")
                          .astype(int)
                          .isin(used_networks))
df = df.loc[:, used_columns]

# Create a custom palette to identify the networks
network_pal = sns.cubehelix_palette(len(used_networks),
                                    light=.9, dark=.1, reverse=True,
                                    start=1, rot=-2)
network_lut = dict(zip(map(str, used_networks), network_pal))

# Convert the palette to vectors that will be drawn on the side of the matrix
networks = df.columns.get_level_values("network")
network_colors = pd.Series(networks).map(network_lut)

# Create a custom colormap for the heatmap values
cmap = sns.diverging_palette(h_neg=210, h_pos=350, s=90, l=30, as_cmap=True)

# Convert colors to dataframe.
colors_df = pd.DataFrame({'colors': network_colors})
colors_df.index = df.columns

# Draw the full plot
sns.clustermap(df.corr(), row_colors= colors_df, linewidths=.5,
               col_colors= colors_df, figsize=(13, 13), cmap=cmap)
```
unknown-4

Supplying as (named) Series:

```{python}
# Draw the full plot
sns.clustermap(df.corr(), row_colors=colors_df['colors'], linewidths=.5,
               col_colors=colors_df['colors'], figsize=(13, 13), cmap=cmap)
```
unknown-5

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -715,19 +715,8 @@ class ClusterGrid(Grid):
             figsize = (width, height)
         self.fig = plt.figure(figsize=figsize)
 
-        # if row_colors is not None:
-        #     if isinstance(row_colors, (pd.DataFrame, pd.Series)):
-        #         # Ensure colors match data indices
-        #         row_colors = row_colors.ix[data.index]
-        #     row_colors = _convert_colors(row_colors)
         self.row_colors, self.row_color_labels = \
             self._preprocess_colors(data, row_colors, axis=0)
-
-        # if col_colors is not None:
-        #     if isinstance(col_colors, (pd.DataFrame, pd.Series)):
-        #         # Ensure colors match data indices
-        #         col_colors = col_colors.ix[data.columns]
-        #     col_colors = _convert_colors(col_colors)
         self.col_colors, self.col_color_labels = \
             self._preprocess_colors(data, col_colors, axis=1)
 
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -52,6 +52,7 @@ def _convert_colors(colors):
             # If we get here, we have nested lists
             return [list(map(to_rgb, l)) for l in colors]
 
+
 def _matrix_mask(data, mask):
     """Ensure that data and mask are compatabile and add missing values.
 
@@ -714,19 +715,21 @@ class ClusterGrid(Grid):
             figsize = (width, height)
         self.fig = plt.figure(figsize=figsize)
 
-        if row_colors is not None:
-            if isinstance(row_colors, (pd.DataFrame, pd.Series)):
-                # Ensure colors match data indices
-                row_colors = row_colors.ix[data.index]
-            row_colors = _convert_colors(row_colors)
-        self.row_colors = row_colors
+        # if row_colors is not None:
+        #     if isinstance(row_colors, (pd.DataFrame, pd.Series)):
+        #         # Ensure colors match data indices
+        #         row_colors = row_colors.ix[data.index]
+        #     row_colors = _convert_colors(row_colors)
+        self.row_colors, self.row_color_labels = \
+            self._preprocess_colors(data, row_colors, axis=0)
 
-        if col_colors is not None:
-            if isinstance(col_colors, (pd.DataFrame, pd.Series)):
-                # Ensure colors match data indices
-                col_colors = col_colors.ix[data.columns]
-            col_colors = _convert_colors(col_colors)
-        self.col_colors = col_colors
+        # if col_colors is not None:
+        #     if isinstance(col_colors, (pd.DataFrame, pd.Series)):
+        #         # Ensure colors match data indices
+        #         col_colors = col_colors.ix[data.columns]
+        #     col_colors = _convert_colors(col_colors)
+        self.col_colors, self.col_color_labels = \
+            self._preprocess_colors(data, col_colors, axis=1)
 
         width_ratios = self.dim_ratios(self.row_colors,
                                        figsize=figsize,
@@ -764,6 +767,32 @@ class ClusterGrid(Grid):
 
         self.dendrogram_row = None
         self.dendrogram_col = None
+
+    def _preprocess_colors(self, data, colors, axis):
+        labels = None
+
+        if colors is not None:
+            if isinstance(colors, (pd.DataFrame, pd.Series)):
+                # Ensure colors match data indices
+                if axis == 0:
+                    colors = colors.ix[data.index]
+                else:
+                    colors = colors.ix[data.columns]
+
+                # Replace na's with background color
+                colors = colors.fillna('white')
+
+                # Extract color values and labels from frame/series
+                if isinstance(colors, pd.DataFrame):
+                    labels = colors.columns
+                    colors = colors.T.values
+                else:
+                    labels = [colors.name]
+                    colors = colors.values
+
+            colors = _convert_colors(colors)
+
+        return colors, labels
 
     def format_data(self, data, pivot_kws, z_score=None,
                     standard_scale=None):
@@ -913,10 +942,6 @@ class ClusterGrid(Grid):
             all_colors = set(itertools.chain(*colors))
             n = len(colors)
             m = len(colors[0])
-        elif isinstance(colors, pd.DataFrame):
-            all_colors = set(itertools.chain(*colors.values))
-            m, n = colors.shape
-            colors = colors.T.values
         else:
             all_colors = set(colors)
             n = 1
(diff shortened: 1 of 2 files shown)


seaborn/matrix.py
```diff
@@ -36,14 +36,21 @@ def _index_to_ticklabels(index):
...
-        # If this works, there is only one level of colors
-        return list(map(to_rgb, colors))
-    except ValueError:
-        # If we get here, we have nested lists
-        return [list(map(to_rgb, l)) for l in colors]
 
+    if isinstance(colors, pd.DataFrame):
+        # Convert dataframe
+        return pd.DataFrame({col: colors[col].map(to_rgb)
+                            for col in colors})
```
mwaskom: I think `colors.applymap(to_rgb)` should work? (Only lightly tested).

seaborn/matrix.py
```diff
@@ -708,9 +715,16 @@ def __init__(self, data, pivot_kws=None, z_score=None, standard_scale=None,
         self.fig = plt.figure(figsize=figsize)
 
         if row_colors is not None:
+            if isinstance(row_colors, (pd.DataFrame, pd.Series)):
+                # Ensure colors match data indices
+                row_colors = row_colors.ix[data.index]
```
mwaskom: I think this is reasonable, but you'll want to be very careful to explicitly test the expected failure modes.

seaborn/matrix.py
```diff
@@ -899,6 +913,10 @@ def color_list_to_matrix_and_cmap(colors, ind, axis=0):
             all_colors = set(itertools.chain(*colors))
             n = len(colors)
             m = len(colors[0])
+        elif isinstance(colors, pd.DataFrame):
```
mwaskom: I'm not sure that it matters, but the order of priority in terms of checking for the different input types seems to vary across methods here. It would be good to try to be consistent.

seaborn/matrix.py
```diff
@@ -964,18 +982,49 @@ def plot_colors(self, xind, yind, **kws):
         if self.row_colors is not None:
             matrix, cmap = self.color_list_to_matrix_and_cmap(
                 self.row_colors, yind, axis=0)
+
+            # Get labels from colors if given as dataframe
+            if isinstance(self.row_colors, pd.DataFrame):
+                xticklabels = self.row_colors.columns
+            elif isinstance(self.row_colors, pd.Series) and \
+                    self.row_colors.name:
```
mwaskom: If you're checking for `None`, do so explicitly.

seaborn/matrix.py
```diff
@@ -964,18 +982,49 @@ def plot_colors(self, xind, yind, **kws):
...
+
             heatmap(matrix, cmap=cmap, cbar=False, ax=self.ax_row_colors,
-                    xticklabels=False, yticklabels=False,
+                    xticklabels=xticklabels, yticklabels=False,
                     **kws)
+
+            # Adjust rotation of labels
+            if xticklabels is not False:
+                self.ax_row_colors.set_xticklabels(
+                    self.ax_row_colors.get_xticklabels(), rotation=90)
```
mwaskom: I'm not sure I like the color labels being always rotated orthogonal to the axis. Also, you can just do `plt.setp(self.ax.row_colors.get_xticklabels(), rotation=90)`.

seaborn/matrix.py
```diff
@@ -964,18 +982,49 @@ def plot_colors(self, xind, yind, **kws):
...
+
             heatmap(matrix, cmap=cmap, cbar=False, ax=self.ax_col_colors,
-                    xticklabels=False, yticklabels=False,
+                    xticklabels=False, yticklabels=yticklabels,
                     **kws)
+
+            # Adjust rotation of labels, place on right side
+            if yticklabels is not False:
+                self.ax_col_colors.set_yticklabels(
+                    self.ax_col_colors.get_yticklabels(), rotation=0)
```
mwaskom: Same point as above.

mwaskom:
I think this is a good addition. It will be very important to add tests. (Ideally, the tests would have been written first, but it's ok, that is a hard standard to adhere to).

One comment is that the logic of handling the different types is spread out throughout the class and is repeated multiple times. I think it would be much cleaner to convert the color arrays to a common representation once, early. I don't have to time right now to fully get my head back around how this code is structured, so I am not certain whether that is possible or exactly what/where the best way/place to do it would be.

If you do want to explore this idea, I would recommend first writing tests against this implementation and then refactoring.

jrderuiter:
I see your point. One solution would be to extract labels early from the dataframe/series and then drop the colors down to the previous array/list format.

seaborn/matrix.py
jrderuiter: For long labels (and even for labels that are not even that long), they quickly overlap if they are not drawn orthogonally. I think this would be a good default, labels can always be rotated back if this is preferred. Making it configurable might be too much of a hassle.

mwaskom: I think that's reasonable...I don't think it should be configurable, but it would be nice if it were easier to predict the relative size of matplotlib text and other elements to automatically choose a good orientation. But I don't think it needs to be handled here.

jrderuiter:
These additions refactor most code into a separate method _preprocess_colors (for lack of a better name), which could be made static, and adds two properties to clustermap: row_color_labels and col_color_labels. The method _preprocess_colors aims to convert colors to a 'standard format' before assigning them to the clustermap, thereby avoiding having to test for types downstream. Labels are extracted as part of this process.

I have also addressed the missing entries issue, which now works by replacing NaNs in the Series/DataFrame by a background color (here hardcoded as white). Is this appropriate? Should I obtain the default background color from somewhere?

mwaskom:
Nice, much cleaner! I think white to represent nans is a good call; the dendrogram axes is always set to white, so those cells will be continuous with the surround in a way that should effectively connote "missing".

hejing3283:
Thanks for the solution. One more questions, How to add category annotation legends?

PR #883, opened 2016-03-06
Fix missing labels in clustermap col_colors when row_colors not given

jrderuiter:
Small adjustments to #877. I discovered a small typo in checking for col_label_colors, which resulted in column color labels not being drawn if no row_label_colors were given. I also changed preprocess_colors to ensure that labels are now a simple list instead of a pandas Index object. Finally, I added some documentation of clustermap to reflect the addition of the Series/Dataframe functionality.

Sorry for the late changes, I didn't expect the pull request to be accepted that quickly after the first revision!

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -758,6 +758,7 @@ class ClusterGrid(Grid):
         self.dendrogram_col = None
 
     def _preprocess_colors(self, data, colors, axis):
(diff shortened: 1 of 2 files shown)


mwaskom:
Nice catch. Would you mind opening up a new PR against the v0.7 branch?

PR #885, opened 2016-03-13
Fix #719: UnicodeEncodingError in FacetGrid

### LINKED ISSUE #719: FacetGrid gives UnicodeEncodeError when data contains utf-8 characters

gkunter:
These commits fix the Unicode-related issue #719 (see the commit messages for details). It may be possible that there are still more places in the code in which a conversion to Unicode is necessary.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1840,4 +1840,3 @@ class JointGrid(object):
         """Wrap figure.savefig defaulting to tight bounding box."""
         kwargs.setdefault("bbox_inches", "tight")
         self.fig.savefig(*args, **kwargs)
-
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -922,6 +922,10 @@ class FacetGrid(Grid):
                 template = row_template
             else:
                 template = " | ".join([row_template, col_template])
+                
+        row_template = utils.to_utf8(row_template)
+        col_template = utils.to_utf8(col_template)
+        template = utils.to_utf8(template)
 
         if self._margin_titles:
             if self.row_names is not None:
@@ -1836,3 +1840,4 @@ class JointGrid(object):
         """Wrap figure.savefig defaulting to tight bounding box."""
         kwargs.setdefault("bbox_inches", "tight")
         self.fig.savefig(*args, **kwargs)
+
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -922,7 +922,7 @@ class FacetGrid(Grid):
                 template = row_template
             else:
                 template = " | ".join([row_template, col_template])
-                
+
         row_template = utils.to_utf8(row_template)
         col_template = utils.to_utf8(col_template)
         template = utils.to_utf8(template)
--- a/seaborn/tests/test_axisgrid.py
+++ b/seaborn/tests/test_axisgrid.py
@@ -581,7 +581,7 @@ class TestFacetGrid(PlotTestCase):
         nt.assert_equal(g._not_na.sum(), 50)
 
     def test_unicode_column_label_with_rows(self):
-        
+
         # use a smaller copy of the default testing data frame:
         df = self.df.copy()
         df = df[["a", "b", "x"]]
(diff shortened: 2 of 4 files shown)


gkunter:
There is a typo in utils.py.

mwaskom:
To be clear, in this comment:

> Under Python 2.7, all tests fail in FacetGrid.set_titles() in one of the lines containing the instruction ```title = template.format(**args)```.

do you mean the tests would fail without the addition of the `utils.to_utf8` conversion?

gkunter:
Yes, exactly that: without this addition, a UnicodeEncodingError will be raised by the tests in Python 2.7 because ``FacetGrid.set_titles()`` attempts to format the Unicode string arguments with the format string ``template``, which is a byte string.

With this addition, no such exception is raised in Python 2.7. In Python 3.x, the tests will always pass, with or without the addition.

mwaskom:
Great. Thanks, this is an excellent PR. Other unicode issues may pop up, but this will make it easier to handle them.

gkunter:
Glad to help. I've had myself so many Unicode problems with a code base that is supposed to run both in Python 2.7 and Python 3.x that I've kind of learned how to deal with them. And I've learned to loath Python 2.7 for its horrible Unicode support in the process...

PR #886, opened 2016-03-15
Heatmap annotate values with other dataframe

olgabot:
The `annot=True` flag in `sns.heatmap()` is very helpful:

```python
import seaborn as sns
import numpy as np
np.random.seed(314)

values = np.random.randint(low=0, high=1e6, size=20).reshape(4, 5)
values[2, 3] += 1000000
sns.heatmap(values, annot=True)
```

image

But sometimes I want to plot the percentages as the heatmap, but annotate with the original values. This is especially when some of the percentages are very low but I'd still like the raw value to tell them apart.

```python
percentages = 100*values/values.sum()
sns.heatmap(percentages, annot=True, annot_data=values, vmin=0, vmax=100)
```
image

This PR adds the keyword argument `annot_data=None` and allows the user to pass a dataframe/numpy array to use as the annotation. It doesn't check for whether the `annot_data` and `data` are the same size but it probably should.

[Here](http://nbviewer.jupyter.org/gist/olgabot/c9d87392ac4f1b44564b) is an nbviewer notebook with the code above.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -172,7 +172,10 @@ class _HeatMapper(object):
         if annot_data is None:
             self.annot_data = plot_data
         else:
-            self.annot_data = annot_data.ix[::-1].values
+            if isinstance(annot_data, pd.DataFrame):
+                self.annot_data = annot_data.ix[::-1].values
+            else:
+                self.annot_data = annot_data[::-1]
 
         self.fmt = fmt
         self.annot_kws = {} if annot_kws is None else annot_kws
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -93,7 +93,7 @@ class _HeatMapper(object):
     """Draw a heatmap plot of a matrix with nice labels and colormaps."""
 
     def __init__(self, data, vmin, vmax, cmap, center, robust, annot, fmt,
-                 annot_kws, annot_data, cbar, cbar_kws,
+                 annot_kws, cbar, cbar_kws,
                  xticklabels=True, yticklabels=True, mask=None):
         """Initialize the plotting object."""
         # We always want to have a DataFrame with semantic information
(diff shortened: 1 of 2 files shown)


mwaskom:
I think I would prefer allowing the user to pass a DataFrame to `annot` rather than adding another kwarg. There's no reason to have both, I think, because I don't imagine a situation where you'd want to pass `annot_data` and *not* have `annot=True` right? It makes it a little bit less obvious that it is an option, but you'd have to be reading the documentation pretty closely to even see that it is possible anyway :)

seaborn/matrix.py
```diff
@@ -304,8 +318,10 @@ def heatmap(data, vmin=None, vmax=None, cmap=None, center=None, robust=False,
     robust : bool, optional
         If True and ``vmin`` or ``vmax`` are absent, the colormap range is
         computed with robust quantiles instead of the extreme values.
-    annot : bool, optional
-        If True, write the data value in each cell.
+    annot : bool, or array, optional
```
mwaskom: I'd change this to "or rectangular dataset" (for consistency with `data` above because it can be a DataFrame right?)

seaborn/matrix.py
```diff
@@ -168,6 +168,19 @@ def __init__(self, data, vmin, vmax, cmap, center, robust, annot, fmt,
         self.data = data
         self.plot_data = plot_data
         self.annot = annot
+
+        if isinstance(self.annot, bool) and self.annot:
+            self.annot_data = plot_data
+        elif not isinstance(self.annot, bool):
```
mwaskom: I think this check will catch `None`? I may have lost track of what `self.annot` is by default at this point, though.

P.S. if you're going to touch this code again it might be cleaner to remove the `self.` since `annot` should still be a local variable at this point I think.

mwaskom: Oh I see, the default is `False` not `None`. What do you think about changing to `None`? It will make this logic a little messier, but it might be a stronger signal that values other than `True` or `False ` are valid for this parameter.

mwaskom:
Thanks!

PR #887, opened 2016-03-15
apionly no longer calls reset_orig() - fixes #695

### LINKED ISSUE #695: import seaborn.apionly breaks %matplotlib inline
See this gist for an example:
https://gist.github.com/shoyer/31e308a90c9882324f31

In contrast, if the `import seaborn.apionly` line is commented out, a plot is produced. If it matters, I'm using the conda matplotlib (1.4.3) on OS X.

I believe this is the same issue @dubourg described in this comment:
https://github.com/mwaskom/seaborn/pull/126#issuecomment-134592134

`import seaborn.apionly` also appears to be somehow responsible for some strange styling issues we're having with xray plotting in the notebook (I'm not entirely sure because I haven't succeeded yet in making a minimal reproducible example):
https://gist.github.com/shoyer/0233061ba31e3ec71009

mdengler:


mwaskom:
I don't understand this PR. It defeats the entire purpose of the `apionly` "module".

mdengler:
I understood that "apionly" was there to only export the seaborn's names and not change the [current style parameters / "aesthetics"](https://stanford.edu/~mwaskom/software/seaborn/installing.html):

> Seaborn will apply its default style parameters to the global matplotlib style dictionary when you import it. [..] To avoid this behavior **and use the default matplotlib aesthetics** [...] you can import the seaborn.apionly namespace.

I see how the "and use the default matplotlib aesthetics" belies my understanding.
 
I think the goal of having a seaborn namespace available while avoiding seaborn applying its default style parameters to the global matplotlib style dictionary is quite an important use case, to allow people to write code that doesn't have to place "import seaborn [for names only]" carefully relative to matplotlib imports, ipython notebook magics, etc:

```
%matplotlib inline
plt.rcParams[...] = [...]
import seaborn.namesonly as sns

sns.heatmap(...)
```

I had read "import sns.apionly" as "import the names only without any side-effects", but I see that was not the intent.  Is there a pull request you might consider that would cater for my use case?

PR #891, opened 2016-03-19
Clustermap: collapse dendrograms, side_colors and colorbar if not given, make ratios configurable.

jrderuiter:
This pull request addresses some issues that were originally discussed in #437. More specifically, I have introduced the following changes:
- Added col_color_ratio and row_color_ratio arguments that allow the user to specific a ratio for the side_colors when building the axes for clustermap. This allows users to adjust the size of the colors if needed (to prevent overlapping labels, or colors that are too compressed to view). I think it is best that this can be overridden by the user, as I don't see a sensible default that will always work (for example, choosing the same height as the rows will not work in heatmaps with many rows/columns, as is generally the case for gene expression matrices etc.)
- Refactored the setting up of axes into a single function _setup_axes. The function accounts for the presence of row/column clustering and row/column colors when setting up the axes, avoiding adding axes of no clustering is performed or no colors are provided.  
- Made the colorbar optional, which as a bonus allows for dropping whitespace if dendrograms are omitted. 

I am not quite happy with the colorbar, as it results in some considerable whitespace when dendrograms are omitted. We could choose to drop the colorbar down to where the row dendrogram would be if no dendrograms are needed, which would avoid at least the vertical whitespace. I don't however see a cleaner option without moving the colorbar to the right, which then runs into issues with the labels of the heatmap.

Some examples are shown below.

Full plot:

unknown

No col_cluster:

unknown-1

No col/row_cluster:

unknown-2

No clustering + no colorbar:

unknown-3

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -20,12 +20,13 @@ from .external.six import string_types
 
 
 __all__ = ["heatmap", "clustermap"]
-    precedence goes to key value pairs in latter dicts.
-    """
-    result = {}
-    for dictionary in dict_args:
-        result.update(dictionary)
-    return result
+
+
+def _merge_dicts(a, b):
+    """Merge two dicts."""
+    merged = dict(a)
+    merged.update(b)
+    return merged
 
 
 def _index_to_label(index):
@@ -65,7 +66,7 @@ def _convert_colors(colors):
 
 
 def _matrix_mask(data, mask):
-    """Ensure that data and mask are compatabile and add missing values.
+    """Ensure that data and mask are compatible and add missing values.
 
     Values will be plotted for cells where ``mask`` is ``False``.
 
@@ -641,6 +642,7 @@ class _DendrogramPlotter(object):
 
         Returns
         -------
+
         dendrogram : dict
             Dendrogram dictionary as returned by scipy.cluster.hierarchy
             .dendrogram. The important key-value pairing is
@@ -756,6 +758,7 @@ def dendrogram(data, linkage=None, axis=1, label=True, metric='euclidean',
 
 
 class ClusterGrid(Grid):
+
     def __init__(self, data, pivot_kws=None, z_score=None, standard_scale=None,
                  figsize=None, row_colors=None, col_colors=None,
                  row_ratios=None, col_ratios=None, row_cluster=True,
@@ -802,17 +805,19 @@ class ClusterGrid(Grid):
         self.fig = plt.figure(figsize=figsize)
 
         # Setup axes grid.
-        height_ratios = self._dim_ratios(figsize=figsize, axis=0,
-                                         ratios=col_ratios)
+        height_ratios = self._dim_ratios(
+            figsize=figsize, axis=0, ratios=col_ratios)
 
-        width_ratios = self._dim_ratios(figsize=figsize,  axis=1,
-                                        ratios=row_ratios)
+        width_ratios = self._dim_ratios(
+            figsize=figsize, axis=1, ratios=row_ratios)
 
-        self.gs = gridspec.GridSpec(len(height_ratios),
-                                    len(width_ratios),
-                                    wspace=0.01, hspace=0.01,
-                                    width_ratios=width_ratios,
-                                    height_ratios=height_ratios)
+        self.gs = gridspec.GridSpec(
+            len(height_ratios),
+            len(width_ratios),
+            wspace=0.01,
+            hspace=0.01,
+            width_ratios=width_ratios,
+            height_ratios=height_ratios)
 
         # Set dendrogram axes.
         ax_index = slice(0, 2) if self.colorbar else 0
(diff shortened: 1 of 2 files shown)


jrderuiter:
And of course an example with col_color_ratio = 0.1:

unknown-6

As a side note, I just realized that I haven't added any tests for col/row_color_ratios. Will do so later.

jrderuiter:
Any ideas/remarks on this?

jrderuiter:
Added a few changes that also allow dendrogram/heatmap ratios to be configurable. This is especially useful for clustermaps with strange aspect ratios, such as this:

unknown

Which with these extra arguments can be made more readable:

```{python}
g = clustermap(data, ..., figsize=(18, 3.5),
                         col_ratios={'side_colors': 0.2, 'dendrogram': 0.5})
```
unknown-1

I think default behaviour might be improved for most users by making smarter defaults. I do however think that exactly specifying ratios is very useful for edge cases (such as these broad/long clustermaps) and for tuning your plots.

jrderuiter:
@mwaskom Does this mean there's no interest in this?

mwaskom:
Lack of time, not interest.

mwaskom:
Here are a few thoughts:

- I agree that having some flexibility in the plot layout would be very useful. This generally seems like a fine approach. The keys to the dictionaries that the user passes are not very discoverable (and at the moment this feature is undocumented), so maybe it would be good to but some effort into an informative error message when keys are wrong. I'm also not sure the key name needs to be `side_colors`, though in retrospect a better name for the "colors" would have been helpful.
- Refactoring the setup code is great (though I haven't read it through closely).
- I also agree that collapsing unused space is good, and that having the colorbar in the upper left hand corner is not ideal, but that there isn't a great alternative. Making it optional, at least, is a good start.

Just from looking at the example plots, I'm a little confused though. In the last plot, you have `{'side_colors': 0.2, 'dendrogram': 0.5}`. I would have thought that this would produce a plot where the color matrix takes up 20% of the vertical space and the dendrogram takes up 50% of the vertical space. That doesn't seem to be what's happening in the image, though. Is this behavior correct?

Just a heads up that I am moving across the country over the course of the next two weeks, so I might be slow to respond, but this work is good and I think has some features that should get merged.

jrderuiter:
Great, thanks for the feedback! Will try to follow up on this in the coming
weeks.

mvacher:
This looks great. 
Will this code be integrated in the next release ?
Thanks.

liyun831229:
I would like to know if I can add color legend to annotate col_colors. This is extremely useful since the color along can not tell anything. I am using pheatmap from R now. I really hope that clustermap can do the same thing.

g = clustermap(data, ..., figsize=(18, 3.5), col_ratios={'side_colors': 0.2, 'dendrogram': 0.5})
I can not set col_ratios in clutermap

iandriver:
I have just been doing this recently. Since placement is an issue you can add a legend to ax_heatmap manually through matplotlib and position it properly:

for each color and the label for that color in col_colors create a handle:
	leg_handles_gene =[]
	leg_handles_gene.append(patches.Patch(color=ycolor, label=ygroup_name))

Then add as legend to ax_heatmap:
	gene_legend = g.ax_heatmap.legend(handles=leg_handles_gene, loc=2, bbox_to_anchor=(1.04, 1), title='Gene groups', prop={'size':fontsize}) 
	plt.setp(gene_legend.get_title(),fontsize=fontsize)

And if you need to add more than one legend (otherwise the next one will overwrite the first):
	cg.ax_heatmap.add_artist(gene_legend)

Hope that helps. Seems like what you were asking.

liyun831229:
Thanks, @iandriver . I am able to add multiple legends to the heatmap now. However, is there any way to add a legend title for each of the legend. In my case add legend title "Time", and 'Group'. 
Here is the code I am using to generate the follow plot:
**[heatmap.txt](https://github.com/mwaskom/seaborn/files/590346/heatmap.txt)**
heatmap

jrderuiter:
@mwaskom I rebased this branch to be up-to-date with the current master. I still feel the work should be integrated, as the configurable ratios is something I use frequently (which is the reason that I have been using my own branch of seaborn for a while).

This also adds a check for invalid keys in the {col/row}_ratios, to improve discoverability of the proper keys (which are now also in the documentation). I need to add a few unit tests, but otherwise I think this is ready for merging.

mwaskom:
I agree it's a nice feature. It's hard to find the time to review a large set of changes thoroughly. Can you put together a notebook file I could download that has examples of the functionality that is added by this PR? (More or less what you've demonstrated in various comments in the thread, just in a format I could download and run). That would help me understand it better I think.

yueqiw:
Hi, I'm looking for a way to increase the height of color bars when having multiple color bar annotations. Adding the option to set `{col/row}_ratios` would be really helpful. 

I can see that this is an old PR with many changes. If `{col/row}_ratios` is the most important feature in this PR, maybe minimizing the code change to only reflect this feature could help integrating it into the next release? 

A simplified implementation might be:  
```
# customize ratios
clustermap(... , width_ratios=[0.1, 0.1, 0.8], height_ratios=[0.1, 0.1, 0.8], ...)
# default
clustermap(... , width_ratios=None, height_ratios=None, ...)
```
If `width_ratios` or `height_ratios` matches the number of subplots, they can then be passed to GridSpec() directly. This avoids complicated changes and invalid key checks.

PR #893, opened 2016-03-24
doc: Add URL links to 2.7 and 3.3 version.

root-nishi:
### Single change to README.md
- Made python version numbers 2.7 and 3.3 as hyperlinks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Seaborn can be cited using a DOI provided through Zenodo: [![DOI](https://zenodo
 Dependencies
 ------------
 
-- Python 2.7 or 3.3+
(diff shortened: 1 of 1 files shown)


mwaskom:
This isn't really helpful because the semantics of that line in the docs is Python 3.3 **and later**, so linking to 3.3 (which really, nobody does or should use) is counterproductive.

Also given that I recommend people use Anaconda or a similar Python distribution, linking to the python.org download page is counterproductive.

root-nishi:
Thanks for helping me understand!

So in my python repos, would it make sense for me to link to Anacondas instead of Python X.Y?

PR #914, opened 2016-05-01
Optimize pruning of candidates in SwarmPlotter

codiply:
Exit the for-loop when the first overlap with a neighbor is found.


mwaskom:
I don't see what effect this would have?

codiply:
I fixed it to actually do what I intended. 
Once `good_candidate` is set to `False` it cannot change back to `True`. Therefore, we can skip checking the rest of the neighbours for overlaps.

mwaskom:
I think it's fine as is.

PR #918, opened 2016-05-08
Fix scale bug in _scipy_univariate_kde

farr:
There is a missing scale factor in `bw` in `_scipy_univariate_kde`.  This patch fixes it.  The bug becomes evident when the scale of the data to which the KDE is applied is much smaller than 1.  I have attached two `kdeplot`s from before and after the fix.  The data were created as `data = 0.01*randn(1000)`; you can see that the scale is properly tracked post-fix.
[pre-fix.pdf](https://github.com/mwaskom/seaborn/files/254088/pre-fix.pdf)
[post-fix.pdf](https://github.com/mwaskom/seaborn/files/254089/post-fix.pdf)

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -345,7 +345,7 @@ def _scipy_univariate_kde(data, bw, gridsize, cut, clip):
             warnings.warn(msg, UserWarning)
     if isinstance(bw, string_types):
         bw = "scotts" if bw == "scott" else bw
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks for catching this bug. I fixed the scale issue in https://github.com/mwaskom/seaborn/commit/758fad534832ba9bfcc28913f219ed45423e75d1 because you had pushes some additional commits to your master branch. Please send PRs from specific feature branches in the future.

PR #922, opened 2016-05-14
Fix header interference in doc linking, fixes #610

### LINKED ISSUE #610: The header interferes with linking to doc sections
Linking to sections in the docs is annoying because the anchor goes to the top of the window and the header gets in the way.

I'm not exactly sure how to fix this in the context of Sphinx; I tried [here](https://github.com/mwaskom/seaborn/commit/c9f8e12af26817b91fcb3145db334cc100ed4abf#diff-1a451631fe18f4ca156584c3c67928ffR21) but this doesn't seem to quite work.

If someone with more css talent wants to fix this, that would be awesome.

adelq:
Title says all, simply some CSS that fixes the anchor link to doc pages

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -18,12 +18,9 @@ code {
     color: #2c3e50 !important;
 }
 
-.headerlink:before {
-
-  display: block;
-  context: " ";
-  height: 200px; /*same height as header*/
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks, this is super helpful!

PR #928, opened 2016-05-18
Add hist2d to allowed joint plots

toobaz:
Notice that the grid is hidden because otherwise it appears over the squares, [as reported here](https://github.com/mwaskom/seaborn/issues/321#issue-44908636). So a better fix could be introduced eventually.


mwaskom:
Closing because as you note there are problems with this implementation, and it is trivial to accomplish using JointGrid directly.

toobaz:
> it is trivial to accomplish using JointGrid directly.

hint?

Issue #909, opened 2016-04-23
Got ValueError from clustermap, says "Valid methods when the raw observations are omitted are 'single', 'complete', 'weighted', and 'average'"


gepcel:
When I run this:

``` python
import seaborn as sns
sns.clustermap(spe_norm, method='ward')
```

`spe_norm` is from [this file](https://gist.githubusercontent.com/gepcel/4c8c657796a481a78a89754298c408ac/raw/e48d3dcdfaa19f65212106dd64f7e3ecd271f640/spe-norm.csv). I got an error:

``` python
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
C:\anaconda\lib\site-packages\seaborn\matrix.py in calculated_linkage(self)
    563         try:
--> 564             return self._calculate_linkage_fastcluster()
    565         except ImportError:

C:\anaconda\lib\site-packages\seaborn\matrix.py in _calculate_linkage_fastcluster(self)
    543     def _calculate_linkage_fastcluster(self):
--> 544         import fastcluster
    545         # Fastcluster has a memory-saving vectorized version, but only

ImportError: No module named 'fastcluster'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-8-224973069d28> in <module>()
      1 import seaborn as sns
      2 rcParams['font.family'] = 'Noto Sans S Chinese'
----> 3 sns.clustermap(spe_norm, method='ward')

C:\anaconda\lib\site-packages\seaborn\matrix.py in clustermap(data, pivot_kws, method, metric, z_score, standard_scale, figsize, cbar_kws, row_cluster, col_cluster, row_linkage, col_linkage, row_colors, col_colors, mask, **kwargs)
   1167                         row_cluster=row_cluster, col_cluster=col_cluster,
   1168                         row_linkage=row_linkage, col_linkage=col_linkage,
-> 1169                         **kwargs)

C:\anaconda\lib\site-packages\seaborn\matrix.py in plot(self, metric, method, colorbar_kws, row_cluster, col_cluster, row_linkage, col_linkage, **kws)
   1006         colorbar_kws = {} if colorbar_kws is None else colorbar_kws
   1007         self.plot_dendrograms(row_cluster, col_cluster, metric, method,
-> 1008                               row_linkage=row_linkage, col_linkage=col_linkage)
   1009         try:
   1010             xind = self.dendrogram_col.reordered_ind

[...]

PR #930, opened 2016-05-20
Use one-step linkage calculation with both scipy and fastcluster

olgabot:
Trying to solve https://github.com/mwaskom/seaborn/issues/909, do linkage calculation in one step rather than two because of a weird compatibility issue with `scipy` 0.17 (https://github.com/scipy/scipy/issues/6180)


mwaskom:
Thanks, can you think of any problems this might raise?

mwaskom:
Also any idea why the tests aren't running?

mwaskom:
Ah, can we do it both ways, i.e. try this way and fall back if that's not possible? (Assuming it raises a predictable error). Or do a version check (though the former way is better). Current minimal scipy version is 0.11 so that would be a big jump.

PR #931, opened 2016-05-21
Fix #929: FacetGrid with unused categories

### LINKED ISSUE #929: FacetGrid fails if dataframe contains unused categories
FacetGrid fails with an unhelpful matplotlib exception if the column being conditioned on is a pandas categorical, and not all categoricals are used.

Here's an example

```python
def neighborhood(borough):
    neighborhoods = {'A': lambda: np.random.choice(list('abcd')),
                     'B': lambda: np.random.choice(list('efgh')),
                     'C': lambda: np.random.choice(list('ijkl')),
                     'D': lambda: np.random.choice(list('mnop'))}
    return neighborhoods[borough]()

n = 10000
df = pd.DataFrame({'val': np.random.randn(n),
                   'borough': np.random.choice(list("ABCD"), size=n)})
df['neighborhood'] = df['borough'].apply(neighborhood)
```

This sets up a dataframe with random float values and, borough and neighborhood columns (strings).

```python
g = sns.FacetGrid(df.query("borough == 'A'"), col='neighborhood', col_wrap=4)
g.map(plt.hist, 'val')
```

works fine, which is to say 4 plots are generated, one for each of the neighborhoods in borough "A".

But if we convert neighborhood to a categorical it fails, presumably because the DataFrame passed to FacetGrid has a mismatch between the categories that actually appear in the DataFrame and the contents of df['neighborhood'].cat.categories :

```python
df['neighborhood'] = df['neighborhood'].astype('category')
g = sns.FacetGrid(df.query("borough == 'A'"), col='neighborhood', col_wrap=4)
g.map(plt.hist, 'val')
```

with the exception:

```
g = sns.FacetGrid(df.query("borough == 'A'"), col='neighborhood', col_wrap=4)
g.map(plt.hist, 'val')
g = sns.FacetGrid(df.query("borough == 'A'"), col='neighborhood', col_wrap=4)
g.map(plt.hist, 'val')
<ipython-input-5-1275d2e2ef07> in <module>()
----> 1 g = sns.FacetGrid(df.query("borough == 'A'"), col='neighborhood', col_wrap=4)
      2 g.map(plt.hist, 'val')

/Users/mike/.virtualenvs/ds3/lib/python3.5/site-packages/seaborn/axisgrid.py in __init__(self, data, row, col, hue, col_wrap, sharex, sharey, size, aspect, palette, row_order, col_order, hue_order, hue_kws, dropna, legend_out, despine, margin_titles, xlim, ylim, subplot_kws, gridspec_kws)
    321                 subplot_kws["sharey"] = axes[0]
    322             for i in range(1, n_axes):
--> 323                 axes[i] = fig.add_subplot(nrow, ncol, i + 1, **subplot_kws)
    324             self.axes = axes
    325 

/Users/mike/.virtualenvs/ds3/lib/python3.5/site-packages/matplotlib/figure.py in add_subplot(self, *args, **kwargs)
   1003                     self._axstack.remove(ax)
   1004 
-> 1005             a = subplot_class_factory(projection_class)(self, *args, **kwargs)
   1006 
   1007         self._axstack.add(key, a)

/Users/mike/.virtualenvs/ds3/lib/python3.5/site-packages/matplotlib/axes/_subplots.py in __init__(self, fig, *args, **kwargs)
     62                     raise ValueError(
     63                         "num must be 1 <= num <= {maxn}, not {num}".format(
---> 64                             maxn=rows*cols, num=num))
     65                 self._subplotspec = GridSpec(rows, cols)[int(num) - 1]
     66                 # num - 1 for converting from MATLAB to python indexing

ValueError: num must be 1 <= num <= 4, not 5
```

It looks to me like the problem is that `categorical_order()` is returning all the categories, including the unused ones. 

Changing `categorical_order()` to return only used categories fixes the problem, in the sense that I get the same result whether or not the column is categorical.

Given the semantics of pandas categories, you could make the case that seaborn should build an axes for every category, including the unused ones.

Should I submit a PR that plots only used categories? Or does someone smarter than me want to figure out how to makes plots for the unused categories?

williamsmj:
Calculates correct number of subplots to draw when unused categories are present

The test confirms the fix solves the ValueError.

It also asserts that the number of axes drawn by FacetGrid should be equal to the number of categories in `.cat.categories`, not the number of unique values that are present in the series.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -271,7 +271,7 @@ class FacetGrid(Grid):
                 err = "Cannot use `row` and `col_wrap` together."
                 raise ValueError(err)
             ncol = col_wrap
-            nrow = int(np.ceil(len(data[col].unique()) / col_wrap))
+            nrow = int(np.ceil(len(col_names) / col_wrap))
         self._ncol = ncol
         self._nrow = nrow
 
--- a/seaborn/tests/test_axisgrid.py
+++ b/seaborn/tests/test_axisgrid.py
@@ -718,7 +718,7 @@ class TestFacetGrid(PlotTestCase):
         df = self.df.copy()
         df['a'] = df['a'].astype('category')
 
-        g = ag.FacetGrid(df.query("a != 'a'"), col="a", col_wrap=1)
+        g = ag.FacetGrid(df[df['a'] == 'a'], col="a", col_wrap=1)
 
         nt.assert_equal(g.axes.shape, (len(df['a'].cat.categories),))
 
--- a/seaborn/tests/test_axisgrid.py
+++ b/seaborn/tests/test_axisgrid.py
@@ -24,6 +24,7 @@ from ..utils import categorical_order
 rs = np.random.RandomState(0)
 
 old_matplotlib = LooseVersion(mpl.__version__) < "1.4"
+pandas_has_categoricals = LooseVersion(pd.__version__) >= "0.15"
 
 
 class TestFacetGrid(PlotTestCase):
@@ -713,6 +714,7 @@ class TestFacetGrid(PlotTestCase):
         g = ag.FacetGrid(df, row="b")
         g = g.map(plt.plot, "x")
 
+    @skipif(not pandas_has_categoricals)
     def test_categorical_column_missing_categories(self):
 
         df = self.df.copy()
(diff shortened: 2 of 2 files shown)


mwaskom:
Ah, I think that test is probably going to fail on older versions of pandas that don't have category datatypes. Can you look at the other tests that use categories to see how that's handled?

williamsmj:
Yes, the minimal 2.7 Travis build failed. I'll take a look.

williamsmj:
Do you want me to squash these down into one commit? Otherwise looks good to me now.

mwaskom:
I'm not particularly disciplined about squashing so don't worry about the three commits. I'll have to do the merge by hand so that it can get into the v0.7 branch, and will try to get to that later today. Thanks!

mwaskom:
Thanks again!

PR #932, opened 2016-05-21
Stylish class, equivalent to mpl.rc_context

toobaz:
See [this question](http://stackoverflow.com/questions/31191063/using-with-sns-set-in-seaborn-plots/37365095) on SO.

(probably worth merging with a more serious name - or maybe just transforming ``set`` itself in a context manager?)


mwaskom:
I think the answer given in that SO post is sufficient.

toobaz:
I agree, that's precisely why I would merge it :-)

PR #939, opened 2016-06-05
made usage of string module in one example compatible to Python3

krother:
In Py3k, `string` does not contain `letters` any more, only `ascii_letters`. It is also available in Python 2.7 with the same content as `letters`.


mwaskom:
Thanks... I implemented your fix in https://github.com/mwaskom/seaborn/commit/f95dc5560d0445bde437d250418f2c215e9ef959 because annoyingly the order of lower and uppercase is swapped in `letters` and `ascii_letters`.

Issue #432, opened 2015-01-23
"transform" kwargs for box/violin plots


phobson:
Curious if you have any interest in transformation parameters for box and violin plots. 

For example, if you had, oh I don't know, lognormally distributed pollutant concentrations in water, you might want to compute the KDEs or boxplot stats in log space (i.e., `transformin = lambda x: np.log(x)`, but still show the plots in native units with or without a log-axis (i.e., `transformout = lambda x: np.exp(x)`.

It's easy enough to simply log-transform your data and label the axes as such. You could even use a fancy matplotlib `FuncFormatter` to fix all the major tick labels. But the transformations aren't always `np.log`/`np.exp`. For example, the discussion [here](http://stats.stackexchange.com/questions/112705/boxplots-with-lognormally-distributed-data) indicates that some folks even draw boxplot on the square-roots of their data. So perhaps a generalized transformation system could be useful.

I think I see in both `_BoxPlotter` and `_ViolinPlotter` where these transformation would need to take place. But, if there's no interest in this, I'll shelve it and get back to the documentation I've been meaning to write.

mwaskom:
I played around with this a bit and then ended up punting because it seemed very complicated. But yeah I agree in general plotting these distributions in some transformed space is useful and it would be nice to make that as easy as possible in seaborn.

For boxplots specifically, I think we'd need to compute the boxplot statistics ourselves in the intermediate space, and then transform those back out? (I'm actually a little confused about why going into and then back out of log space would influence how the boxplot gets drawn, since they're just order statistics, but maybe it influences the whisker range and what counts as an outlier?) IIRC you added the ability to draw a boxplot from the statistics in 1.4, but that might be a bit of an issue since we aim to support 1.1+ (I think; the real rule is whatever keeps the debian stable builds happy). Of course it wouldn't be that hard to draw boxes ourselves, but we'd lose (or have to reimplement) stuff like notches, which we currently get for free.

Let's revisit this after I finish unifying box/violin/stripplots with bar/pointplots, at which point the architecture will be a lot more stable.

phobson:
> but maybe it influences the whisker range and what counts as an outlier

That's exactly what it does, and the difference can be pretty dramatic.
![bxp](https://cloud.githubusercontent.com/assets/1163939/5904560/4d522cd8-a53e-11e4-8f0e-920f1d09a7a3.png)

[...]

PR #944, opened 2016-06-14
Logscale kde

farr:
I have found it nice to sometimes plot p(log(x)) from samples of x, automatically log-ing the x-axis.  This branch adds a `logspace=` optional argument to the KDE density plots which does exactly this.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -255,12 +255,17 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
 
 
 def _univariate_kdeplot(data, shade, vertical, kernel, bw, gridsize, cut,
-                        clip, legend, ax, cumulative=False, **kwargs):
+                        clip, legend, ax, logspace=False, cumulative=False,
+                        **kwargs):
     """Plot a univariate kernel density estimate on one of the axes."""
 
     # Sort out the clipping
     if clip is None:
         clip = (-np.inf, np.inf)
+
+    if logspace:
+        # Do the KDE-ing and plotting on the log of the data, then transform back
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks for sending this. I'm going to close if for now because I wouldn't want to add this to `kdeplot` without also adding it to `violinplot`, which would be trickier and I wouldn't have the bandwidth to review in the near future. Sorry, annoying I know, but better than letting the PR fester.

farr:
No problem.  I will make a note that this should be added to the violin
plots, and maybe I'll get around to it in the near future.

mwaskom:
Adding it to violinplot adds some additional complexity because it entails that other categorical plots (at least boxplots) should also compute their summary stats in logspace. Always a tradeoff between having useful options and a consistent API.

phobson:
FWIW I've thought about #432 a lot and I see a path forward for transformed boxplots. Namely, that would be computing the boxplots stats ourselves and then passing the the stats to `Axes.bxp` rather than the data to `Axes.boxplot`. However, this would required matplotlib >= 1.4 (when the `bxp` method was introduced). I've thought about this enough to know that while the above sounds simple enough, we would be introducing non-trivial complexity since we'd have to handle kwargs meant for `boxplot` that aren't intended for `bxp`.

@farr rather than a `logspace` option, I would prefer using something like `transform_in` and `transform_out` to be more general. But that's just my 2c. Admittedly, I'd only use `np.log` and `np.exp`, so I'd be happy either way.

farr:
To be honest, the only transform I ever use is log/exp, but I would be happy with the option to be more general as well.

PR #1001, opened 2016-08-31
ENH for heatmap: adds invert_yaxis to flip the y axis

stsievert:
Often times in my statistical visualizations with `sns.heatmap`, it'd be nice to have invert the y-axis. This is possible through `ax.invert_yaxis` as found in [this StackOverflow answer] but it'd be nice to have a new default option -- I'd rather not have to search for this.

A basic test (really a use case):

screen shot 2016-08-31 at 9 51 03 am

[this StackOverflow answer]:http://stackoverflow.com/a/34444939

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -296,7 +296,7 @@ def heatmap(data, vmin=None, vmax=None, cmap=None, center=None, robust=False,
             linewidths=0, linecolor="white",
             cbar=True, cbar_kws=None, cbar_ax=None,
             square=False, ax=None, xticklabels=True, yticklabels=True,
-            mask=None,
+            mask=None, invert_y=False,
             **kwargs):
     """Plot rectangular data as a color-encoded matrix.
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks for sending. I'm going to reject because my policy is not to hide every detail of the matplotlib implementation for customizing plots like this. There may be cases in which there are arguments to seaborn functions that simple call a simple matplotlib method, but that's unsustainable and so I think it's best to avoid them where they are newly introduced.

stsievert:
> I'm going to reject because my policy is not to hide every detail of the matplotlib implementation for customizing plots like this

Yes, seaborn is a thin wrapper around a matplotlib frontend. It should remain so and details for customizing plots should rely on mpl.

The reason I sent this pull request is that I want to do this in *every* heatmap I generate.  I would use this much more than the `square` keyword argument, which calls a single simple matplotlib function (at [matrix.py#L494]).

`ax.invert_yaxis` is not a new feature and has existed for at least 3 years ([source]). Worst case, we can always use `ax.set_ylim` (which the implementation of `invert_yaxis` relies on) which is far from new.

[matrix.py#L494]:https://github.com/mwaskom/seaborn/blob/master/seaborn/matrix.py#L494
[source]:https://github.com/matplotlib/matplotlib/blame/d308eec9fa2fe07be7a3c96d0f6765698615510d/lib/matplotlib/axes/_base.py#L2986

stsievert:
The use case I have in mind: I referred to this pull request so I didn't have to google again.

Issue #958, opened 2016-06-26
Find new home for docs


mwaskom:
My Stanford website will be deactivated sometime in the fall. Need to find a new home and set up appropriate redirects.

yarmenti:
Right, the site is now out.

anoroozian:
Could you temporarily make the documentation available as a compressed file on Github for offline use?

TomAugspurger:
@mwaskom want me to grab the `seaborn` Github name and throw the HTML docs up at `seaborn.github.io`? Should only take a few minutes.

mwaskom:
Irritatingly, the `seaborn` user/org name is already taken on github, even though https://github.com/seaborn doesn't exist.

TomAugspurger:
Oh... that is irritating.

mwaskom:
Maybe I'll ask Stanford to set up redirect to https://altair-viz.github.io/ :)

japborst:
I use seaborn for plots in my thesis, which is due rather soon. I don't have a Linux machine available, so cannot run the makefile myself unfortunately and don't have much experience with Sphinx, I'm afraid. Anyone can help me get the docs in some form?

Ps. why not host the the docs in this repo through `gh-pages` ?

TomAugspurger:
https://www.dropbox.com/s/2xbf7ou8ccsyplr/sb.tar.gz?dl=0

japborst:
@TomAugspurger awesome! Thanks a ton

mwaskom:
Notwithstanding my trolling about this on Twitter, the serious answer here is that I actually think Stanford deactivating my website was a mistake (I have an approved request for account extension, but the original expiry was last night) and am trying to sort that out with Stanford ID. But in the medium term the plan is to move to pydata.org.

nils-werner:
Why not readthedocs?

pmadhikar:
As a short term work around, I found a snapshot on archive.org [here](http://web.archive.org/web/20160804163120/https://stanford.edu/~mwaskom/software/seaborn/)
It was made on Aug. 8 though so it might be slightly out of date.

TomAugspurger:
> Why not readthedocs? It will take 3 minutes to set up.
 
The build process for seaborn's docs is a bit more exotic than RTD can handle.

rasbt:
> Irritatingly, the seaborn user/org name is already taken on github, even though https://github.com/seaborn doesn't exist.

[...]

PR #1052, opened 2016-10-13
DOCS: update docs links and remove dev docs link

nateGeorge:
Couldn't find the dev docs on the github pages, so I deleted those few lines.

#958


mwaskom:
Thanks. I'm not sure that github.io is going to be the permanent home but I guess this is currently accurate.

PR #1054, opened 2016-10-14
Add option ``fitted_kws`` in ``distplot``

gsagnol:
This option alo=lows to pass some kw_args arguments to the 
`fit( )` function, such as `floc=0`.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -36,7 +36,7 @@ def _freedman_diaconis_bins(a):
 
 
 def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
-             hist_kws=None, kde_kws=None, rug_kws=None, fit_kws=None,
+             hist_kws=None, kde_kws=None, rug_kws=None, fit_kws=None, fitter_kws=None,
              color=None, vertical=False, norm_hist=False, axlabel=None,
              label=None, ax=None):
     """Flexibly plot a univariate distribution of observations.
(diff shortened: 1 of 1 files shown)


seaborn/distributions.py
```diff
@@ -236,7 +240,7 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
         clip = fit_kws.pop("clip", (-np.inf, np.inf))
         bw = stats.gaussian_kde(a).scotts_factor() * a.std(ddof=1)
         x = _kde_support(a, bw, gridsize, cut, clip)
-        params = fit.fit(a)
+        params = fit.fit(a,**fitter_kws)
```
mmagnuski: space after comma

mmagnuski:
I've had a very similar situation when I needed to use `floc=0` because otherwise gamma distribution would not fit well. I would like such option (passing args to `fit`) but having `fit_kws` and `fitter_kws` would be confusing...

gsagnol:
Yes. The names might not be optimal. If you have a better idea, feel free to modify ! I think it might be a useful option, though.

mmagnuski:
Actually, I think `fit_kws` should be the ones passed to `fit()` and something like `plotfit_kws` should be those passed to `plot()` the fitted distribution. But that would require a deprecation etc. - not sure it is worth it. I normally fit the distribution myself and then plot it to the distplot axis.

mwaskom:
Another option is to make your own object and pass that in, e.g..

```python
class normfit(stats.norm):
    def fit(*args, **kwargs)
         return super(normfit, self).fit(loc=0, *args, **kwargs)
sns.distplot(x, fit=normfit)
```

Not super convenient for one-off use, but completely flexible.

mwaskom:
>  I normally fit the distribution myself and then plot it to the distplot axis.

I think this is the right approach. `distplot` is meant to be a function for quickly checking aspects of the distribution, not a general-purpose frontend for fitting parametric distributions.

PR #1069, opened 2016-11-18
Better error message when using 'split' with less than 2 hues.

stepthom:
My dataset had less than two hue levels, and the resultant error message was confusing. Hopefully, this helps.


phobson:
Hey at @stepthom.

I think this is a good change, but the linter doesn't like that the line you modified has more than 79 characters in it. Plenty of ways to skin that cat, but here's how I'd do it:

```python
        if split and self.hue_names is not None and len(self.hue_names) != 2:
            msg = "There must be exactly two hue levels to use `split`."
            raise ValueError(msg)
```

PR #1075, opened 2016-12-10
Fixed documentation issue on load_dataset

margaretpearce:
I edited a comment on load_dataset() in seaborn/utils.py, correcting the default path. A very small pull request, but the incorrect default data path caused me some confusion while I was trying to debug an issue with retrieving the iris data set behind a firewall.


mwaskom:
Thanks!

PR #1085, opened 2016-12-26
fixing husl int bug

choldgraf:
this fixes a bug that popped up when I was trying to use the `husl` module. Basically it's converting numbers to a scale of 255 and then calling `np.round`, but they're left as floats. This means that when they are converted to something like hex with `%02x`, it was returning an error because it required an int, not a float. This PR just explicitly turns RGB values into `int` after calling `np.round`.

### DIFF
--- a/seaborn/external/husl.py
+++ b/seaborn/external/husl.py
@@ -170,7 +170,7 @@ def rgb_prepare(triple):
         # instead of Python 2 which is rounded to 5.0 which caused
         # a couple off by one errors in the tests. Tests now all pass
         # in Python 2 and Python 3
-        ret.append(round(ch * 255 + 0.001, 0))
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks — you might also want to commit upstream to the pyhusl project, if they haven't fixed this yet. PS congrats on your Nature Comms paper — I liked the plots.

choldgraf:
The main project is [this one](https://github.com/hsluv/hsluv-python), yeah? It looks like they've already changed their API for converting to hex anyway, so I think this is fixed in there.

Thx for the congrats on the paper! Man it's nice to have that finally out. Been a 1.5 year review process :P

mwaskom:
Hm, yeah I think so. Looks like that package is an evolution of the old pyhusl that I copied into the seaborn source to minimize dependencies.

PR #1088, opened 2017-01-07
Replace str with to_utf8 methods

moosekaka:
Replaced str function calls with utils.to_utf8 to avoid unicode errors when labels had unicode in them. Tested with python 2.7.


mwaskom:
Looks like Travis is failing due to pep8 checks.

mwaskom:
Thanks!

PR #1101, opened 2017-01-24
Fixed broken link in doc/installing.rst

wonderb0lt:
I've also replaced the sourceforge link (that only redirects to matplotlib.org) with it's redirect target. It's a small but annoying issue :)


mwaskom:
Thanks

PR #1104, opened 2017-01-26
offset option for individual spines

rasi:
I wanted the offset to be set for each spine individually. So allowed offset to be a dict of ints.

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -173,9 +173,11 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
         Specific axes object to despine.
     top, right, left, bottom : boolean, optional
         If True, remove that spine.
-    offset : int or dict of ints or None  (default), optional
+    offset : int or dict or None  (default), optional
         Absolute distance, in points, spines should be moved away
-        from the axes (negative values move spines inward). Dict for each side can also be given.
+        from the axes (negative values move spines inward).
+        Can also be given as a dict with offset values  
+        for each spine, for example: {"left": 5, "bottom": 0}.
     trim : bool, optional
         If true, limit spines to the smallest and largest major tick
         on each non-despined axis.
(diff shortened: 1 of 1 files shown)


seaborn/utils.py
```diff
@@ -173,9 +173,9 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
         Specific axes object to despine.
     top, right, left, bottom : boolean, optional
         If True, remove that spine.
-    offset : int or None  (default), optional
+    offset : int or dict of ints or None  (default), optional
         Absolute distance, in points, spines should be moved away
-        from the axes (negative values move spines inward).
+        from the axes (negative values move spines inward). Dict for each side can also be given.
```
mwaskom: This line is failing the PEP8 check because it is too long. Also please be more clear about what the dict should look like, i.e. keys should be a spine name and value should be the size of the offset.

seaborn/utils.py
```diff
@@ -173,9 +173,9 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
         Specific axes object to despine.
     top, right, left, bottom : boolean, optional
         If True, remove that spine.
-    offset : int or None  (default), optional
+    offset : int or dict of ints or None  (default), optional
```
mwaskom: "dict of ints" does not make sense because the keys of the dict are not ints. Just say `dict` and provide more information below.

seaborn/utils.py
```diff
@@ -199,7 +199,10 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
             is_visible = not locals()[side]
             ax_i.spines[side].set_visible(is_visible)
             if offset is not None and is_visible:
-                _set_spine_position(ax_i.spines[side], ('outward', offset))
+                if type(offset) == int:
```
mwaskom: Use `isinstance()` not `type()`.

rasi:
I have made the changes you suggested.

mwaskom:
I added this feature in #1112 because your tests were still failing with simple errors. Please check that your code works before submitting a pull request in the future.

rasi:
Sure, will do. Thanks

Issue #1108, opened 2017-01-30
Matplotlib deprecation warning triggered in ClusterGrid class.


JWarmenhoven:
`MatplotlibDeprecationWarning: The axisbg attribute was deprecated in version 2.0. Use facecolor instead.`

https://github.com/mwaskom/seaborn/blob/master/seaborn/matrix.py:
``` python
 self.ax_row_dendrogram = self.fig.add_subplot(self.gs[nrows - 1, 0:2],
                                                axisbg="white")
 self.ax_col_dendrogram = self.fig.add_subplot(self.gs[0:2, ncols - 1],
                                              axisbg="white")
```

mwaskom:
Closed with #1111

PR #1109, opened 2017-01-30
Change MPL keyword 'axisbg' to 'facecolor'

JWarmenhoven:
#1108

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -760,9 +760,9 @@ class ClusterGrid(Grid):
                                     height_ratios=height_ratios)
 
         self.ax_row_dendrogram = self.fig.add_subplot(self.gs[nrows - 1, 0:2],
-                                                      axisbg="white")
+                                                      facecolor="white")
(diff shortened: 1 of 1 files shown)


mwaskom:
This change isn't sufficient because it will break on older versions of matplotlib. It's necessary to maintain backwards compatibility.

JWarmenhoven:
Hmm...yeah I realized that after testing some more.

mwaskom:
I don't actually know why the dendrogram axes are being set to white background + fully despined + no ticks; it would be easier just to do `set_axis_off()`.

JWarmenhoven:
Yes, a quick test shows that `set_axis_off()` on the col/row dendrogram axes appears to do the trick. But as you mentioned, there may be another use case where the background of the axes need to be set to white?

mwaskom:
No obvious reasons come to mind, and `set_axis_off` will work better in other circumstances, e.g. when you want a transparent figure background.

PR #1113, opened 2017-02-04
Update internally packaged six version

### LINKED ISSUE #966: Problem with pickle and old version of library six in the repository.
If you import seaborn and try to serialize, as example, `Ellipsis` object, then sometimes you will see an error `ImportError: No module named 'winreg'`:

``` python
import seaborn
import pickle

pickle.dumps(Ellipsis)
```

It reproduces not all the time, but usually for second or third try.

OS X El Capitan 10.11.5
Python 3.5.1
seaborn 0.7.0

As I understand, the reason is that seaborn uses outdated version of six library. When I tried to change six library from repository to an actual version the problem was solved.

mwaskom:
Closes #966

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ coverage:
 
 lint:
 
-	pyflakes -x W seaborn
+	pyflakes -x W -X seaborn/external/six.py seaborn
 	pep8 --exclude external seaborn
 
 hexstrip:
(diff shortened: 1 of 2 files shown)


Chilipp:
I confirm, that this solved the bugs :) Thanks a lot!

PR #1129, opened 2017-03-02
`sns.swarmplot` that scales better

lukauskas:
This PR changes the way swarmplot checks for point overlaps.

The way it currently works:
1. Swarmplot tries placing each of the swarm points one-by one.
2. Each time finding all the points the newly-placed point could be overlapping (i.e. points that are within one diameter from the point [NB: I suspect code uses "diameter" when it actually means "radius" but it doesn't matter here]).
3. Then it proposes new locations for points to be to the left/right of these points.
4. The algorithm then loops through all of these proposed points and check if they overlap with any of the previously found neighbours. If they do, these points are removed (leaving only non-overlapping points).
5. The algorithm then selects the middle-most point of the remaining points and places the new point there. Repeating everything.

Since overlap checking was the bottleneck, 
I changed the algorithm parts [4-5]  to pre-sort the candidate points by how close to the mid-point they are before checking for overlaps. In this scenario we need to only find one point that is good and thus can terminate early (a6ffdee)
I also moved some of the computations to numpy layer (4c4a29d). 
Tried out additional optimisations (other commits) but they did more bad than good so they were reverted.

To test the performance, I used the following script:

```python
import pandas as pd
import numpy as np
import click
import seaborn as sns
import timeit

print(sns.__version__)

N = 1000
KS = ['a', 'b', 'c']


def generate_plot(n, close=True):
    values = np.random.randn(n)
    facets = np.random.choice(KS, size=n)

    data = pd.DataFrame({'x': facets, 'y': values})

    sns.swarmplot(x='x', y='y', data=data)
    if close:
        plt.close()


@click.command()
@click.option('-n', type=int, default=N)
@click.option('--show-plot', is_flag=True)
def main(n, show_plot):
    np.random.seed(2317)
    print('N = {}'.format(n))
    if show_plot:
        generate_plot(n, close=False)
        plt.show()
    else:
        n_loop = 10
        times = timeit.repeat('generate_plot(n)',
                               setup='from __main__ import generate_plot; n = {!r}'.format(n),
                               number=n_loop,
                               repeat=3)

        import pint
        ureg = pint.UnitRegistry()

        for time in sorted(times):
            time /= n_loop
            print('{:.2f~} per iter ({!r} s)'.format((time * ureg.seconds).to_compact(), time))


if __name__ == '__main__':
    main()
```

Before (993c6a2) and after images of the generated swarm with N=1000 (`python swarm_example.py -n 1000 --show-plot`).

Before:

swarm-before

After:

swarm-after

(there should be no difference in how they look)

On the other hand, here are the stats for runtimes.
It is clear that the performance scales better now (yellow line vs red line):

image

| Commit	| N |	Best	| Best (s) |
| --------- | --- | -----------| -------- |
| 993c6a2 | 10 | 77.05 ms | 0.077048394 |
| 993c6a2 | 100 | 103.49 ms | 0.103489597 |
| 993c6a2 | 1000 | 1.17 s | 1.173759292 |
| 993c6a2 | 2000 | 7.89 s | 7.894506943 |
| a6ffdeef7 | 10 | 77.98 ms | 0.077982094 |
| a6ffdeef7 | 100 | 108.26 ms | 0.108259719 |
| a6ffdeef7 | 1000 | 713.11 ms | 0.713108945 |
| a6ffdeef7 | 2000 | 3.35 s | 3.349439375 |
| 4c4a29da | 10 | 77.75 ms | 0.077749728 |
| 4c4a29da | 100 | 110.44 ms | 0.110443773 |
| 4c4a29da | 1000 | 641.97 ms | 0.641972809 |
| 4c4a29da | 2000 | 1.84 s | 1.835392294 |
| ed06906 | 10 | 81.05 ms | 0.08104773 |
| ed06906 | 100 | 109.21 ms | 0.109209078 |
| ed06906 | 1000 | 770.27 ms | 0.770273886 |
| ed06906 | 2000 | 2.27 s | 2.27069 |
| 684725e43186 | 10 | 80.99 ms | 0.080990762 |
| 684725e43186 | 100 | 113.14 ms | 0.113138432 |
| 684725e43186 | 1000 | 785.46 ms | 0.785456822 |
| 684725e43186 | 2000 | 2.23 s | 2.228930535 |
| 41afc3fa952e2 | 10 | 79.99 ms | 0.079991932 |
| 41afc3fa952e2 | 100 | 110.46 ms  | 0.110459668 |
| 41afc3fa952e2 | 1000 | 650.15 ms | 0.650152328 |
| 41afc3fa952e2 | 2000 | 1.80 s | 1.795588099

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1202,8 +1202,6 @@ class _SwarmPlotter(_CategoricalScatterPlotter):
         """Return a list of all swarm points that could overlap with target.
 
         Assumes that swarm is a sorted list of all points below xy_i.
-
-        Returns a sorted list of neigbors
         """
         _, y_i = xy_i
         neighbors = []
@@ -1213,13 +1211,7 @@ class _SwarmPlotter(_CategoricalScatterPlotter):
                 neighbors.append(xy_j)
             else:
                 break
-
-        neighbors = np.array(neighbors)
-        if len(neighbors) == 0:
-            return neighbors
-
-        neighbors = neighbors[np.argsort(neighbors[:, 0])]
-        return neighbors
+        return np.array(list(reversed(neighbors)))
 
     def position_candidates(self, xy_i, neighbors, d):
         """Return a list of (x, y) coordinates that might be valid."""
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks, this is good work.

The best path forward for further optimization is to handle points in the "gutters" (maximum displacement from center) separately. With large Ns, this will be the majority of points, and if much of the processing can be skipped on points that will end up in the gutter, it will save a lot of time.

I've always felt that it isn't necessary to put too much effort into optimization in the swarming algorithm as the plot itself doesn't scale that well to large number of points (as most end up overlapping themselves in the gutter). But even when the plot isn't going to work, it's nice to be able to find out quickly!

PR #1145, opened 2017-03-28
Don't fail on PerfectSeparationError in regplot logistic regression

mwaskom:
Instead, return an array of nans as yhat which will propogate through to no regression plot (when the data have a perfect separation) or no error bands (when at least one bootstrap sample has a perfect separation).

Arguably the nans should be pruned from at least the bootstraps, but this seems like a good way to indicate "something isn't quite right with the confidence intervals".

### DIFF
--- a/seaborn/linearmodels.py
+++ b/seaborn/linearmodels.py
@@ -252,7 +252,8 @@ class _RegressionPlotter(_LinearPlotter):
             try:
                 yhat = model(_y, _x, **kwargs).fit().predict(grid)
             except glm.PerfectSeparationError:
-                yhat = np.ones(len(grid)) * np.nan
+                yhat = np.empty(len(grid))
+                yhat.fill(np.nan)
             return yhat
 
         yhat = reg_func(X, y)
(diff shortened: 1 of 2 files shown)


seaborn/linearmodels.py
```diff
@@ -246,7 +246,15 @@ def fit_statsmodels(self, grid, model, **kwargs):
...
         X, y = np.c_[np.ones(len(self.x)), self.x], self.y
         grid = np.c_[np.ones(len(grid)), grid]
-        reg_func = lambda _x, _y: model(_y, _x, **kwargs).fit().predict(grid)
+        import statsmodels.genmod.generalized_linear_model as glm
+
+        def reg_func(_x, _y):
+            try:
+                yhat = model(_y, _x, **kwargs).fit().predict(grid)
+            except glm.PerfectSeparationError:
+                yhat = np.ones(len(grid)) * np.nan
```
mwaskom: Perhaps more performant to do `np.empty(...); np.fill(...)`, I just learned.

PR #1156, opened 2017-04-10
Fix incorrect indent level in tsplot

darothen:
The variable `x` is only set when multiple traces are plotted via `data.groupby(...)`; this causes the later call to `ax.set_extent()` to fail, because it relies on that `x`. I intermittently and inconsistently ran into a bug here when analyzing a dataset, and this fix worked fine. An alternative would be to be sure to set `x` outside of the groupby loop, if for some reason it isn't entered.


mwaskom:
Closed with #1217

PR #1163, opened 2017-04-24
CI: Test against pandas 0.20.0rc1

TomAugspurger:
Time for another game of let's see what broke!

Just one failure locally (though probably some warnings, I'll followup with a PR).

Pandas issue for the thing causing the failure https://github.com/pandas-dev/pandas/issues/16107 (`pd.unique` with strings and NaN casts the NaN to a string).

### DIFF
--- a/devel_requirements_py2.txt
+++ b/devel_requirements_py2.txt
@@ -4,4 +4,3 @@ sphinx_bootstrap_theme
 numpydoc
 PIL
 nose
-pytest
--- a/devel_requirements_py2.txt
+++ b/devel_requirements_py2.txt
@@ -4,3 +4,4 @@ sphinx_bootstrap_theme
 numpydoc
 PIL
 nose
+pytest
--- a/devel_requirements_py3.txt
+++ b/devel_requirements_py3.txt
@@ -4,4 +4,3 @@ sphinx_bootstrap_theme
 numpydoc
 pillow
 nose
-pytest
--- a/devel_requirements_py3.txt
+++ b/devel_requirements_py3.txt
@@ -4,3 +4,4 @@ sphinx_bootstrap_theme
 numpydoc
 pillow
 nose
+pytest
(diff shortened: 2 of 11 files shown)


TomAugspurger:
@mwaskom rebased and squashed down to https://github.com/mwaskom/seaborn/pull/1163/commits/4d89197e94dd65d0ab80f77fc380ca3307950ba7

That commit should be mergeable. I removed all the references to the pandas RC, so the CI run isn't being tested against the pandas RC, but everything passes locally. The one issue that failed earlier is fixed on master.

The only remaining todo item is to remove imports from `pandas.core`, which is now considered private to pandas. The only one I see `remove_na` in `categorical.py`. I think I have something to replace it with, but need to do a bit more testing. Will do a followup.

mwaskom:
So the issue is that the pandas asserts (`assert_frame_equal`, etc.) are calling `pytest` functions, but that the pandas conda build doesn't depend on py.test?

TomAugspurger:
`assert_frame_equal` and friends don't need pytest, and pytest isn't a dependency of pandas (just like nose wasn't before).

The only reason I needed `pytest` for the CI runs is that the `pandas.util.testing.network` decorator is now a [pytest marker](https://docs.pytest.org/en/latest/mark.html).  If you'd prefer, I can replace that decorator with something else, and remove the requirement to have pytest to run the tests.

TomAugspurger:
I should say, I couldn't see where the `@network` decorator was actually being exercised. It's used in `test_utils.py` but I didn't see any `--skip-network` config settings or anything. Maybe it can just be deleted.

TomAugspurger:
Oh, I think I recall, it was a a try/except to see if the network was available. Let me just re-implement it quick.

TomAugspurger:
All green, sorry about the noise. No idea what was going on with the py27 failures.

PR #1167, opened 2017-05-04
Test against pandas 0.20.0rc2

TomAugspurger:


TomAugspurger:
I'm pretty sure [these](https://travis-ci.org/mwaskom/seaborn/jobs/228790701) failures are unrelated to RC2.

mwaskom:
Hm. Those are new.

mwaskom:
Looks like matplotlib has ticked up to 2.0.1 since your last PR. Could be the cause. Interesting that it doesn't fail on the minimal versions build.

TomAugspurger:
> Looks like matplotlib has ticked up to 2.0.1 since your last PR.

Good catch. I'm able to reproduce the failures locally with matplotlib 2.0.1 (regardless of pandas rc2)

mwaskom:
Oh goodie.

mwaskom:
If that environment is handy can you run [this example](http://seaborn.pydata.org/examples/factorplot_bars.html) in it and see what looks different?

TomAugspurger:
2.0.0:
mpl200

2.0.1:

mpl201

Differences are in the errorbars: The top of `second, male` moved down and all of `third, female` moved down

mwaskom:
Hm weird, a little movement of the errorbars make sense because of bootstrapping, but the test failures are about the bottom of the bars not being a 0 on the quantitative axis, which we're not seeing here.

PR #1184, opened 2017-05-22
Improvements and changes to defaults in heatmap

### LINKED ISSUE #1169: Decrease strength of assumption that heatmap data that crosses 0 is diverging
Currently it's pretty annoying to get limits that are not symmetric around 0 if the data range crosses 0.

### LINKED ISSUE #1192: Large heatmaps should skip some tick labels by default
The temporal limiting factor for larger heatmaps is drawing the ticklabels, which also are not informative when they overlap. `xticklabels` and `yticklabels` should default to `"auto"` which should do something smart to minimize drawing time/overlap.

mwaskom:
A few changes are wrapped up in here:

- Changed the behavior of `heatmap` (and by extension `clustermap`) when plotting divergent dataesets (i.e. when the ``center`` parameter is used). Instead of extending the lower and upper limits of the colormap to be symettrical around the ``center`` value, the colormap is modified so that its middle color corresponds to ``center``. This means that the full range of the colormap will not be used (unless the data or specified ``vmin`` and ``vmax`` are symettric), but the upper and lower limits of the colorbar will correspond to the range of the data. Here is an example:

```
sns.heatmap(corrmat, square=True, center=0, cmap="RdBu_r", ...)
```

image

Note how the colorbar stops at -.4 and does not include dark blue values. The recentering also respects custom colormap limits:

```
sns.heatmap(corrmat, square=True, center=0, cmap="RdBu_r", vmin=-.15, vmax=.6, ...)
```

image

- Removed automatic detection of diverging data in `heatmap` (and by extension `clustermap`). If you want the colormap to be treated as diverging (see above), it is now necessary to specify the ``center`` value. When no colormap is specified, specifying ``center`` will still change the default to be one that is more appropriate for displaying diverging data.

- Added four new colormaps, created with `viscm` for perceptual uniformity. There are two new sequential colormaps ("rocket" and "mako") and two new diverging colormaps ("icefire" and "vlag"). The colormaps and reversed ("_r") variants are automatically registered in matplotlib on seaborn import. They look like this:

colormaps

- Changed the default heatmap colormaps to "rocket" and "icefire" for sequential and diverging data, respectively.

- Changed the default tick labels in `heatmap` to `"auto"`, which estimates how many ticks can be labeled without overlap.

- Changed the `clustermap` API examples to use the brain networks dataset which is a little bit closer to the kind of dataset you would be using it with.

Closes #1169 
Closes #1192


mwaskom:
A possible TODO: in the reversed versions of the new sequential colormaps, I don't like how it tops out in a dark black and prefer a "short" version of the colormap that does not include the darkest ~40 values:

image

There's not currently a convention for describing such colormaps (akin to "_r"), nor is there a mechanism for easily cropping a colormap. One idea would be to monkey patch matplotlib's `get_cmap` function to allow a spec like "rocket_r_:10:-40", which would slice out the colors in the lut and return a new colormap (could also do floats). Possibly outside of the scope of this PR.

mwaskom:
This PR should maybe include handling of #1192

mmagnuski:
❤️ 🎉

PR #1188, opened 2017-05-26
added list input support for kdeplot

### LINKED ISSUE #1186: kdeplot doesn't coerce list into numpy array
if you pass a list of data to `kdeplot` you get

```python
    575         ax = plt.gca()
    576 
--> 577     data = data.astype(np.float64)
    578     if data2 is not None:
    579         data2 = data2.astype(np.float64)

AttributeError: 'list' object has no attribute 'astype'
```
seems like it should try to call `np.array(data)` and check that `data.ndim == 1`

pyeguy:
added `np.asarray` for input if it is a list. works with 1d and 2d lists , fixes #1186


mmagnuski:
You'd need to add a test and update the docs.

mwaskom:
Thanks

PR #1189, opened 2017-05-27
updated docs and added test to examples of kdeplot.

pyeguy:
as @mmagnuski  pointed out this should have been in #1186

### DIFF
--- a/'
+++ b/'
@@ -1,0 +1,12 @@
+white space fix in example section. 
+# Please enter the commit message for your changes. Lines starting
+# with '#' will be ignored, and an empty message aborts the commit.
+#
+# Committer: Cameron Pye <Cameron Pye>
+#
+# On branch master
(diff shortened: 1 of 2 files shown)


seaborn/distributions.py
```diff
@@ -451,8 +451,8 @@ def kdeplot(data, data2=None, shade=False, vertical=False, kernel="gau",
 
     Parameters
     ----------
-    data : 1d array-like
-        Input data.
+    data : 1d array-like or 2d array-like
```
mwaskom: This form of input is being deprecated.

'
```diff
@@ -0,0 +1,12 @@
+white space fix in example section.
```
mwaskom: What is this file?

mwaskom:
I don't think it's necessary, thanks.

pyeguy:
No worries, felt like overkill

Issue #1204, opened 2017-06-20
jointplot: specific keyerror error should be raised when passing data and column is missing


Prikers:
I think a specific `KeyError ` should be raised when data is provided and either `x` or `y` is a string but not a valid key of the DataFrame. 

This code:
```python
import pandas as pd
import numpy as np
import searbon as sns

df = pd.DataFrame(columns = list('ABCD'),
                  data = np.random.rand(100, 4))

sns.jointplot(x='A', y='NOT_A_VALID_COLUMN', data=df)
```
returns the following error:
```python
Traceback (most recent call last):

  File "<ipython-input-20-0d129181776f>", line 1, in <module>
    sns.jointplot(x='A', y='NOT_A_VALID_COLUMN', data=df)

  File "C:\Users\...\Anaconda3\lib\site-packages\seaborn\distributions.py", line 798, in jointplot
    xlim=xlim, ylim=ylim)

  File "C:\Users\...\Anaconda3\lib\site-packages\seaborn\axisgrid.py", line 1661, in __init__
    y = y[not_na]

TypeError: string indices must be integers
```
A `KeyError` would be less confusing when debugging. And the figure could be not created.

mwaskom:
Yes, the cost of input flexibility is that it's harder to do input validation ...

Prikers:
I will propose a pull request then. Thanks!

Prikers:
#1211 added

PR #1211, opened 2017-06-27
Added checks for input types for distributions.jointplot

Prikers:
Added a couple of checks for `distributions.jointplot`:
* Check if `data` is a `pandas.DataFrame` (if passed)
* Check if `x` and `y` are valid column names in `data`

Added a test unit as well

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -4,6 +4,8 @@ from distutils.version import LooseVersion
 import warnings
 from textwrap import dedent
 
+from six import string_types
+
 import numpy as np
 import pandas as pd
 import matplotlib as mpl
@@ -1649,10 +1651,17 @@ class JointGrid(object):
 
         # Possibly extract the variables from a DataFrame
         if data is not None:
-            if x in data:
-                x = data[x]
-            if y in data:
-                y = data[y]
+            msg = "'{}' is not a valid column name of 'data'"
+            if isinstance(x, string_types):
+                if x in data:
+                    x = data[x]
+                else:
+                    raise KeyError(msg.format(x))
+            if isinstance(y, string_types):
+                if y in data:
+                    y = data[y]
+                else:
+                    raise KeyError(msg.format(y))
 
         # Possibly drop NA
         if dropna:
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -800,15 +800,6 @@ def jointplot(x, y, data=None, kind="scatter", stat_func=stats.pearsonr,
         ...                   s=40, edgecolor="w", linewidth=1)
 
     """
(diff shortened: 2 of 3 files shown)


Prikers:
I am not sure to understand why `make lint` is failing for python 2 here. Anyone could help?

mmagnuski:
Travis tells you why:
`seaborn/tests/test_distributions.py:136:1: W293 blank line contains whitespace`

Prikers:
(feeling stupied...) I was not able to find the actual error message from Travis. Thanks. All checks are passed now. This should fix #1204

mmagnuski:
You'll have to wait for @mwaskom to review and accept.
I'm +1 on this but I'm just a passerby. :)

mwaskom:
There are a few problems with this implementation.

First, the input validation should happen in `JointGrid`, not `jointplot`.

Second, this PR reduces flexibility in a few ways:

- Currently it's not required that `data` be a `DataFrame`, it just has to be any object that implements `.get`. While not fully documented, that flexibility should be maintained.
- Currently, you can pass one variable as a reference to a field in `data` and one variable as an array. Or two variables as arrays and pass `data` (I'm not sure when this would be useful, but it's possible). The correct way to do the input validation would be to check if the x and y variables are string types (accounting for Py 2/3 compatibility) after trying to get them from `data`.

Prikers:
Sorry about that! I am still trying to get more familiar with the source code. I moved the input checks to `JointGrid`.

However as of today, the first checks are like this: 
```python
if data is not None:
    if x in data:
        x = data[x]
```
so if we pass a `DataFrame` in `data` and an `array` in `x` we get a `TypeError: unhashable type:  'numpy.array'`. That is why I thought we could not have those mixed-type situations. Therefore I moved the `string_type` verification upfront the verification `if x in data`.  
Does it cover all situations?

Prikers:
Sorry about that! I am still trying to get more familiar with the source code. I moved the input checks to `JointGrid`.

However as of today, the first checks are like this: 
```python
if data is not None:
    if x in data:
        x = data[x]
```
so if we pass a `DataFrame` in `data` and an `array` in `x` we get a `TypeError: unhashable type:  'numpy.array'`. That is why I thought we could not have those mixed-type situations. Therefore I moved the `string_type` verification upfront the verification `if x in data`.  
Does it cover all situations?

mwaskom:
Hm good point, I was thinking of the way things are handled in the categorical plots, which I now view as the best way of doing things: https://github.com/mwaskom/seaborn/blob/6ddedad5c9e734e4bb8751adb4efa851cc323b00/seaborn/categorical.py#L140

mwaskom:
There were a couple of other issues and it was easiest to just demonstrate rather than keep nitpicking: #1213. Thanks for catching.

PR #1219, opened 2017-07-02
Don't depend on ticks for heatmap annotations

### LINKED ISSUE #837: Annotations do not appear when xticklabels or yticklabels are False in heatmaps
In heatmaps when either xticklabels or yticklabels are false annotations do not appear on the heatmap.

```
import numpy as np; np.random.seed(0)
uniform_data = np.random.rand(10, 12)
ax = sns.heatmap(uniform_data, annot = True)
```

image

```
ax = sns.heatmap(uniform_data, annot = True, xticklabels = False)
```

image

This bug is present in 0.7.0 but is not in 0.6.0

NasaGeek:
Will now show annotations on all cells regardless of what value
x/yticklabels take. Fixes #837.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -233,7 +233,8 @@ class _HeatMapper(object):
     def _annotate_heatmap(self, ax, mesh):
         """Add textual labels with the value in each cell."""
         mesh.update_scalarmappable()
-        xpos, ypos = np.meshgrid(ax.get_xticks(), ax.get_yticks())
+        height, width = self.annot_data.shape
+        xpos, ypos = np.meshgrid(np.arange(width) + .5, np.arange(height) + .5)
         for x, y, m, color, val in zip(xpos.flat, ypos.flat,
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks — it would be good to have a test...

NasaGeek:
Added a test, let me know if you think any more would be appropriate.

mwaskom:
Test looks good but there's a PEP8 issue.

NasaGeek:
Woops. That reminds me, make test seems to fail as pyflakes can't take any flags. Are you using a different version of pyflakes? I'm on 1.5 from pip.

mwaskom:
> Are you using a different version of pyflakes? I'm on 1.5 from pip.

Yep: https://github.com/mwaskom/seaborn/blob/master/.travis.yml#L16

NasaGeek:
Should be good now, thanks for the pointers.

mwaskom:
Thanks!

Issue #1067, opened 2016-11-16
Pairplot does not raise an error when passed a Pandas series


jaypeedevlin:
I came across this issue when someone who I was teaching pairgrid mangled the syntax and passed three pandas series' as the first three variables to `pairplot()`.

The issue is that rather than raising an error, the code seems to get caught up and eventually times out.

**Version Info:**

Seaborn `0.7.1`
Matplotlib `1.5.3`
Plotting in Jupyter Notebook using `%matplotlib inline`

**To reproduce:**

```
import seaborn as sns
import pandas as pd
import numpy as np

s = pd.Series(np.random.randint(0,100,size=100))
sns.pairplot(s)
sns.plt.show()
```

mwaskom:
Closed by #1227

PR #1227, opened 2017-07-09
type check for pairplot

pyeguy:
added a check for the `data` argument in the pairplot function . Raises a `TypeError` if it is not a `DataFrame` . 

Fixes issue #1067


pyeguy:
finally installed pep8==1.5 to catch those whitespace errors before i commit.

PR #1230, opened 2017-07-09
Distplot emptycheck issue1227

### LINKED ISSUE #1136: plotting a dataframe with no rows should throw a specific ValueError
I've been teaching pandas/seaborn to novices this week and noticed the following quite a few times - minimal example:

```python
import pandas as pd
import seaborn as sns

df = pd.DataFrame({
    'fruit' : ['apple', 'apple', 'banana', 'banana'], 
    'number' : [5,6,7,8]
})

# show distribution of numbers for apple
sns.distplot(df[df['fruit'] == 'apple']['number'])

# accidentally make a typo
sns.distplot(df[df['fruit'] == 'aple']['number'])

ZeroDivisionError                         Traceback (most recent call last)
...
```

Accidentally filtering a dataframe to produce no rows (either via a typo as above or something more complex) leads to a `ZeroDivisionError`, and the hapless student starts digging through the dataframe to find zero values or missing data. Would it be possible to check the number of rows in the input dataframe before starting the plot, and raise a `ValueError` with a more helpful message?

pyeguy:
added a check to make sure array-like argument of `distplot` is not empty, raises `ValueError` if it is. I moved the array creation and check before  the `ax`  gets created so it errors before instantiating an axis. 

Fixes #1136 .

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -2004,10 +2004,10 @@ def pairplot(data, hue=None, hue_order=None, palette=None,
         ...                  diag_kws=dict(shade=True))
 
     """
-    if not isinstance(data,pd.DataFrame):
+    if not isinstance(data, pd.DataFrame):
         raise TypeError(
             "'data' must be pandas DataFrame object, not: {typefound}".format(
-            	typefound=type(data)))
+                typefound=type(data)))
 
     if plot_kws is None:
         plot_kws = {}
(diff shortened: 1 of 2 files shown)


mwaskom:
Please develop proposed changes on feature-specific branches; this PR has commits that overlap with #1227 because you are developing on the `master` branch.

pyeguy:
oh yeah, sorry i don't know what i was doing there... looks like i made a branch with the other issue # in the name (dumb typo) and then didn't even checkout properly. i'll try to fix it, let me know if you have a suggestion on how.

pyeguy:
I'm going to close this PR and open a new one with a correct name etc...

PR #1236, opened 2017-07-11
Fix legend_out with GUI backend

tkf:
There was a bug that FacetGrid.add_legend() with legend_out=True did not
work when the backend was "GUI"-based ones such as Qt5Agg and
NbAgg (checked with matplotlib 2.0.2).  Note that it was working
correctly with Agg backend.  This is likely because pyplot.draw calls
canvas.draw_idle() which may not force re-draw [1].

[1] https://github.com/matplotlib/matplotlib/blob/v2.0.2/lib/matplotlib/pyplot.py#L691

You can check the bug with the following script.

```python
def plot_with_backend(backend):
    import matplotlib as mpl
    mpl.use(backend)
    import seaborn as sns

    tips = sns.load_dataset('tips')
    g = sns.lmplot('total_bill', 'tip', tips, hue='smoker', fit_reg=False)
    g.savefig(backend + '.png')

# plot_with_backend(Agg')  # works fine w/o this PR
plot_with_backend('Qt5Agg')
```

Before this patch, it produces:
image

After this patch, it produces:
image

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -93,7 +93,7 @@ class Grid(object):
             figlegend._legend_title_box._text.set_font_properties(prop)
 
             # Draw the plot to set the bounding boxes correctly
-            plt.draw()
+            self.fig.draw(self.fig.canvas.get_renderer())
 
             # Calculate and set the new width of the figure so the legend fits
             legend_width = figlegend.get_window_extent().width / self.fig.dpi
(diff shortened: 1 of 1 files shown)


tkf:
I'm not adding the test since it seems the backend is hard-coded in testing/matplotlibrc and not sure what is the best way to implement backend-aware testing.  Perhaps the easiest way to do the testing is to prepare multiple matplotlibrc and switch them in .travis.yml?  Anyway, I prefer this PR to be merged as-is rather than drastically changing the test suite here.

mwaskom:
Nice! That's always been a vexing issue.

mwaskom:
In other exciting news, applying this change to the calls to `plt.draw` in `sns.heatmap` gets the test suite running on an interactive backend.

tkf:
Thanks for the merge!

PR #1238, opened 2017-07-14
Add example script making a Joy Divsion plot

mwaskom:
People seem really interested in these lately, anyway, it's not too hard `¯\_(ツ)_/¯`

image


mwaskom:
Change in `kdeplot` isn't essential for the example I converged on, but it was useful for a few other candidates (i.e. using a solid fill color that matchee the background, which is closest to the original album illustration).

mwaskom:
It's even simpler if you're not fussy, but where's the fun in that?

```python
g = sns.FacetGrid(df, row="g", hue="g", aspect=12, size=.5, palette="tab10")
g.map(sns.kdeplot, "x", clip_on=False, shade=True, alpha=1, bw=.25, lw=1.5)
g.add_legend(title="")
g.set_titles("")
g.fig.subplots_adjust(hspace=-.25)
g.despine(bottom=True, left=True)
g.set(yticks=[])
```

lukauskas:
How would you feel about making this a high-level function, just like`sns.violinplot` is? 
I think it would be used as much, if not more.

mwaskom:
I would be opposed.

lukauskas:
Why? maintenance overhead?

mwaskom:
I think they're fundamentally pretty gimmicky and not actually a good way to visualize data. The point of the example is mostly to take something people think is neat and use it to demonstrate how to go beyond the normal usage pattern for `FacetGrid`.

lukauskas:
I don't know... Yes it is a popular gimmick nowadays, but I think it would be useful in `seaborn`.

Particularly, it would fill in a gap where we currently have no high-level way to facet `kdeplots` without the `FacetGrid`. There is `violinplot`, but its weakness is the mirroring of the KDE around one axis. I've run into a few cases where my violin plot was breaking down because I had too many 'violins', but could have easily been readable if I just drew half a violin only. Sort of like `split=True` parameter, but without the second half. This is essentially what a joy plot does. Overlap between KDEs simply reduces the space even more.

Whether it is as a useful visualisation feature, that's subjective, but in my opinion it's great if:

1. You're only interested in distribution in one dimension;
2. The distribution cannot be summarised easily (e.g. bimodal);
3. You cannot simply plot all the points (perhaps there is too much data?);
4. The axis you are faceting is ordinal.

Twitter has a few nice examples ([1](https://twitter.com/ClausWilke/status/888794088014848000), [2](https://twitter.com/ClimateOfGavin/status/887522165196820480), [3](https://twitter.com/mikedecr/status/886376620004700160)). Even though they do not meet all the criteria above.

I think I've seen a similar visualisation in the lab as well, I think FACS uses a similar visualisation for its results, which means people familiar with it would find these easy to understand too.

mwaskom:
I remain unconvinced, but I'll remind you that `ggjoy` is not actually part of ggplot. If you think it would be useful, you should write a function that builds on this example and put it on PyPi.

PR #1270, opened 2017-08-30
typo

alvarouc:
observerations should be observations


mwaskom:
Not necessarily, what if each data point represents a different compute server? :)

PR #1283, opened 2017-09-14
update comments in regression example

phobson:
comments were about the tips dataset, but the examples users iris.


mwaskom:
I meant petal tip, obviously :)

PR #1285, opened 2017-09-17
Add lineplot function

mwaskom:
This PR adds a major new function, `lineplot`, that replaces `tsplot` while behaving much more like other seaborn functions (the categorical plotters) and adding some new capabilities (subsetting with different visual semantics on up to three dimensions, mapping the `hue` variable numerically, and handling dates properly). 

The basic signature is similar to other functions:

`lineplot(x, y, <semantic mappers>, data, <control kwargs>)` 

but with three parameters that control subsetting/semantic mapping: `hue`, `size`, and `style`. These operate independently, so you can in principle show up to five variables on one plot (but be careful: many such plots will be very hard to interpret). The multiple semantics should also make it easy to generate plots that are accessible (i.e. using more than just color to map some variable) without much effort.

Here are some quick examples. For more information, I've copied the new tutorial notebook source to a [gist](https://gist.github.com/mwaskom/bfb56b5754db27721e9fa98b61c15551).

```python
fmri = sns.load_dataset("fmri")
sns.lineplot(x="timepoint", y="signal", data=fmri)
```
image

```python
sns.lineplot(x="timepoint", y="signal", hue="event", style="event",
             markers=True, dashes=False, data=fmri)
```
image

```python
sns.lineplot(x="timepoint", y="signal", hue="region", style="event", data=fmri)
```
image

```python
sns.lineplot(x="time", y="firing_rate",
             hue="choice", size="coherence",
             legend="full", data=dots)
```
image

```python
values = np.random.randn(365, 4).cumsum(axis=0)
dates = pd.date_range("1 1 2016", periods=365, freq="D")
data = pd.DataFrame(values, dates, columns=["A", "B", "C", "D"]).rolling(7).mean()

sns.lineplot(data=data, palette="tab10", linewidth=2.5)
```
image

What's more, most of the underlying architecture is going to be re-used to support a forthcoming `scatterplot` function that will behavior similarly while drawing scatterplots instead of line plots. However because that's not implemented yet, it's possible that doing so will force some small changes in `lineplot` behavior. That means I wouldn't use `lineplot` in production quite yet, but definitely start throwing some real datasets at it.

## TODO

- [x] Clean up style- and size-related code
- [x] Clean up estimation code
- [x] Axes labeling
- [x] Legends
- [x] Decisions about default estimator/ci behavior
- [x] Update `tsplot` deprecation message
- [x] Unit tests
- [x] Release notes
- [x] Docstring / API examples
- [x] Gallery examples
- [x] Tutorial documentation

## Related TODOs outside the scope of this PR:

- Implement `scatterplot`
- Improve colormap handling with numeric ``hue`` data: current behavior uses the rcParams image colormap (like matplotlib functions), but this is often suboptimal for lines/points because it often includes high luminance colors that give very low contrast with the background. Ideally matplotlib 2.0 would have had different colormaps for image/object data, but this was unfortunately rejected.
- Improve "ticking" for brief legends
- Implement `units` for faster bootstrapping in the case of repeated measures data
- Also allow `units` to draw multiple unconnected lines with the same semantics (e.g. to plot individual subject data without differentiating)


beojan:
Can we have the ability to specify the uncertainties rather than using bootstrapping?

mwaskom:
@beojan no, that's not the way things work in seaborn

PR #1297, opened 2017-10-06
Typo lvplot docstring

joelostblom:


mwaskom:
Thanks @joelostblom  (thought *doesn't* is a perfectly valid contraction...)

PR #1308, opened 2017-10-21
Avoid turning off axes autoscaling in kdeplot

joelostblom:
This related to #1272 .

The reason the maximum axes limits were not being updated correctly is that `set_ylim` and `set_xlim` [turn off axes autoscaling by default](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/axes/_base.py#L2857-L2859). So the axes would be scaled to the first data set that was plotted with `sns.kdeplot`, but then remain at that level for all following plots into the same axes. By passing `auto=None`, the axes' current autoscaling setting is kept intact and there is no need to specify the max value.

 I initially didn't realize you already fixed this since I was running 0.8.0, but the approach in this PR might avoid future issues related to turning off the axes autoscaling, so I am posting it here in case you prefer it.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -329,9 +329,9 @@ def _univariate_kdeplot(data, shade, vertical, kernel, bw, gridsize, cut,
     # Set the density axis minimum to 0
     xmargin, ymargin = ax.margins()
     if vertical:
-        ax.set_xlim(0, max(ax.get_xlim()[1], (1 + xmargin) * x.max()))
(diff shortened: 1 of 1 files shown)


mwaskom:
Probably a better solution (or at least more consistent with the old behavior).

mwaskom:
BTW when submitting PRs please make a feature branch on your side and submit from that, rather than making changes on your own copy of the master branch.

joelostblom:
Sorry about that, I created PR #1312 from a different branch.

PR #1312, opened 2017-10-27
Avoid turning off kde axes autoscaling

joelostblom:
Moved #1308 here


mwaskom:
Thanks @joelostblom  I agree this is a much cleaner solution.

PR #1313, opened 2017-10-27
Rotate all axis labels when setting ticklabels

joelostblom:
This would change `g.set_xticklabels` and `g.set_yticklabels` to not only apply to bottom and left axes.

I recently came across [this SO post](https://stackoverflow.com/questions/41434126/rotate-x-axis-ticks-for-all-facetgrid-plots), where the `g.set_xticklabels` only rotates the bottom x-axis labels when there are x-axis labels on multiple rows in the plot. I would also expect that all visible x-axis labels are rotated when using `g.set_xticklabels`, what is your opinion on this?

The suggested fix in this PR always applies `g.set_xticklabels` to all labels instead of only the bottom labels. I did not see any immediate drawbacks to applying this to hidden labels, but if you do, I can add a condition to only apply it to visible xticklabels. I also included the same suggested fix for the y-labels.

## xticklabels

```python
tips = sns.load_dataset('tips')
g = sns.FacetGrid(tips, col='day', col_wrap=2, sharex=False)
g.map(sns.barplot, 'sex', 'tip')
g = g.set_xticklabels(rotation=45)
```
### Current behavior

image

### Suggested new behavior

image

## yticklabels

```python
tips = sns.load_dataset('tips')
g = sns.FacetGrid(tips, col='day', col_wrap=2, sharey=False)
g.map(sns.barplot, 'tip', 'sex')
g = g.set_yticklabels(rotation=45)
```

### Current behavior

image

### Suggested new behavior

image

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -882,7 +882,7 @@ class FacetGrid(Grid):
 
     def set_xticklabels(self, labels=None, step=None, **kwargs):
         """Set x axis tick labels on the bottom row of the grid."""
-        for ax in self._bottom_axes:
+        for ax in self.axes.flat:
             if labels is None:
                 labels = [l.get_text() for l in ax.get_xticklabels()]
                 if step is not None:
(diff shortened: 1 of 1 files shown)


mwaskom:
I need to think about this. I no longer feel that the set_xticklabels(rotation=45) is the best way to rotate existing ticklabels...it's too confusing given that it doesn't work in the same method on an Axes (I have a stackoverflow answer that frequently gets down voted by people who don't understand this distinction).

mwaskom:
Thanks @joelostblom; I followed up on this in #1384 

I still think that I will deprecate the approach of calling `set_xticklabels` without a list of labels to modify properties and instead add an explicit method (i.e. `rotate_ticklabels`), which should reduce confusion. But this will make things work a little better in the meantime.

PR #1326, opened 2017-11-14
fix AttributeError that occurs when using with ci='sd' and err_style='ci_bars'

rollator:
`sns.tsplot(data=all_data_frames, condition=..., unit=..., time=...,value=..., err_style="ci_bars", ci='sd')` will throw an AttributeError in seaborn version 0.8.1. This is caused by the _plot_ci_bars function which assumes that ci has a transpose.
With err_style="ci_bars"  this is not the case as ci is a tuple.

This pull request fixes said problem.


mwaskom:
I'm going to reject this because `tsplot` is officially deprecated in favor of `lineplot` and I don't have time to test that this really works for all the `tsplot` cases (I am suspicious that it might not).

PR #1328, opened 2017-11-15
set_figwidth in add_legend should forward its change

ImportanceOfBeingErnest:
Using `add_legend`, the legend overlaps with the axes. This should be prevented by enlarging the figure width via `self.fig.set_figwidth`. However, this will only take effect if the figure is actually saved, else, one would need to call `set_figwidth(..., forward=True)`.

Example code:

```
import numpy as np
import pandas as pd
import seaborn as sns

df = pd.DataFrame(np.random.rand(20,2), columns=['x', 'y'])
df["class"] = np.random.choice(["Class A","Class B"], size=20)

fg = sns.FacetGrid(data=df, hue='class')
fg.map(plt.scatter, 'x', 'y').add_legend()

plt.show()
```

produces 

forwardfalse

If instead `forward=True` is set in Line 101 the output would be

forwardtrue

as expected. This concerns only cases where the figure is not being saved beforehands, so it will not affect the inline images in notebooks or saved images. However the issue is present when showing the figure or e.g. using the `%matplotlib notebook` backend.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -98,7 +98,7 @@ class Grid(object):
             # Calculate and set the new width of the figure so the legend fits
             legend_width = figlegend.get_window_extent().width / self.fig.dpi
             figure_width = self.fig.get_figwidth()
-            self.fig.set_figwidth(figure_width + legend_width)
+            self.fig.set_figwidth(figure_width + legend_width, forward=True)
 
             # Draw the plot again to get the new transformations
             self.fig.draw(self.fig.canvas.get_renderer())


ImportanceOfBeingErnest:
Oh, this seems to fail on the travis python2.7 with `DEPS=pinned`. 
Alternatively I could imagine to replace

```
figure_width = self.fig.get_figwidth()
self.fig.set_figwidth(figure_width + legend_width)
```
by

```
figure_width, fig_height = self.fig.get_size_inches()
self.fig.set_size_inches(figure_width + legend_width,fig_height )
```

as `get_size_inches` and `set_size_inches` have not changed in recent versions of matplotlib and have the same effect.

mwaskom:
It looks like `forward=True` is now the default in matplotlib and so this is unnecessary?

ImportanceOfBeingErnest:
Correct. From matplotlib 2.2 on there would be no change needed.  
I think it depends on whether you would want to see differing behaviour depending which matplotlib version people use.  

The fully bulletproof solution for all versions would still be 

```
figure_width, fig_height = self.fig.get_size_inches()
self.fig.set_size_inches(figure_width + legend_width,fig_height )
```

Issue #399, opened 2014-12-11
Subplots and JointGrid


amelio-vazquez-reina:
Say I create a figure that will hold multiple plots, one of which I want to be a JointGrid

```
f = plt.figure(figsize=(12,6))
gs = gridspec.GridSpec(4,1 , height_ratios=[1.5,1.5,1.5, 1.5])
subplots = list(gs)
first_axis = f.add_subplot(subplots[0])
```

I would like to create a `JointGrid` plot on this axis. Is there any way to do that?

JointGrid does not seem to accept an `ax` parameter, I guess because it creates several axes (multiple of them, presumably), i.e. this doesn't work:

`sns.JointGrid(xcol, ycol, df, size=12, ax=first_axis)`

How can I use `sns.JointGrid` with subplot?

JanSchulz:
Actually there is a way to "insert a figure into an ax", if this is all done within gridspecs and subplotspecs: http://matplotlib.org/users/gridspec.html#gridspec-using-subplotspec.

If a plotting interface accepts a GridSpec and uses that to create GridSpec for the subplots:

```python
def plot_something(..., subplotspec=None):
    if subplotspec is None:
         # create a new subplotspec, by creating a 1x1 GridSpec and taking the only 
         # element in there, which will then be the whole plotting area
         subplotspec = gridspec.GridSpec(1, 1)[0]
    # create a three times three subplot area on the available subplotspec
    gs = gridspec.GridSpecFromSubplotSpec(3, 3, subplot_spec=subplotspec)
    ax00 = plt.subplot(gs[0,0]) # ax at 0,0
    ...
```

A caller, who wants to combine two plots would then have to do this:

```
import matplotlib as mpl
gs_outside = mpl.gridspec.GridSpec(1, 2)
ax_left = plt.subplot(gs_outside[0,0])
# use ax_left to plot something there, completely unrelated ... :-)
subgs_right =  gs_outside[0,0]
plot_something(..., subplotspec=subgs_right)
```

One could even use that for all plotting function, if the plotting function would guard against receiving ax as a subplotspec:

```
def plot_whatever(..., ax=None):
    if ax is None:
         # create figure + ax
    else:
          if isinstance(ax, mpl.gridspec.SubplotSpec):
              ax = plt.subplot(ax)
    # now plot something on the ax...
```

ragrawal:
Can you provide a concrete example. I am trying to render two different plots and not able to get the above solution working.

bendichter:
[...]

PR #1338, opened 2017-12-02
Pair grid as subplot

bendichter:
The PR allows you to create PairGrid type plots as a nested subplot within a pre-existing figure e.g.
```python
from matplotlib import gridspec
import seaborn as sns

fig = plt.figure(figsize=(16,8))
gs = gridspec.GridSpec(1,2, wspace=.2)

fig.add_subplot(gs[0])
iris = sns.load_dataset("iris")
g = sns.PairGrid(iris, subplot_spec=gs[1])
g = g.map(plt.scatter)
```

This feature has been frequently requested (https://stackoverflow.com/questions/35042255/how-to-plot-multiple-seaborn-jointplot-in-subplot)(#399). This would have been an overhaul in the past, but recent changes in the structure of matplotlib make this change now rather straightforward. The key change is that subplots() is now a method of Figure, which means you can call it on a pre-existing figure. The code submitted is mostly
* finding or creating a figure
* parsing a subplot_spec into gridspec_kwargs
* passing fig and gridspec_kwargs to Figure.subplots()

I'm new to contributing to open source, so let me know if there is anything else needed in this PR


mwaskom:
After looking into this some, there are enough challenges in properly implementing this that i think it's best if it remains an external hack rather than part of the API.

bendichter:
fair enough. I understand the need to use care when deciding what features to support.

bendichter:
for a better solution, see https://github.com/mwaskom/seaborn/issues/399#issuecomment-348827066

PR #1342, opened 2017-12-06
allow custom string annotations for matrix

arodiss:
This allows to pass `np.array(dtype=np.str)` as annotations for matrix 
Without this fix there is an error "unknown format code 'g' for object of type numpy.str_"


mwaskom:
This already exists:

```python
a = np.arange(6).reshape(2, 3)
s = np.array(list("abcdef")).reshape(2, 3)
sns.heatmap(a, annot=s, fmt="s")
```
image

arodiss:
Thank you

PR #1349, opened 2017-12-18
Add option for probability mass function to distplot()

manuels:
[On](https://stackoverflow.com/questions/5498008/pylab-histdata-normed-1-normalization-seems-to-work-incorrect) [popular](https://stackoverflow.com/questions/3866520/plotting-histograms-whose-bar-heights-sum-to-1-in-matplotlib) [demand](https://stackoverflow.com/questions/22241240/how-to-normalize-a-histogram-in-python) a lot of people want to plot the probability mass function instead of the probability density function, i.e. the height of the histogram bars should sum up 1 instead of the integral (height * width) sum up to 1.

This patch adds the option `sns.distplot(norm_hist='mass')` to achieve this.

### DIFF
--- a/doc/tutorial/distributions.ipynb
+++ b/doc/tutorial/distributions.ipynb
@@ -80,7 +80,23 @@
    "outputs": [],
    "source": [
     "x = np.random.normal(size=100)\n",
-    "sns.distplot(x);"
+    "sns.distplot(x)"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    "The integral of the probability density function is 1 (roughly speaking the sum of width * height for all bars). If you want the histogram heights to sum up to one, then plot the probability mass function using the `norm_hist='mass'` keyword."
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


manuels:
The failing CI tests are on the master branch (`categorical.py`) and not related to this pull request.

mwaskom:
Sorry but I think this is a bad Approach to the API. Also I don’t find “popular demand” a compelling reason to add features, and I’d characterize those stack overflow posts more as popular confusion than popular demand.

manuels:
Let's disentangle the two points:

1) Do you think plotting the probability mass function is out of the scope of seaborn?
2) The design of the API can be discussed if the first issue is resolved.

mwaskom:
It doesn't make sense in `distplot`, which draws a continuous distribution.

manuels:
The docs say:

> This function combines the matplotlib hist function (with automatic calculation of a good default bin size) with the seaborn kdeplot() and rugplot() functions.

What function do you have in mind for non-continous distributions?

Issue #1352, opened 2017-12-22
list as input in jointplot


Pompelmo:
**Not blocking error, and maybe intended, but just FYI.**
I wanted to use jointplot with two list and gives error when dropna=True.

Example code:
```import seaborn as sns
a = [1,2,3]
b = [4,5,6]
sns.jointplot(a, b)
```

produce the error:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-4-86833c44289b> in <module>()
----> 1 sns.jointplot(a, b)

~/python3.5/site-packages/seaborn/axisgrid.py in jointplot(x, y, data, kind, stat_func, color, size, ratio, space, dropna, xlim, ylim, joint_kws, marginal_kws, annot_kws, **kwargs)
   2239     grid = JointGrid(x, y, data, dropna=dropna,
   2240                      size=size, ratio=ratio, space=space,
-> 2241                      xlim=xlim, ylim=ylim)
   2242 
   2243     # Plot the data using the grid

~/python3.5/site-packages/seaborn/axisgrid.py in __init__(self, x, y, data, size, ratio, space, dropna, xlim, ylim)
   1686         if dropna:
   1687             not_na = pd.notnull(x) & pd.notnull(y)
-> 1688             x = x[not_na]
   1689             y = y[not_na]
   1690 

TypeError: only integer scalar arrays can be converted to a scalar index
```

I guess the problem is that it doesn't convert the lists in np arrays, indeed

```sns.jointplot(np.array(a), np.array(b))```
and ```sns.jointplot(a, b, dropna=False)```

work correctly.

versions:
```
python 3.5.2
matplotlib 2.1.0
seaborn 0.8.1
```

Many thanks
Caterina

ravimakhija:
It seems as if list inputs should indeed be permitted, given that they are accepted elsewhere in the seaborn library (such as for `lvplot` and `rugplot`) and also since they are converted to numpy.ndarray objects anyway after the error-causing code. 

I have added a pull request (#1356) that addresses this issue by converting the x and y parameters to numpy.ndarray objects earlier.

PR #1356, opened 2018-01-11
Fix for list inputs to JointGrid

ravimakhija:
This fix should address issue #1352. In particular, the x and y parameters are converted to arrays earlier, preventing an indexing bug in the case that they are lists.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1682,12 +1682,6 @@ class JointGrid(object):
                 err = "Could not interpret input '{}'".format(var)
                 raise ValueError(err)
 
-        # Possibly drop NA
-        if dropna:
-            not_na = pd.notnull(x) & pd.notnull(y)
-            x = x[not_na]
-            y = y[not_na]
-
         # Find the names of the variables
         if hasattr(x, "name"):
             xlabel = x.name
@@ -1696,9 +1690,18 @@ class JointGrid(object):
             ylabel = y.name
             ax_joint.set_ylabel(ylabel)
 
-        # Convert the x and y data to arrays for plotting
-        self.x = np.asarray(x)
-        self.y = np.asarray(y)
+        # Convert the x and y data to arrays for indexing and plotting
+        x_array = np.asarray(x)
+        y_array = np.asarray(y)
+
+        # Possibly drop NA
+        if dropna:
+            not_na = pd.notnull(x_array) & pd.notnull(y_array)
+            x_array = x_array[not_na]
+            y_array = y_array[not_na]
+
+        self.x = x_array
+        self.y = y_array
 
         if xlim is not None:
             ax_joint.set_xlim(xlim)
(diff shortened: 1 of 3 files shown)


mwaskom:
This approach does not work. Look at the code around the changes you made. Your change has the effect of stripping the name attributes from Pandas Series that are input. This causes a test to fail. (It looks like a few other tests are failing for unrelated reasons, I guess something changed in matplotlib).

ravimakhija:
Ah I see. OK, I've moved the cast down further after the names are retrieved. Though, it seemed reasonable to just use a new variable name instead (e.g. x_array) to make the code more explicit, but I've kept the re-ordering anyway as it seems more readable that way as well. 

The one test made to fail by the prior commit (e.g. before commit amending) is succeeding now in seaborn/tests/test_axisgrid.py. 

I'm not sure about the other tests failing, it looks like the last build job is succeeding now. So, perhaps it's like you said, a dependency changing. Also fyi, I'm not able to complete the `make test` test runs locally, as I get a "Segmentation fault 11" when getting to subtests/test_matrix.py; this may of course be unrelated to the other tests/builds failing.

ravimakhija:
Added two commits to address the other tests failing. It looks like all tests have passed with this last update. 

The other failing tests seemed to be due to two different dependency issues, one due to matplotlib and the other numpy. The two added commits and re-ordering via git rebase try and address these issues. I have tried to implement updates that will work for the latest offending package versions as well as prior versions of numpy and matplotlib. More details on these two commits are below. 

Commit title: Align doctest with new numpy array formatting

There was a change in numpy==v1.14.0 in the way numpy arrays are being formatted when printed which breaks a seaborn doctest. A commit was added to address this by having the doctest convert the numpy array to a list and then providing a list as the expected result instead of a numpy array. This fix should work for both numpy==1.14.0 and earlier releases, e.g. numpy==1.13.1. 

Commit title: Update to doctests for matplotlib compatibility

In matplotlib==2.1.1 a particular comparison in module legend.py did not support comparing numpy arrays. This was later addressed in the commit matplotlib/matplotlib@fb2c15b361cd5de5121d16699e7609376ccae72a. This update has not been released as of this writing, and the compatibility issue remains for matplotlib==2.1.1. Therefore, this seaborn commit uses a work around to remain compatible with matplotlib==2.1.1 as well as other versions. The work around creates a new figure for each doctest in pointplot and stripplot, which avoids the offending numpy array comparison in matplotlib.

mwaskom:
Please put these commits in a feature branch...it's never a good idea to rebase master directly.

ravimakhija:
OK, these commits have been put in a feature branch and a new PR #1358 should supersede this one. 

Also to clarify, I had only used rebase to re-order my three commits in this pull request so that the dependency issue fixes come first, and did not move any prior commits on the master branch. However, I would agree that using a feature branch seems cleaner. Please let me know if I've misunderstood.

PR #1358, opened 2018-01-16
Jointgrid and dependency fix

ravimakhija:
This PR supersedes PR #1356 which addressed issue #1352 as well as two dependency issues found along the way. This pull request was created to organize the three commits in a feature branch. More details can be found in the original PR, #1356.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1682,12 +1682,6 @@ class JointGrid(object):
                 err = "Could not interpret input '{}'".format(var)
                 raise ValueError(err)
 
-        # Possibly drop NA
-        if dropna:
-            not_na = pd.notnull(x) & pd.notnull(y)
-            x = x[not_na]
-            y = y[not_na]
-
         # Find the names of the variables
         if hasattr(x, "name"):
             xlabel = x.name
@@ -1696,9 +1690,18 @@ class JointGrid(object):
             ylabel = y.name
             ax_joint.set_ylabel(ylabel)
 
-        # Convert the x and y data to arrays for plotting
-        self.x = np.asarray(x)
-        self.y = np.asarray(y)
+        # Convert the x and y data to arrays for indexing and plotting
+        x_array = np.asarray(x)
+        y_array = np.asarray(y)
+
+        # Possibly drop NA
+        if dropna:
+            not_na = pd.notnull(x_array) & pd.notnull(y_array)
+            x_array = x_array[not_na]
+            y_array = y_array[not_na]
+
+        self.x = x_array
+        self.y = y_array
 
         if xlim is not None:
             ax_joint.set_xlim(xlim)
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2670,9 +2670,11 @@ stripplot.__doc__ = dedent("""\
     .. plot::
         :context: close-figs
 
+        >>> import matplotlib.pyplot as plt
         >>> import seaborn as sns
         >>> sns.set_style("whitegrid")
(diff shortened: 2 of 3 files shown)


mwaskom:
Thanks. I don’t think I understand the motivation for adding a call to ‘plt.figure’ in every docstring example. It adds a lot of noise.

ravimakhija:
I agree that it does add some noise. I will share my train of thought and welcome any ideas on improving the approach. 

To take a particular case of where doctest was failing: consider the pointplot function. The second example uses the `hue=smoke` parameter in its call. I believe the use of that parameter (there may be other causes as well) changes the _internal state of the Matplotlib objects/context_ in such a way that the error-causing code of Matplotlib v2.1.1 (e.g. the code that does not handle the numpy array comparisons) ends up running for subsequent seaborn plot calls, which are affected by that state. Therefore, the third example fails. 

If the second example of pointplot is removed and the third example remains, it will no longer fail at the point of the third example because the first test does not change the state in such a way that triggers the error-causing code of Matplotlib. However, the fourth example will probably fail since the third example does change the state in this way. 

If there is another way (e.g. perhaps a seaborn or matplotlib option that can be set) for this _internal state_ to be cleared after each example, then I think that would also work and perhaps be cleaner. That is essentially what plt.figure() is doing currently, and it is assigned to a variable just to avoid returning the figure, since then it will be expected by the doc test. 

Alternatively, if you don't mind the tests failing for only Matplotlib v2.1.1, then the original doctest code will also likely work (since as mentioned the Matplotlib issue was addressed after v2.1.1).

mwaskom:
Can you please link to the matplotlib issue describing the bug that you say only exists on 2.1.1?

The problem *arises* in the doctests because when pytest executes them, the state of the matplotlib figure persists across each chunk (unlike what happens when the docs themselves are actually built). I'm [not sure](https://stackoverflow.com/questions/47514043/use-pytest-fixture-for-each-chunk-of-doctests-in-a-single-docstring) if there's a way to avoid this. So what's happening is that most or all of the doctest examples for a given function are actually being plotted onto the same axes.

But it should in general be possible to draw multiple, e.g., point plots onto the same axes. So if the doctests are failing when they try to do that, it's not really a problem with the doctests so much as seaborn/matplotlib, and your approach of instantiating a separate figure for each chunk simply hides the problem.

If it's actually a bug in matplotlib that's isolated to a specific minor version, then the Travis build can be configured not to use that version and it will just be a transient nuisance for some users.

mwaskom:
I suspect it's this: https://github.com/matplotlib/matplotlib/issues/10053

ravimakhija:
Indeed, that looks like the issue. I had found the issue via PR matplotlib/matplotlib#10031, where there is a commit titled "Make comparison work with arrays" and also where user dstansby comments:

> The problem is the original 'color comparison' code in the figure legend method didn't allow for colors to be numpy arrays. Long term I think this should be avoided by a general function for comparing colors, which I've put in #10032

mwaskom:
OK if it's a problem on a specific version of matplotlib and will be fixed in 2.1.2 then I think it's best to just skip testing on that version.

I've also opened an [issue on pytest](https://github.com/mwaskom/seaborn/blob/master/doc/sphinxext/plot_directive.py) looking for guidance on configuring a fixture to properly close figures between each chunk the way that the matplotlib sphinx extension does when actually building the docs.

mwaskom:
Thanks @ravimakhija. I cherry picked your `JointGrid` commit to #1383 but I'm skipping the changes to the tests for now because the current version of matplotlib no longer causes errors and the right way to handle the figure cleanup is through a `pytest` plugin.

Issue #1278, opened 2017-09-09
Categorical scatterplot docs need updates


mwaskom:
The `split` parameter was renamed to `dodge` to match the other categorical plots, but the docstring parameter information still references `split`.

ccatalfo:
I submitted a pull request which fixes this: https://github.com/mwaskom/seaborn/pull/1377

LeData:
`split `is still present in the doc, in the detailed list of parameters.

PR #1377, opened 2018-02-22
Update docstring parameters from split to dodge for swarmplot and str…

ccatalfo:
Fixes https://github.com/mwaskom/seaborn/issues/1278 (closes https://github.com/mwaskom/seaborn/pull/1279 , which fixes it partially, only for stripplot) by renaming the split->dodge in doctring for swarmplot and stripplot.


mwaskom:
Thanks @ccatalfo. Fixes for the (unrelated) test failures should be on master so I will merge (and hope nothing breaks!)

PR #1394, opened 2018-03-29
`sns.scattermap` implementation

lukauskas:
One more PR from me today.

Here I propose a `sns.scattermap` function, which behaves in the same way as heatmap, but instead of filled "boxes" that heatmap uses, it plots dots using `plt.scatter`. In R community this is sometimes called a `dotplot`.

Some example usage below:

Heatmap:

```python
flights = sns.load_dataset("flights")
flights = flights.pivot("month", "year", "passengers")
ax = sns.heatmap(flights)
```
image

Scattermap:

```python
flights = sns.load_dataset("flights")
flights = flights.pivot("month", "year", "passengers")
ax = sns.scattermap(flights)
```

image

Heatmap (no `size` option):

```python
with sns.axes_style('white'):
    corr = np.corrcoef(np.random.randn(10, 200))
    ax = sns.heatmap(corr, vmax=.3, square=True, 
                     linewidth=.5, linecolor='black')
```

image

Scattermap (with dynamic `marker_size`, and marker set to square):

```python
with sns.axes_style('white'):
    corr = np.corrcoef(np.random.randn(10, 200))
    ax = sns.scattermap(corr, vmax=.3, square=True, 
                        linewidth=.5, linecolor='black',
                        marker_size=np.abs(corr)*300,
                        marker='s')
```
image

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -18,7 +18,7 @@ from .utils import (despine, axis_ticklabels_overlap, relative_luminance,
 from .external.six import string_types
 
 
-__all__ = ["heatmap", "clustermap"]
+__all__ = ["heatmap", "clustermap", "scattermap"]
 
 
 def _index_to_label(index):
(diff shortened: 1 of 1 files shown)


asalt:
This is really cool - you can do fun things like this:

clustermap_circle_col_colors_example

mwaskom:
I'm a little bit more interested in this than I was last week but I'm going to need better arguments than that it would be "fun" or "cool". How does it actually make the visualization more effective?

asalt:
I'd argue that it is more comprehensible changing the col_colors from a heatmap to circles as above. Here's a [https://www.nature.com/articles/ncomms14864/figures/3](figure) that is arguably more readable with the use of circles (or blanks) than a heatmap would be where all colors would touching.

I think in combination with the use cases stated above does this become appealing.

mwaskom:
What makes it arguable? Why are circles more readable than squares?

mwaskom:
If the main benefit (spacing between cells) can be achieved with the existing parameters, then there is not a very compelling reason to add a whole new function!

asalt:
I think the main benefit is the extra dimension you get with the variable size of the markers, which as far as I am aware is not available in an existing function.

lukauskas:
Yup, I guess the visual spacing could be emulated with `linewidth=4` or something. 
But obviously the main benefit is extra dimension plus the choice of glyph.

The extra dimension is maybe not useful the way row/column annotations are currently implemented, but it's useful in general heatmaps.

mwaskom:
I'm going to close because, between `sns.scatterplot` and matplotlib's native handling of categorical axes, it's possible to make these plots with current tools:

```python
flights = sns.load_dataset("flights")
sns.relplot(x="year", y="month",
            hue="passengers", size="passengers",
            palette="mako_r", sizes=(50, 350),
            data=flights)
```
image

Issue #1262, opened 2017-08-22
FigureCanvasPdf object has no attribute get_renderer


marcelm:
Hi, commit 7b445822bc4c1edf22c8e8e11f67e632ffb90f86 introduced a regression when plotting a clustermap to a PDF.

The following code (based on the example code) works when using seaborn 0.8.0:
```python
import matplotlib
matplotlib.use('pdf')
import seaborn as sns
sns.set(color_codes=True)
iris = sns.load_dataset("iris")
species = iris.pop("species")
g = sns.clustermap(iris)
g.savefig("out.pdf")
```
On master, the following error is raised:
```
Traceback (most recent call last):
  File "plot.py", line 7, in <module>
    g = sns.clustermap(iris)
  File ".../seaborn/matrix.py", line 1299, in clustermap
    **kwargs)
  File ".../seaborn/matrix.py", line 1129, in plot
    row_linkage=row_linkage, col_linkage=col_linkage)
  File ".../seaborn/matrix.py", line 1024, in plot_dendrograms
    ax=self.ax_row_dendrogram, rotate=True, linkage=row_linkage)
  File ".../seaborn/matrix.py", line 749, in dendrogram
    return plotter.plot(ax=ax)
  File ".../seaborn/matrix.py", line 700, in plot
    ax.figure.draw(ax.figure.canvas.get_renderer())
AttributeError: 'FigureCanvasPdf' object has no attribute 'get_renderer'
```
I need to choose the pdf backend at the beginning to ensure that I don’t get error messages when I run the script non-interactively (no `DISPLAY` variable).

mwaskom:
> I need to choose the pdf backend at the beginning to ensure that I don’t get error messages when I run the script non-interactively (no DISPLAY variable).

Try the `Agg` backend. Also you'll want to rasterize the heatmap if you're plotting to pdf.

marcelm:
Thanks, it works with `use('agg')`. Thanks also for recommending `rasterized=True`, I had forgotten about that!

(If I understand correctly, matplotlib switches to the pdf backend anyway during `g.savefig("out.pdf")`, so it’s a bit weird this works.)

blaiseli:
[...]

PR #1397, opened 2018-04-06
PDF backend not compatible with using get_renderer

blaiseli:
I introduced a check on the backend, so that PDF backend can still be used.

See https://github.com/mwaskom/seaborn/issues/1262

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -311,7 +311,11 @@ class _HeatMapper(object):
         ytl = ax.set_yticklabels(yticklabels, rotation="vertical")
 
         # Possibly rotate them if they overlap
-        ax.figure.draw(ax.figure.canvas.get_renderer())
+        if mpl.get_backend().upper() == "PDF":
+            # https://github.com/mwaskom/seaborn/issues/1262
+            plt.draw()
+        else:
+            ax.figure.draw(ax.figure.canvas.get_renderer())
         if axis_ticklabels_overlap(xtl):
             plt.setp(xtl, rotation="vertical")
         if axis_ticklabels_overlap(ytl):
@@ -696,7 +700,11 @@ class _DendrogramPlotter(object):
         ytl = ax.set_yticklabels(self.yticklabels, rotation='vertical')
 
         # Force a draw of the plot to avoid matplotlib window error
-        ax.figure.draw(ax.figure.canvas.get_renderer())
+        if mpl.get_backend().upper() == "pdf":
+            # https://github.com/mwaskom/seaborn/issues/1262
+            plt.draw()
+        else:
+            ax.figure.draw(ax.figure.canvas.get_renderer())
         if len(ytl) > 0 and axis_ticklabels_overlap(ytl):
             plt.setp(ytl, rotation="horizontal")
         if len(xtl) > 0 and axis_ticklabels_overlap(xtl):
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -700,7 +700,7 @@ class _DendrogramPlotter(object):
         ytl = ax.set_yticklabels(self.yticklabels, rotation='vertical')
 
         # Force a draw of the plot to avoid matplotlib window error
-        if mpl.get_backend().upper() == "pdf":
+        if mpl.get_backend().upper() == "PDF":
             # https://github.com/mwaskom/seaborn/issues/1262
             plt.draw()
         else:


mwaskom:
I'd like to understand this problem a bit more before merging any changes. The following code works perfectly fine for me:

```python
import numpy as np
import seaborn as sns
x = np.random.randn(10, 10)
g = sns.clustermap(x)
g.savefig("test.pdf")
```

PaulRivaud:
I am encountering a similar issue.
I work on headless servers and need to use non-interactive backends.
Some of my imports using python 3.6:
```python
import matplotlib
matplotlib.use('PDF')
import seaborn as sns
```
Generating a pdf:
```python
ax = sns.heatmap(df, annot=True, fmt=".0%", cmap=params['classify']['cell_classification']['table_plot_cmap'])

plt.xlabel('Predicted sample')
plt.ylabel('Cells origin')
plt.title('Cell classification accuracy')
plt.xticks(rotation='vertical')
plt.yticks(rotation='horizontal')
plt.grid(False)
plt.tight_layout()

plt.savefig(os.path.join(dir_, 'cell_classification_table.pdf'))
```

Leads to this error:
```python
  File "<stdin>", line 1, in <module>
  File "/Users/paul/Documents/graham/methods/Classify.py", line 182, in ClassifyCells
    plot_cells_table(Res, labels, params)
  File "/Users/paul/Documents/graham/methods/Classify.py", line 144, in plot_cells_table
    ax = sns.heatmap(df, annot=True, fmt=".0%", cmap=params['classify']['cell_classification']['table_plot_cmap'])
  File "/Users/paul/anaconda/envs/py36/lib/python3.6/site-packages/seaborn/matrix.py", line 528, in heatmap
    plotter.plot(ax, cbar_ax, kwargs)
  File "/Users/paul/anaconda/envs/py36/lib/python3.6/site-packages/seaborn/matrix.py", line 315, in plot
    ax.figure.draw(ax.figure.canvas.get_renderer())
AttributeError: 'FigureCanvasPdf' object has no attribute 'get_renderer'
```

mwaskom:
That doesn't really answer the question of what is breaking and what the best way to get it to work is. Calling `plt.draw()`, which has its own problems, isn't it.

blaiseli:
I finally managed to create a minimal example illustrating my problem:

```python
import numpy as np
import matplotlib as mpl

# use_pdf = False
use_pdf = True
if use_pdf:
    mpl.use("PDF")

# Suppose we need usetex, for instance to use colours in text
# (see https://stackoverflow.com/a/42768093/1878788)
from matplotlib.backends.backend_pgf import FigureCanvasPgf
mpl.backend_bases.register_backend('pdf', FigureCanvasPgf)
TEX_PARAMS = {
    "text.usetex": True            # use LaTeX to write all text
}
mpl.rcParams.update(TEX_PARAMS)

from matplotlib.backends.backend_pdf import PdfPages

import pandas as pd

from seaborn import matrix as mat


rs = np.random.RandomState(sum(map(ord, "heatmap")))
x_norm = rs.randn(4, 8)
letters = pd.Series(["A", "B", "C", "D"], name="letters")
df_norm = pd.DataFrame(x_norm, index=letters)

def test_heatmap(data):

    g = mat.heatmap(data)
    plt.savefig("/tmp/test_heatmap.pdf")

def test_pdfpages(data):

    pp = PdfPages("/tmp/test_pdfpages.pdf")
    data.T.plot("A", "B")
    plt.savefig(pp, format="pdf")
    data.T.plot("A", "C")
    plt.savefig(pp, format="pdf")
    pp.close()


if use_pdf:
    test_pdfpages(df_norm)
    # No error
    test_heatmap(df_norm)
    # AttributeError: 'FigureCanvasPdf' object has no attribute 'get_renderer'
else:
    test_heatmap(df_norm)
    # No error
    test_pdfpages(df_norm)
    # ValueError: filename must be a path or a file-like object
```

There's an incompatibility between using tex to render text and using `PdfPages` to produce multipage pdf documents that is solved when using "PDF" as backend, but this causes the `AttributeError` in `seaborn.matrix._HeatMapper.plot`.

mwaskom:
I don't seem to have any problem using PdfPages without your change:

```python
from matplotlib.backends.backend_pdf import PdfPages
with PdfPages('/Users/mwaskom/Desktop/test.pdf') as pdf:
    g = sns.clustermap(corrmat, rasterized=True, center=0, vmax=.5, cmap="vlag")
    pdf.savefig(g.fig)
```

This runs without error and generates a nice plot.

mwaskom:
Oh I just saw that you have some problem with tex too.

blaiseli:
@mwaskom I think the real issue here is that to use tex and PdfPages at the same time, I need `mpl.use("PDF")`, which is incompatible with `heatmap`. Now what is the goal of https://github.com/mwaskom/seaborn/commit/7b445822bc4c1edf22c8e8e11f67e632ffb90f86? Is it useful when backend is "PDF"?

mwaskom:
> Now what is the goal of 7b44582? 

See https://github.com/mwaskom/seaborn/pull/1236

blaiseli:
I'm not sure I fully grasp the problem, but it seems to me that not all backends need this switch to `ax.figure.draw`. Am I correct that keeping `plt.draw` only when the backend is "PDF" (as I tried to do in my pull request) will not interfere with the resolution of this "GUI-backend"-related issue?

mwaskom:
No, I would like to avoid `plt.draw` and use the object-oriented interface.

PR #1415, opened 2018-04-23
Speed up rugplot for large inputs

ttbrunner:
Hi, I recently noticed that rugplot is extremely slow when the input array is large. On my computer, I can't plot any series with more than 20k datapoints. This PR should address the issue.

**Problem:**
I found out that ax.axvline() / axhline() both call ax.autoscale_view(), so calling these functions for every datapoint in the series causes the runtime of rugplot to become quadratic.

**Proposed fix:**
Use ax.plot() to draw the entire array of lines at once. This delivers the same result and is much faster. To the best of my knowledge the calls should be equivalent, but I'm not a Matplotlib expert (yet). So if a different solution would be preferable - I'd appreciate your feedback.


mwaskom:
My general observation that a rugplot is not particularly useful with anything close to 20k observations remains, however.

ttbrunner:
Thanks for your fast reply, and for linking the issues. Of course somebody else had it figured out already...

Understood about the usefulness with large datasets, but for me personally it still holds some merit, because:
- I have some distributions where most samples bunch up in the middle, but even at high sample counts the outliers are still quite sparse. So it's still useful for investigating concentrations in specific regions.
- Often I run plots with low sample sizes, where rugplot is very helpful. I sometimes run experiments interactively, epeat an experiment with greatly increased sample size, but after waiting some time I discover that actually rugplot has frozen my script, and I need to redo it. Of course one could change the experiment layout, but still it's a minor nuisance that can be easily fixed.

ttbrunner:
I could either add a LineCollection directly, or use the vlines / hlines call instead. Both work with the axes transform and generate the correct result - as far as I can tell from my use cases. Which one would you prefer?

ttbrunner:
Thanks, I changed it to use LineCollection, which is indeed slightly faster (17s vs 14s for a rugplot of 1M observations) for me.

ttbrunner:
Is there is anything I can do to help you out with verifying this change? I've been using the branch for some time now and it seems good to me. If you have an idea about how to best write tests for this kind of plot, I'd be happy to oblige.

mwaskom:
Hi @ttbrunner thanks for the PR and sorry for the delay, I needed to find time to thoroughly test this.

I found a couple of small problems:

- In your implementation, the axes didn't get autoscaled to show the whole rug. This was easy to fix.
- A trickier problem was that switching to `LineCollection` means a few kwarg abbreviations that `plt.plot` would have accepted would now cause errors. To avoid breaking existing code, I added some kwarg normalization.

I am still a bit concerned that this might break some people's code, so I looked into pursuing your original idea, given that the timing was not too different.

Interestingly, I compared the two approaches and found that your initial idea (drawing a single line) actually generates the matplotlib object much faster, but then loses even more time rendering on most backends. On some backends it actually crashed on large arrays. But the main reason that approach would not work is that you lose the ability to use alpha blending to deal with over-plotting (because lines don't blend with themselves, apparently). 

I think the speedup for moderately sized arrays is probably worth it in the end, but apologies to people that ended up here because their code expected `rugplot` to draw a bunch of `Line2D`s.

mwaskom:
Dang I guess `normalize_kwargs` postdates the minimum supported matplotlib version.

mwaskom:
Sigh. Welcome to testing a package that has to maintain compatibility with many versions of its dependencies.

ttbrunner:
Thanks for your help, and for taking the time. Also, I learned something new :)

PR #1419, opened 2018-05-04
ENH: add stacked option to barplot

CRiddler:
I know you're not a fan of stacked bar plots, but they do have their uses and I feel that seaborn is more than capable of supporting them. I've added a "stacked" kwarg to the `barplot` function as well as the underlying code to properly plot stacked bars. I've kept in plotting of confidence intervals on top of the stacked bars (even though if you want CI's, then you probably shouldn't be using stacked bars…).

The biggest issue I see is that if you want to use a FacetGrid with the stacked barplot, then you need to partial `sns.barplot` with `stacked=True`. This also means that `factorplot` is unable to support stacked barplots while `stacked` is merely an argument. A remedy to this would be to make a separate `stacked_barplot` function, but I'm not sure if its worth copying/pasting all that code as of right now.

A quick test to see how stacked barplots come out with this PR:

    import seaborn as sns

    df = sns.load_dataset('tips')

    fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 6))
    sns.barplot(x='day', y='tip', hue='sex', data=df, ax=ax1)
    sns.barplot(x='day', y='tip', hue='sex', data=df, ax=ax2, stacked=True)

    plt.show()

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1596,7 +1596,7 @@ class _BarPlotter(_CategoricalStatPlotter):
     def __init__(self, x, y, hue, data, order, hue_order,
                  estimator, ci, n_boot, units,
                  orient, color, palette, saturation, errcolor,
-                 errwidth, capsize, dodge):
+                 errwidth, capsize, dodge, stacked):
         """Initialize the plotter."""
         self.establish_variables(x, y, hue, data, orient,
                                  order, hue_order, units)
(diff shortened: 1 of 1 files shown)


mwaskom:
Sorry @CRiddler but, as you seem aware,  I have a dim view of stacked bar plots. They are about as bad as, if not worse than, pie charts.

The only case in which I feel that they are acceptable is when you are plotting a "part/whole" relationship, which isn't supported in general by the `hue` api but is easy enough to do by calling `barplot` twice.

In light of that, I'm going to reject this PR. Sorry.

mwaskom:
By the way [this blog post](https://eagereyes.org/techniques/stacked-bars-are-the-worst) links to some actual data showing that stacked bars are worse than pie charts.

PR #1421, opened 2018-05-07
Fix doc building for sphinx >= 1.6

tobycheese:
I get the following error when building the documentation:

```
$ make html 
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.7.4

Extension error:
Could not import extension ipython_directive (exception: cannot import name 'Directive')
make: *** [html] Error 2
```

According to the release notes of [Sphinx 1.6](http://www.sphinx-doc.org/en/master/changes.html#id35):

> sphinx.util.compat.Directive class is now deprecated. Please
> use instead docutils.parsers.rst.Directive

Please not that in order to successfully build, I also [downgraded matplotlib](https://github.com/matplotlib/matplotlib/pull/10881) for now with

```
pip uninstall matplotlib
pip install -Iv https://github.com/matplotlib/matplotlib/archive/v2.2.1.tar.gz
```

### DIFF
--- a/doc/sphinxext/ipython_directive.py
+++ b/doc/sphinxext/ipython_directive.py
@@ -124,7 +124,7 @@ except ImportError:
 import sphinx
 from docutils.parsers.rst import directives
 from docutils import nodes
-from sphinx.util.compat import Directive
+from docutils.parsers.rst import Directive
 
 # Our own
 from traitlets.config import Config


tobycheese:
Oh, and here is the error I get without above change. From within doc/

```
$ make html 
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.7.4

Extension error:
Could not import extension ipython_directive (exception: cannot import name 'Directive')
make: *** [html] Error 2
```

mwaskom:
I'm not sure that the current version of the docs actually use the ipython plugin.

mwaskom:
> Please not that in order to successfully build, I also downgraded matplotlib for now with

Is this link meant to explain why downgrading is necessary? It seems unrelated.

tobycheese:
The matplotlib thing is not related to the ipython_directive issue. I just mentioned it because it is another thing I needed to do in order to successfully build the docs (maybe helps someone). The second line of the linked PR references matplotlib issue #10874, which is what inhibited me from building.

I couldn't build the latest version of the docs without my commit. If it works for you, I am confused? I tried with the master branch.

mwaskom:
Thanks for catching this ... it seemed like the cleanest fix was to remove the unused dependencies.

tobycheese:
Yes, thanks, I agree. Sorry for not having followed up on this, but life interfered...

PR #1434, opened 2018-05-16
allow for combinations of stick/box/point/quart

kilojoules:
I wanted to make a violin plot with inner quartiles and points so I made these changes to allow users to specify combinations of stick/box/point/quart. Useful combinations are `pointstick`, `boxstick`, `boxquart`, and `quartpoint`.


mwaskom:
Please send PRs from feature branches, not changes directly on master.

I'm not convinced this is useful given the existence of `stripplot` (which post-dates the "point" inner style.

PR #1448, opened 2018-06-05
Tutorials minor fixes

joelostblom:
I was rereading the docs and noticed a few things so I went through them all briefly and include some minor fixes here. Let me know if you want to remove any of the commits. A couple of general comments:

1. In the [categorical data tutorial](https://seaborn.pydata.org/tutorial/categorical.html) you are referring to the *previous* tutorial on linear regression visualization. However, they way these are [ordered on the tutorial page](https://seaborn.pydata.org/tutorial.html), the categorical data tutorial comes before the linear regression tutorial.
2. Whenever I click a link from the tutorial page to get into one of the headings of a specific tutorial, the section heading is hidden by the page banner when that page loads. Subsequent clicks on the headings inside a specific tutorial renders the respective heading just below the banner as expected. Refreshing the page makes the banner cover the section heading again. Not sure if this is a simple fix that is worth the effort.


mwaskom:
> In the categorical data tutorial you are referring to the previous tutorial on linear regression visualization. However, they way these are ordered on the tutorial page, the categorical data tutorial comes before the linear regression tutorial.

I intend to do a fairly thorough revision of the tutorials before a 0.9 releases, so hopefully things like this will get cleaned up.

> Whenever I click a link from the tutorial page to get into one of the headings of a specific tutorial, the section heading is hidden by the page banner when that page loads. Subsequent clicks on the headings inside a specific tutorial renders the respective heading just below the banner as expected. Refreshing the page makes the banner cover the section heading again. Not sure if this is a simple fix that is worth the effort.

I think you're referring to this issue: https://github.com/mwaskom/seaborn/issues/610

Someone submitted a fix but I don't know if it never actually worked or broke again.

> Maybe I am misunderstanding, but why does it not count that you can do it via ax.spine?

Those rc params were added at some time well after `despine` and this tutorial were written.

> Include lvplot

This is fine as is your other PR but `lvplot` is kind of intentionally unadvertised because it was a provisional addition and I was never fully committed to supporting it.

mwaskom:
Thanks!

mwaskom:
> Whenever I click a link from the tutorial page to get into one of the headings of a specific tutorial, the section heading is hidden by the page banner when that page loads. 

FYI @joelostblom finally think I've solved this with https://github.com/twbs/bootstrap/issues/1768#issuecomment-46519033

joelostblom:
@mwaskom That's great! Thanks for letting me know!

PR #1450, opened 2018-06-09
Making the error message for jet more informative

rasbt:
Just stumbled upon discussions regarding the rather brief error message when a user attempts to use 'jet' as a color map. To make the `ValueError` more informative, and to provide some basic information to people who are not (yet) aware of jet's issues, I modified the error message.


mwaskom:
Thanks! But I like the old message.

PR #1460, opened 2018-06-17
Avoid error when backend has no canvas renderer

mwaskom:
Closes #1397


zachary822:
Is it possible to release this change to pypi?

Issue #1440, opened 2018-05-30
Improve multiple semantic legend titles


mwaskom:
Currently the legends for lineplot (#1285) and scatterplot (#1436) don't indicate what variable each of the levels represent. I punted on this but it needs a solution. One option is to cram everything into the legend title but that's not ideal. I think doing something with legend entries that have invisible handles and and then bumping the legend text so it looks centered would work:

```python
f, ax = plt.subplots()

h1 = ax.scatter([], [], s=0, label="Title 1")
h2 = ax.scatter([], [], label="level 1")
h3 = ax.scatter([], [], s=0, label="Title 2")
h4 = ax.scatter([], [], label="level 2")

legend = ax.legend()
for t in legend.get_texts()[::2]:
    xfm = transforms.offset_copy(t.get_transform(), ax.figure, x=-10, units="points")
    t.set_transform(xfm)
```

![image](https://user-images.githubusercontent.com/315810/40725495-95896280-63f1-11e8-96d6-3d1077e4a501.png)

But getting this to be really robust is going to be tricky.

CRiddler:
Had my eye on this one for a few days now, and I just got some time to look into it, so here are some thoughts.

> I think doing something with legend entries that have invisible handles and and then bumping the legend text so it looks centered would work

This was my thought at first too, however even when you specify `s=0` and/or `marker=""` a legend marker is still drawn and is technically visible. At least based on your example, instead of looking for a marker thats invisible, we can look for a marker with a size of 0 (or an alpha of 0 if we want to go that route) and set it to be invisible. Once we set the marker to be invisible, we can set the legend column(s) to have a center align and that will properly align the elements that no longer have a visible marker. Copy paste example:

    import matplotlib.pyplot as plt
    import matplotlib

    f, ax = plt.subplots()

    h1 = ax.scatter([], [], alpha=0, label="Title 1") #example with alpha=0
    h2 = ax.scatter([], [], label="level 1")
    h3 = ax.scatter([], [], s=0, label="Title 2") #example with size=0
    h4 = ax.scatter([], [], label="level 2")

    legend = ax.legend()

    for vpack in legend._legend_handle_box.get_children():
        vpack.align = 'center'
        
        for hpack in vpack.get_children():
            draw_area, text_area = hpack.get_children()
            for collection in draw_area.get_children():
                alpha = collection.get_alpha()
                sizes = collection.get_sizes()
                if alpha == 0 or all(sizes == 0):
[...]

PR #1465, opened 2018-06-21
Omnibus documentation updates for v0.9

mwaskom:


chris1610:
Like I said in my tweet, I really like the new intro. It does a good job of orienting people so thanks for all that hard work. I know it is really challenging to write docs!

Here are a few observations that I noted as I went through it:

- In your first relplot example, I think it would be easier for people to follow if you let them know what relplot stood for. Something like "All of this was accomplished using a single call to the seaborn function relplot(). This function name is short for 'relationship plot' and allows the user to quickly plot scatter and line plots with multiple variables." I think giving this hint will help orient new users. I suspect most will know what a line or scatter plot is but there's a bit of mental translation we want to make with relplot.

- In the "dots" line plot example, I recommend showing the first 5 rows of data as a table like you do at the end of the page for the "tips" data. The "tips" data is a little more self-explanatory (in the first example) but I think the dots data needs to be shown in order for the plot to make since. I know that I went back and forth with the dots lineplot before I understood what it was telling me.

- When you introduce the "kind" parameter, I would expand a bit for relplot. Something like this: ' Some seaborn functions have a convenient kind parameter to let you easily swap the visual representation of the data. For relplot, the default is "scatter" but you can also use "line" to create a line chart.'

- For the statistical fmri plot, it might be good to show the first 5 rows of data as well. It's not quite clear to me how the data is being averaged. Seeing the table might make it more intuitive.

I like the work you put into figure-level vs axes-level functions. That is an important distinction that takes some time to sink in. I think you captured it well.

My final thought is that you did not spend much time talking about color palettes. I know that's a huge part of seaborn's appeal. An example of creating and using palettes might be nice.  The page is getting long so I'll respect the decision if you think it's too much to add at this point.

Once again, thanks for doing all this work!

mwaskom:
Thanks a lot, @chris1610, all great suggestions.

mwaskom:
> I know that I went back and forth with the dots lineplot before I understood what it was telling me.

Part of the reason this is confusing ATM is probably that work on adding subtitles to legends with multiple semantic variables is not yet complete (https://github.com/mwaskom/seaborn/issues/1440)

PR #1466, opened 2018-06-21
Deprecate JointGrid statistical annotation

mwaskom:


canthonyscott:
I too have found this to be useful. Is there going to be a replacement for this feature or a better way I should do it?

PR #1476, opened 2018-06-29
Change so that seaborn.lvplot can take alpha argument.

llzenoll:


mwaskom:
Line opacity can be changed post-hoc like so:

```python
ax = sns.lvplot(x="day", y="total_bill", data=tips)
plt.setp(ax.lines, alpha=1)
```
image

PR #1480, opened 2018-07-05
Intersphinx

arokem:
Adds intersphinx to the documentation configuration. 

Also: changes to the introduction notebook to link to MPL and Pandas documentation.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -35,6 +35,7 @@ extensions = [
     'sphinx.ext.coverage',
     'sphinx.ext.mathjax',
     'sphinx.ext.autosummary',
+    'sphinx.ext.intersphinx',
     'matplotlib.sphinxext.plot_directive',
     'gallery_generator',
     'numpydoc',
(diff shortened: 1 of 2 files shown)


arokem:
If you would rather add the intersphinx config against `master`, we can also just cherry-pick the first of these commits into another/that branch.

mwaskom:
Thanks @arokem this works more smoothly than I expected! I guess sphinx embeds some metadata in the docs that lets this work?

I don't understand why it's apparently causing the tsplot doctests to fail, though.

arokem:
Yeah - each project has an `objects.inv` file that maps objects to their URI.  

Here's seaborn's file: https://seaborn.pydata.org/objects.inv

Did you just run the CI again? Looks like it's passing now?

mwaskom:
> Did you just run the CI again? Looks like it's passing now?

Yep I couldn't reproduce locally so I decided kicking the side of Travis and trying again might work :)

PR #1490, opened 2018-07-12
Rename factorplot to catplot and change the default plot kind to strip

mwaskom:
Moderately disruptive but well-intentioned changes herein:

First:

I've decided to abandon the original R-inflected name for `factorplot` and change it to `catplot`, which better corresponds to the appropriate terminology in pandas (and the terminology that is used more broadly in seaborn).

Along with a better name, it gives the opportunity for a less-disruptive approach to changing the default kind to "strip", which is a much better place to start with categorical plotting. (Originally factorplot *only* drew what are now called point plots, and it kept that default when it was expanded to a more general interface to categorical plots).

The other part of the change is to make `jitter=True` the default in `stripplot`. I found myself setting this 98% of the time I made a stripplot, which is a strong argument for making it a default.

Also as part of the change, the default `height` is increasing to `5` to match `lmplot` and the default for the new `relplot`.

`factorplot` will issue a warning but otherwise pass its arguments down to `catplot` (setting `kind="point"` and `height=4` as defaults) so existing code won't break. I don't plan to remove the `factorplot` stub any time soon.

Second:

To try to pack most of the disruptive changes I've had in mind into the same release, I've also changed the name of `lvplot` to `boxenplot`. `lvplot` was never a good name because it's not clear what it stands for, and even when it does, "letter value plot" was not a good name because the concept of letter values is (at least in my experience) quite obscure. So I expect that a name that describes what the plot looks like will make it fit in better with "box" and "violin" plot. (If you're wondering, "boxen" is of course the plural of box).

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -30,7 +30,7 @@ Categorical plots
     swarmplot
     boxplot
     violinplot
-    lvplot
+    boxenplot
     pointplot
     barplot
     countplot
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -25,7 +25,7 @@ Categorical plots
 .. autosummary::
     :toctree: generated/
 
-    factorplot
+    catplot
     stripplot
     swarmplot
     boxplot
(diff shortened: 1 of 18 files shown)


mmagnuski:
That's great news! I hope `puppyplot` and `catplot` will get on together. 😄

PR #1494, opened 2018-07-15
Added legend (label) to pointplot

jeffysonar:
import pandas as pd
import seaborn as sns
df = pd.DataFrame([[1,2,3,4,5],[2,4,5,6,8]],index=['one','two']).transpose()
splot = sns.pointplot(x='one',y='two',data=df,label='1v/s2 ')
splot.legend()
 would throw a warning from matplotlib, as no labels were passed

 one way of adding legend would be
 initliase fig,ax = plt.subplots()
 pass ax to pointplot function
 then ax.legend(handles=ax.lines[::len(df)+1], labels = ['1v/s 2'], loc = 1)

 thus to simplify it, i have added label :
 1) added label parameter to def pointplot (line no : 3333)
 2) adding label to constructor (__init__), set self.label = label in other plot components of _PointPlotter (3338, 1675, 1705)
 3) in _PointPlotter.draw_points(), added label=self.label while plotting plot (1740, 1743)
 4) to display legend if label self.label is not None, call legend. (1816, 1817)

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1672,7 +1672,7 @@ class _PointPlotter(_CategoricalStatPlotter):
     def __init__(self, x, y, hue, data, order, hue_order,
                  estimator, ci, n_boot, units,
                  markers, linestyles, dodge, join, scale,
-                 orient, color, palette, errwidth=None, capsize=None):
+                 orient, color, palette, errwidth=None, capsize=None, label=None):
         """Initialize the plotter."""
         self.establish_variables(x, y, hue, data, orient,
                                  order, hue_order, units)
@@ -1702,6 +1702,7 @@ class _PointPlotter(_CategoricalStatPlotter):
         self.linestyles = linestyles
 
         # Set the other plot components
+        self.label = label
         self.dodge = dodge
         self.join = join
         self.scale = scale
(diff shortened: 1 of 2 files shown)


jeffysonar:
why would it fail???

jeffysonar:
Hey @mwaskom, can you help me with why build failed whenever i tried to commit?? Also, i would like to contribute, how can i start?

seaborn/categorical.py
```diff
@@ -1672,7 +1672,7 @@ class _PointPlotter(_CategoricalStatPlotter):
     def __init__(self, x, y, hue, data, order, hue_order,
                  estimator, ci, n_boot, units,
                  markers, linestyles, dodge, join, scale,
-                 orient, color, palette, errwidth=None, capsize=None):
+                 orient, color, palette, errwidth=None, capsize=None, label=None):
```
mohit-surana: If you look at the logs, you will see that the build fails because this line is too long.

```
seaborn/categorical.py:1675:80: E501 line too long (82 > 79 characters)
make: *** [lint] Error 1
The command "make lint" exited with 2.
```

jeffysonar: Yes, thank you for reply....

PR #1496, opened 2018-07-16
Add catplot companian

mwaskom:
Woof


mmagnuski:
🎉 
BTW - you previously mentioned that dailypuppy was down but I can access the site without problems. Maybe puppyplot shouldn't have been sent to a farm upstate...

mwaskom:
It's not so much that the site went down as that they moved away from a feed of featured pups to a kind of weird aggregation of listicles , making it harder to programmatically grab a "daily puppy" the way that function used to work.

PR #1509, opened 2018-07-20
Reference with :doc: to avoid hiding titles under the navbar

joelostblom:
Clicking the introduction, installation and API links on the front page causes the section heading to be hidden under the navbar. Linking to the page with `:doc:` instead of the specific heading with `:ref:` does not have this problem. The example gallery and tutorial links were working fine with `:ref:` so I didn't touch those.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -72,8 +72,8 @@ attractive and informative statistical graphics.
 For a brief introduction to the ideas behind the library, you can read the
 :doc:`introductory notes <introduction>`. Visit the :doc:`installation page
 <installing>` to see how you can download the package. You can browse the
-:ref:`example gallery <example_gallery>` to see what you can do with seaborn,
-and then check out the :ref:`tutorial <tutorial>` and :doc:`API reference
+:doc:`example gallery <examples/index>` to see what you can do with seaborn,
+and then check out the :doc:`tutorial <tutorial>` and :doc:`API reference
 <api>` to find out how.
 
 To see the code or report a bug, please visit the `github repository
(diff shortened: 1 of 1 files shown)


mwaskom:
> Clicking the introduction, installation and API links on the front page causes the section heading to be hidden under the navbar.

I can't replicate this. Which browser are you using?

joelostblom:
I see it on Firefox 52.6 and 61.0 (addons disabled). Chromium and Vivaldi works fine with `:ref:`.

mwaskom:
Weird! I don't see it on Firefox 60 either. Happy to go with something that seems more robust but I think it would be best to be consistent and use `:doc:` everywhere unless there are costs to doing so.

joelostblom:
Hmm ok, I will try on a few computers in lab tomorrow to see if this only affects my laptop. I believe the only downside with `:doc:` is that it has to reference a top level item, but I don't have that much experience with sphinx so I might be mistaken.

joelostblom:
I have tested this on a couple of more machines (Windows7 Firefox 60 and
OpenSuse Firefox 50 and 60), and I was able to reproduce the issue on both. I'll
include some more info about it below, although since the issue itself is minor
I am not sure how much time is worth spending getting to the bottom of it.

The issue seems to be related to the fixed navbar style, because when I edited
the built HTML files and changed to the default navbar behavior, the headings
were always fully visible. I don't think it is related to your recent CSS fix
for the navbar, because it did not help when I commented that out.

The issue only occurs when a link is referencing a sphinx section tag. If it
references the entire page or the actual full title of the subheading everything
works as expected. To clarify, these two links do not hide the heading

- https://seaborn.pydata.org/introduction.html
- https://seaborn.pydata.org/introduction.html#an-introduction-to-seaborn

But following a link with a sphinx section name goes to a page where the
heading is almost completely hidden under the navbar:

- https://seaborn.pydata.org/introduction.html#introduction

Changing the sphinx section name to the same as the heading (e.g.
`an-introduction-to-seaborn`) still causes the heading to be hidden, so
I can't think of a way to use `:ref:` and get this working.

Using `:doc:` works fine for any heading that is top level, such as all the
links in the introductory paragraphs and the `Tutorials` section. The syntax for linking
with `:doc:` is the same as you have used with the `toctree` for the `Contents`
section, so that might already be using `:doc:` internally.

I don't know what would be a good way of linking the API sections, since those
are subsections on the same page. I guess the HTML subsections could be linked
explicitly.

As for the two links I did not change in my first commit, I believe the tutorial
link works simply because that page is not long enough to scroll, and my guess
for why example page heading was not hidden would be that it somehow relates to
the gallery-like layout of that page.

PR #1531, opened 2018-08-04
Fix deprecation warning typo

benjaminrose:
Fixed two typos.


mwaskom:

PR #1543, opened 2018-08-14
Handle missing values within distplot automatically

### LINKED ISSUE #1542: Handle missing values within distplot
```python
import numpy as np
import seaborn as sns

x = np.random.randn(100)
x[0] = np.nan

sns.distplot(x)
```

Missing values are extremely common when one wants to plot a univariate distribution, but the current behavior of `distplot` is to abort and throw a somewhat cryptic error message when this happens.  I think `distplot` should remove these automatically instead of forcing the user to do so manually before plotting, issuing a simple warning that this has happened.

dsaxton:
As the title suggests, the proposed change is to remove missing values automatically within `distplot` and issue a warning instead of an error. Fixes #1542.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -177,6 +177,14 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
     if a.ndim > 1:
         a = a.squeeze()
 
+    # Remove np.nans, warn user if present
+    nan_mask = np.isnan(a)
+    if nan_mask.sum() > 0:
+        msg = ("a contains missing values, plotting "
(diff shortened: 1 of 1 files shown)


dsaxton:
I'd like to add some tests to ensure that `distplot` returns a warning but not an error with missing data.  Should this go under `test_distributions.py`?

mwaskom:
I don't think a warning is needed here. Most seaborn/matplotlib functions silently ignore missing data ... iirc the error was in the math for computing the default bin size.

Please link to the issue that you created so that it will get closed if this is merged.

PR #1544, opened 2018-08-17
Allow specifying a function to calculate confidence intervals

toddrjen:
This change allows users to specify their own function to calculate the confidence interval.  For example someone could use [`scipy.stats.sem`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sem.html).


mwaskom:
Sorry, but as stated elsewhere, the complexity of the `ci` parameter will not be expanded beyond its current state.

PR #1554, opened 2018-08-31
Added ci="range" option to lineplot

zml:
Added ci="range" option to lineplot in order  to show the max/min values, in alternative to ("sd")


zml:
No problem. Are there alternatives to get  max/min information in this kind of plots?

PR #1557, opened 2018-09-04
Fix error message

niemasd:
The strings are simply concatenated, so without explicit spaces, they print the following:

```
ImportError: Cumulative distributions are currentlyonly implemented in statsmodels.Please install statsmodels.
```


mwaskom:

Issue #1568, opened 2018-09-19
Using regplot() on a figure with multiple axes causes problems


dallascard:
When using `regplot()` on a figure with multiple axes, the x and y limits on the last subplot are partially determined by data plotted in the other axes.

For example, the following code will produce a figure where the range of x values is way too large on the second subplot:

```import matplotlib.pyplot as plt
import seaborn
import numpy as np
import pandas as pd

np.random.seed(32)
y = np.random.randn(20, 1)
x0 = np.random.randn(20, 1) + 100
x1 = np.random.randn(20, 1)

df = pd.DataFrame(np.hstack([y, x0, x1]), columns=['y', 'x0', 'x1'])

fig, axes = plt.subplots(1, 2)
seaborn.regplot(x='x0', y='y', data=df, ax=axes[0])
seaborn.regplot(x='x1', y='y', data=df, ax=axes[1])

plt.show();
```

![regplot](https://user-images.githubusercontent.com/2933570/45769602-f87a9e80-bbf4-11e8-88fb-95a5ca88ea8f.png)

The reason seems to be that the `plot()` function in `_RegressionPlotter` uses `plt.plot()` to create (and then remove) a line, in order to determine what colour to use for the regplot (line 327 of regression.py).

`lines, = plt.plot(self.x.mean(), self.y.mean())`

This has the unintended side effect of updating the limits of one of the plots. Because this is done with `plt` instead of `ax`, this affects the last of the subplots created using `subplots()`, rather than the ax that has been passed to regplot. In the example above, because the first and second columns of X have very different means, the xlim of the second plot is pulled very far to the right.

I believe this can be fixed by changing `plt.plot()` on line 327 of regression.py to `ax.plot()`, and will submit a PR. The proposed change will also cause all subplots to be plotted in the same colour by default (unlike above). However, this will also be more consistent with `lmplot`.

Version info:
python: 3.6.1
seaborn: 0.9.0
matplotlib: 2.0.2

piotrdzwiniel:
I've faced another problem related to the multiple regplots and axes. When one want to draw two regplots on a single axis `truncate=False` works only for the last drawn plot resulting for the first regplot to be limited by its data points. See the example code and the plot below (problem wasn't there in 0.8.1 seaborn version):

```
import pandas
import matplotlib.pyplot as plt
import seaborn

data = {
    'x_a': [-0.0045, -0.0457, 0.2457, -0.1302],
[...]

PR #1569, opened 2018-09-19
fixing regplot to work with multiple subplots

dallascard:
The proposed change in this PR is to replace a call to `plt.plot()` with `ax.plot()` in `_RegressionPlotter` in regression.py. The reason for the change is to solve problems that emerge when `regplot()` is used to plot data on a figure with multiple axes. (See issue #1568) 

A side effect of this change is that multiple calls to `regplot`, each on a different axes, will now be plotted in the same colour by default, as opposed to each using different colours, as is the case now. However, using the same colour will also be more consistent with `lmplot`.

The behaviour of `regplot` when not using multiple subfigures, or not specifying an `ax` is unchanged.

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -324,7 +324,7 @@ class _RegressionPlotter(_LinearPlotter):
 
         # Use the current color cycle state as a default
         if self.color is None:
-            lines, = plt.plot(self.x.mean(), self.y.mean())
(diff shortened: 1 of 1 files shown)


mwaskom:
Great catch. I'd noticed this behavior before but never was able to figure out why it happened.

PR #1575, opened 2018-10-01
[WIP] Add check for categorical hue variable (fixes #1515)

### LINKED ISSUE #1515: relplot error when using str as hue
When I am trying to use a string or categorical column within a dataframe as hue within a relplot we get the following error:
`AttributeError: 'str' object has no attribute 'view'`

Minimal example:
```
test_df = pd.DataFrame(
[{"a": 0.0, "b": 1.0, "c": "1", "d": "1"},
 {"a": 0.0, "b": 1.0, "c": "2", "d": "2"}]
)
g = sns.relplot(x="a", y="b", col="c",  hue="d", data=test_df)
```

benlindsay:
To fix the issue I was seeing on #1515, I added a check that overrides the variable type returned by `_semantic_type()` if the data is a Pandas categorical type. I agree that it makes sense to convert numeric strings into numbers since, as you said in #1515,

>it is not uncommon in pandas (i.e. the data library most commonly used with seaborn) to have numeric data stored with an object datatype, for example if you have integers and missing data.

but I can't think of a case where a series would be a categorical variable if the user didn't explicitly do that. Please let me know if I'm mistaken.

Let me know if you're open to a change like this, and I can look into adding necessary tests for this.

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -357,6 +357,10 @@ class _RelationalPlotter(object):
             if isinstance(palette, (dict, list)):
                 var_type = "categorical"
 
+            # Override if data is pandas categorical series
+            if isinstance(data, pd.Series) and data.dtype.name == "category":
+                var_type = "categorical"
+
         # -- Option 1: categorical color palette
 
         if var_type == "categorical":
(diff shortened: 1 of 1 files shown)


mwaskom:
This change should go in the `semantic_type` function and would need tests.

benlindsay:
Thanks for merging. Sorry I wasn't able to make the time to add tests

Issue #1573, opened 2018-09-26
Customizing color of catplot variables -- can't specify hue


AllanLRH:
Below I'll provide some code to...
1. Give a preview of the data
2. Demonstrate a bar-plot using the desired color-coding
3. Show a desired catplot-figure without the desired color-coding
4. Show how to produce the "error", which I suspect is necessary for some reason
5. Provide the data required to produce the plots

Hope someone will be able to sort this out, and help me and [this other guy at Stackoverflow](https://stackoverflow.com/questions/51772467/specify-color-in-seaborn-catplot).

First a bit of setup:

```python
colorcycle = [(0.220, 0.424, 0.690),
              (0.745, 0.682, 0.831),
              (0.992, 0.753, 0.525),
              (0.498, 0.788, 0.498)]
sns.set_palette(colorcycle)
```

Having data like this (JSON blob at the bottom of this post)

```python
print(repr(df.head()))
# prints the following
Communication channel         B7         B3        G7        G3        S7    ...           C6        S5        S1        C5        C1
Tree                                                                         ...                                                     
0                      44.911739  35.679640  0.976231  1.277292  1.337043    ...     0.254748  0.172246  0.279923  0.171319  0.181755
1                      77.268876   2.934542  2.970511  0.707058  0.798130    ...     0.249481  0.196723  0.159604  0.162971  0.181244
2                      75.404045   4.641294  2.561013  2.451872  0.564517    ...     0.298924  0.180989  0.173417  0.176871  0.152236
3                      52.733662  33.823561  0.718640  1.674308  0.695750    ...     0.220801  0.222329  0.176601  0.183128  0.201160
4                      80.813430   3.792741  1.388992  2.500923  0.651349    ...     0.198659  0.179466  0.188586  0.184450  0.172293

[5 rows x 32 columns]
```
I compute some summary statistics for the data, and use that for a barplot:

```python
cat_dct = {'B': 0.0, 'C': 1.0, 'G': 2.0, 'S': 3.0} 
cat = pd.Series(fi.columns.str[0].map(lambda s: cat_dct[s]), index=fi.columns) 
df_agg = pd.DataFrame([df.mean(), df.std(), cat], index=['Feature importance', 'std', 'cat']).T

fig, ax = plt.subplots(figsize=(8, 12))
bars = sns.barplot(data=df_agg, x='Feature importance', y=df_agg.index, 
[...]

PR #1577, opened 2018-10-03
Add example with dictionary-specified colors for catplot.

AllanLRH:
I've added an example demonstrating how to specify colors of a catplot using a dictionary of – it was something I've struggled with, and previously posted an issue about [#1573](https://github.com/mwaskom/seaborn/issues/1573).

The example use artificial randomly generated data showing wine ratings, and coloring each type of wine in an associated color:

wine_example


mwaskom:
Thanks — but the examples are for showing off different kind of plots, not comprehensive illustrattion of the API (this is covered in the API docs and tutorial).

PR #1579, opened 2018-10-05
Prevent _finalize_grid from breaking legend_out.

lewisamarshall:
Hi!

In 0.9.0, using the FacetGrid.map or FacetGrid.map_dataframe functions breaks the alignment on legend_out=True. This is shown in the plot below, in which the "grouped_barplot.py" example has g.map() called with a dummy function before rendering. 

original_grouped_barplot

This is came from the function _finalize_grid, which was calling tight_layout as a final step. This pull-request removes that line.

new_grouped_barplot

Thanks for such an awesome package! ✨ 

Code to reproduce: 

`
import seaborn as sns

sns.set(style="whitegrid")

titanic = sns.load_dataset("titanic")

g = sns.catplot(x="class", y="survived", hue="sex", data=titanic,
                height=6, kind="bar", palette="muted")
g.despine(left=True)
g.set_ylabels("survival probability")

def f(*args, **kwargs):
    pass

g.map(f)
`

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -845,7 +845,8 @@ class FacetGrid(Grid):
         """Finalize the annotations and layout."""
         self.set_axis_labels(*axlabels)
         self.set_titles()
-        self.fig.tight_layout()
+        # NOTE: Removed figure tightening because it doesn't respect legend_out=True
+        # self.fig.tight_layout()
 
     def facet_axis(self, row_i, col_j):
         """Make the axis identified by these indices active and return it."""


mwaskom:
This can't be the right way to fix this. It might avoid your specific problem, but it will make the library behave completely differently in all other cases.

lewisamarshall:
Yes, definitely a good point. I was trying to understand this, and couldn't come up with what this was trying to address. _finalize_grid is only used in the map functions.

Would it be better to re-apply the legend location calculations after map?

mwaskom:
I don't know, If you look at the examples in the docs, there isn't the problem that you have. I also don't see it on my system. So it is probably some interaction with your specific version of matplotlib, matplotlib backend, etc. The correct way to solve the problem is to start by identifying its specific cause. I can't engage with this any more deeply right now, however.

mwaskom:
I'm going to close this PR but you can re-open as an issue, ideally providing all the information that would be necessary to reproduce.

PR #1585, opened 2018-10-09
 Added More Comments and explanation to imports

MalwareNinja:
Was a great Documentation on seaborn the best i could ask for


mwaskom:
Thanks, but many of these comments are superfluous or simply incorrect.

PR #1588, opened 2018-10-12
Differentiated contours and inner lines and parameterized line color for violinplots.

TheChymera:
It used to be that the contours of violinplots could not be removed without removing inner elements as well (since they would disappear at a `linewidth` of zero). Consequently the `linewidth` parameter was a bit misleadingly summarized as controling the outer lines.

This is now improved with the `linewidth` parameter in fact controlling only the outer lines, and `inner_linewidth` controling the line width for inner parameters.

Additionally, it used to be that dark colours could not be used for the violin plot, as the gray color for the outlines was hard coded. This is addressed by adding the `linecolor` parameter.

These modest improvements enable beautiful figures such as the following (note the white lines, which avoid adding another shade of gray to the spectrum, and the contours being thinner than the inner delimiters --- the can, of course, now also be removed entirely).

article

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -557,7 +557,7 @@ class _ViolinPlotter(_CategoricalPlotter):
     def __init__(self, x, y, hue, data, order, hue_order,
                  bw, cut, scale, scale_hue, gridsize,
                  width, inner, split, dodge, orient, linewidth,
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks — but the previous behavior was intentional.

PR #1589, opened 2018-10-16
Create pair_line_heatmap.py

ydup:
An example for pairing the line plot and heat map


mwaskom:
Thanks — but I don’t think this fits well in the example gallery

PR #1602, opened 2018-11-08
BUGfix for kind 'point' in catplot

stfnrpplngr:
When one of the data points is NaN there is a mismatch between color array and position arrays. 

`ValueError: 'c' argument has 2 elements, which is not acceptable for use with 'x' with size 0, 'y' with size 0.`

For those cases, emptying the color array fixes the error.


mwaskom:
I would need to see a reproducible example of the problem, ideally one that could be turned into at test.

stfnrpplngr:
I added a test. It shows generated data resulting in the `ValueError` described above.

mwaskom:
Please look at how testing is done in seaborn.

PR #1604, opened 2018-11-08
correct lineplot documentation for err_kws

mtd91429:
The variable `err_band` isn't used by lineplot. This seems to be a typo for `err_kws`.


mwaskom:

PR #1607, opened 2018-11-11
GH1571 Fix test_cbar_ticks in test_matrix.py

### LINKED ISSUE #1571: TestHeatmap unit test failure with matplotlib 3.0.0
I am running the unit tests for seaborn under matplotlib 3.0.0 and I am getting a failure.  This failure does not happen in the 2.x series of matplotlib.  I am not sure what the problem is.

```
=================================== FAILURES ===================================
_________________________ TestHeatmap.test_cbar_ticks __________________________

self = <seaborn.tests.test_matrix.TestHeatmap object at 0x7fbc03564828>

    def test_cbar_ticks(self):
        max_n_ticks = 3
    
        locator = mpl.ticker.MaxNLocator(max_n_ticks)
        f, (ax1, ax2) = plt.subplots(2)
        mat.heatmap(self.df_norm, ax=ax1, cbar_ax=ax2,
                    cbar_kws=dict(ticks=locator))
>       nt.assert_equal(len(ax2.yaxis.get_ticklabels()), max_n_ticks)

seaborn/tests/test_matrix.py:397: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib64/python3.6/unittest/case.py:829: in assertEqual
    assertion_func(first, second, msg=msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <nose.tools.trivial.Dummy testMethod=nop>, first = 5, second = 3
msg = '5 != 3'

    def _baseAssertEqual(self, first, second, msg=None):
        """The default assertEqual implementation, not type specific."""
        if not first == second:
            standardMsg = '%s != %s' % _common_shorten_repr(first, second)
            msg = self._formatMessage(msg, standardMsg)
>           raise self.failureException(msg)
E           AssertionError: 5 != 3
```

It doesn't look like seaborn's travis currently tests matplotlib 3.0.0 so I am not sure if this is an error in my build or a general problem.

JustinZhengBC:
Closes #1571 

This test has been causing many failures in PR's. Printing the ticks given in the test gives
```
Text(1, -2.0, '−2')
Text(1, -1.0, '−1')
Text(1, 0.0, '0')
Text(1, 1.0, '1')
Text(1, 2.0, '2')
```
so although in this case the ticker returns the correct number of ticks (-2 and 2 are not visible) calling len on this returns 5 while only 3 are visible, as the figure below (test output) demonstrates.

test_cbar_ticks

Unfortunately, the python2 tests behave as intended, so changing the expected answer would only cause those to start failing instead. This PR replaces the test in question with a similar one that works on both versions, but still tests the intended functionality.

### DIFF
--- a/seaborn/tests/test_matrix.py
+++ b/seaborn/tests/test_matrix.py
@@ -390,7 +390,7 @@ class TestHeatmap(object):
     def test_cbar_ticks(self):
         max_n_ticks = 3
 
-        locator = mpl.ticker.MaxNLocator(max_n_ticks)
+        locator = mpl.ticker.LinearLocator(max_n_ticks)
         f, (ax1, ax2) = plt.subplots(2)
         mat.heatmap(self.df_norm, ax=ax1, cbar_ax=ax2,
                     cbar_kws=dict(ticks=locator))


mwaskom:
Thanks for identifying the issue here — I ended up fixing this a bit differently as part of a broader effort to clean up the build in #1712.

Issue #1609, opened 2018-11-14
Regplot does automatically update axes limits of the wrong plot when no color is provided


AKuederle:
Ok this is an odd one:

If you have a matplotlib figure with multiple subplots and try to plot a regplot inside one of the the axes, regplot will update the axes limits of the last axes of the subplot.

The cause for this is pretty simple. in [this line](https://github.com/mwaskom/seaborn/blob/29d1075cd56878ad2ba71f537df2cd73170ad847/seaborn/regression.py#L327) the regplot function calls `plt.plot` instead of `ax.plot`. Hence the currently active axes (meaning the last subplot) is targeted by this line, automatically updating its limits.

AKuederle:
After some testing: This will change the way the color cycle will be exhausted. Before, the global colorcycle was queried. Now only the colorcycle of the current axes is queried. I am not sure if anything else depends on this functionality

AKuederle:
Ok, I find a solution, which does not require to plot anything, which seems to preserve the original functionality, while still solving this bug

mwaskom:
Thanks for raising; this is a duplicate of #1568

PR #1610, opened 2018-11-14
The "test plot" to determine the color is now performed in the correct axes

AKuederle:
This solves #1609 and prevents the regplot function from updating properties of the wrong axes in case an `ax` object is provided, which is not equal to the currently active axes


AKuederle:
The old approach however changes the behavior of the plot function. If this is the expected behavior, I am happy to change it.

For example, before if you would call regplot for multiple subplots, the scatter would have a different color in each subplot, as seaborn would advance the color rotation of a single axes (the last one). Using just `ax.plot()` in the color selection logic will cycle through the colorcycle of this specfific axis. 
I think this makes way more sense, but I wanted to make sure, that I do not change existing behavior.

But if this is wanted, I will reopen the old one.

Thanks for the input!

mwaskom:
The old behavior was wrong.

AKuederle:
Reopened a pull request using the correct axes instance: #1611

PR #1611, opened 2018-11-14
The "test plot" to determine the color is now performed without actualy plotting anything

AKuederle:
This solves #1609 and prevents the regplot function from updating properties of the wrong axes in case an `ax` object is provided, which is not equal to the currently active axes


mwaskom:
This cannot be accepted because it uses a private method of a matplotlib object. Also it is a duplicate of a correct fix in #1569

Issue #1536, opened 2018-08-08
[discussion] Adding ECDF to seaborn?


ericmjl:
@mwaskom referencing [this tweet](https://twitter.com/story645/status/1027042216206852096) re: ECDFs; I have a simple implementation ready to go which I have stored in textexpander, but I think it might be a useful contribution to `seaborn` users. 

The simplest unit of visualization is a scatterplot, for which an API might be:

```python
def ecdf(df, column, ax=None, step=True):
    #### "if ax" logic goes here" ####
    np.sort(df[column]), np.arange(1, len(df)+1) / len(df)
    if step:
        ax.step(x, y)
    else:
        ax.scatter(x, y)
    return ax
```

With this plotting unit, it can be easily inserted into the pairplot as a replacement for the histogram that occurs on the diagonal (as an option for end-users, of course, not mandatory). I can also see extensions to other kinds of plots, for example, plotting multiple ECDFs on the same axes object.

As I understand it, `distplot` exists, and yes, granted, visualizing histograms is quite idiomatic for many users. That said, I do see some advantages of using ECDFs over histograms, the biggest one being that all data points are plotted, meaning it is impossible to bias the data using bins. I have more details in a [blog post](https://ericmjl.github.io/blog/2018/7/14/ecdfs/), but at a high level, the other biggest advantage I can see is reading off quantiles from the data easily. Also, compared to estimating a KDE, we make no assumptions regarding how the data are distributed (though yes, we can debate whether this is a good or bad thing). 

If you're open to having ECDFs inside seaborn, I'm happy to work on a PR for this. Might need some guidance to see if there's special things I need to look out for in the codebase (admittedly, it'll be my first PR to seaborn). Please let me know; I'm also happy to discuss more here before taking any action.

deniederhut:
I'd like to see this, and would be happy to help implement it.

dsaxton:
I like this idea as well.  Perhaps it would be best to plot it as a simple step function as in R's `ecdf`?  (That is, horizontal lines with circles at the discontinuity points, open from below and closed from above.)

ericmjl:
@dsaxton great idea! I have updated the code sample above such that the step function is the default, while the scatter is the alternative.

ericmjl:
@mwaskom totally understand, no worries! Thanks for taking the time to reply nonetheless :smile:. Also, thanks for the pointers!

mwaskom:
[...]

PR #1621, opened 2018-12-03
ECDF plotting proposal per issue #1536

lmcinnes:
Following some of the discussion from #1536, here is a strawman API for adding ECDF plotting -- essentially it would be an extra option of what you view is a distplot. I've tried to hew as closely to the existing API for such things, but would be happy to discuss. I am sure you are currently very busy, and I have no expectations for this PR; I merely wanted to provide a basic strawman in case you were interested.


mwaskom:
Thanks — as you suspect I don't have time to think through all the relevant issues right now, but this looks very good foundation to build from.

mwaskom:
Thanks for kicking this off a while back.

I've now implemented it in #2141. As the distributions module has been completely revamped in the meantime, it was not possible to pull these commits into the history. But please weight in on #2141 if you have any thoughts on the features or implementation.

Issue #1642, opened 2019-01-08
MatplotlibDeprecationWarning in seaborn.reset_orig()


koki-sato:
## What

`seaborn.reset_orig()` causes `MatplotlibDeprecationWarning`

```python
>>> import seaborn as sns
>>> sns.__version__
'0.9.0'
>>> sns.mpl.__version__
'3.0.2'
>>> sns.reset_orig()
/Users/xxxx/.anyenv/envs/pyenv/versions/3.6.5/lib/python3.6/site-packages/matplotlib/__init__.py:855: MatplotlibDeprecationWarning:
examples.directory is deprecated; in the future, examples will be found relative to the 'datapath' directory.
  "found relative to the 'datapath' directory.".format(key))
/Users/xxxx/.anyenv/envs/pyenv/versions/3.6.5/lib/python3.6/site-packages/matplotlib/__init__.py:846: MatplotlibDeprecationWarning:
The text.latex.unicode rcparam was deprecated in Matplotlib 2.2 and will be removed in 3.1.
  "2.2", name=key, obj_type="rcparam", addendum=addendum)
```

It is because `matplotlib.rcParams.copy()` returns dictionary of rcParams and it includes deprecated rcParams like `examples.directory` and `text.latex.unicode`. However, they are included in the default rcParams of matplotlib.

This is a problem caused by matplotlib, but I want seaborn to catch and discard the warnings internally.

mhooreman:
Same issues here, which is systematic those days by pip installing matplotlib and seaborn with default versions.

PR #1643, opened 2019-01-08
Catch and discard MatplotlibDeprecationWarning in reset_orig()

koki-sato:
## What

Closes https://github.com/mwaskom/seaborn/issues/1642

### DIFF
--- a/seaborn/rcmod.py
+++ b/seaborn/rcmod.py
@@ -2,6 +2,7 @@
 from distutils.version import LooseVersion
 import functools
 import matplotlib as mpl
+import warnings
 from . import palettes, _orig_rc_params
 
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks. I agree that this warning is annoying. But it's ultimately harmless, and `reset_orig` should be a rarely called function. I would worry that catching all deprecation warnings would cause us to miss cases where matplotlib deprecates parameters that *are* relevant to seaborn. Ultimately the costs of doing so outweigh the benefits of reducing a bit of noise, in my view.

koki-sato:
Thank you for your comment! I see, I think so. 

Certainly `seaborn.set()` should not catch deprecation warnings. But `seaborn.reset_orig()` just restores the original setting, so catching warnings in it does not seem to be problematic.

Issue #1648, opened 2019-01-10
Feature request: allow non-symmetrical violint plots


Naereen:
Hi!
I love violinplots, but I always found that it might not be needed to plot them as symmetrical, the information is doubly present for no real reason.
I will send a tiny pull request that adds a key-word argument `nonsymmetrical` to `seaborn.violint`, it can be `False` (the default), or `"left"` or `"right"` (or `"bottom"` or `"top"`) to allow to only plot half of the violin.

Here are two tiny demos, for a single violin or more violins, for horizontal (top/bottom) or vertical (left/right) cases:
![proposal nonsymmetrical violin plots 1](https://user-images.githubusercontent.com/11994719/50970591-1771a900-14e2-11e9-9b49-8ee9346daa1e.png)
![proposal nonsymmetrical violin plots 2](https://user-images.githubusercontent.com/11994719/50970594-19d40300-14e2-11e9-9ca3-fdd4531525cf.png)

Naereen:
It is a very small feature and I think it is useful, when doing violin plots with lots of datasets, it reduces the visual clutter to only plot left/top part of the violin.

mwaskom:
I don't agree that the symmetry is redundant. It facilitates comparison of each violin with its neighbors.

In general I'm opposed to adding too many additional parameters, especially ones that control specific aspects of plot appearance without helping the plots show information any better.

Naereen:
Okay, I understand your point of view.

njvack:
For what it's worth, I'm looking for half-violin plots for [Raincloud plots](https://github.com/RainCloudPlots/RainCloudPlots#python-environment). That package has its own implementation but... it's mainly a copy of your code.

You can jigger seaborn to do half-violins by [adding fake off-scale data to yourset and abusing the `hue` parameter,](https://stackoverflow.com/questions/53872439/half-not-split-violin-plots-in-seaborn) but that's pretty gross.

A more useful alternative, though, might be to add `kind='kde'`to `sns.catplot` — really, a violin plot is just two kdeplots pasted together.

Thoughts?

givenx:
`kind='kde'`to `sns.catplot` would be perfect. We have had multiple requests from our clients to drop the symmetry as it can be off-putting. I think it is more important than you realise, @mwaskom, imho.

PR #1649, opened 2019-01-10
Proposal of nonsymmetrical violin plots (#1648)

Naereen:
Here is my proposal of the feature requested in #1648.
It is naively implemented but works perfectly fine.
I can add a demo/example script, or a test case (if needed, could someone please show me where to put it? thanks)

Here are two tiny demos, for a single violin or more violins, for horizontal (top/bottom) or vertical (left/right) cases:
proposal nonsymmetrical violin plots 1
proposal nonsymmetrical violin plots 2

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -557,7 +557,7 @@ class _ViolinPlotter(_CategoricalPlotter):
     def __init__(self, x, y, hue, data, order, hue_order,
                  bw, cut, scale, scale_hue, gridsize,
                  width, inner, split, dodge, orient, linewidth,
-                 color, palette, saturation):
+                 color, palette, saturation, nonsymmetrical):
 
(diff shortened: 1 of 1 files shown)


mwaskom:
I'm not completely opposed to supporting this use case but I'm not in favor of the approach here. The name of the new keyword argument isn't really clear about what it does, and it should not be necessary to have to align the argument value with whether the plot is horizontal or vertical.

Perhaps an idea to explore in future attempts would be making use of the existing `split` parameter and the existing logic. It might mean that "nonsymettrical" violins are always only on one side, but I think that's fine. And it might mean that "nonsymettrical" violins don't work when there is a ``hue`` semantic, as what is wanted  becomes ambiguous.

Also in general something like this would need to be well tested to get merged.

PR #1671, opened 2019-02-11
Minor example improvement

jeroenboeye:
Simpler code to create mask for upper triangle


mwaskom:

Issue #1653, opened 2019-01-16
Incorrect legend in lineplot when using numeric values in hue


bjkomer:
I believe this is related to #1560 but happens even when `style` is not used and gives incorrect behaviour with `lineplot`.

Here's an example. The legend should have two items, 0.2 and 0.3

```
import matplotlib.pyplot as plt
import seaborn as sns
import pandas as pd

df = pd.DataFrame(
    [
        [0, 0, .2],
        [1, 1, .2],
        [2, 2, .2],
        [0, 0.5, .3],
        [1, 1.5, .3],
        [2, 2.5, .3],
    ],
    columns=['x','y','hue']
)

sns.lineplot(data=df, x='x', y='y', hue='hue')
plt.show()
```
![seaborn_bug_1](https://user-images.githubusercontent.com/1361151/51222511-12f72700-190c-11e9-8c29-c1978740498f.png)

Interestingly it works if the `hue` column contains strings or integers as the categorical value rather than floats. I guess casting that column to a string could be a workaround in the meantime.

bjkomer:
It turns out casting the column to a string doesn't work at all, and produces the following error:

```
Traceback (most recent call last):
  File "seaborn_legend.py", line 55, in <module>
    sns.lineplot(data=df, x='x', y='y', hue='hue')
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/seaborn/relational.py", line 1078, in lineplot
    sort=sort, err_style=err_style, err_kws=err_kws, legend=legend,
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/seaborn/relational.py", line 670, in __init__
    self.parse_hue(plot_data["hue"], palette, hue_order, hue_norm)
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/seaborn/relational.py", line 375, in parse_hue
    data, order, palette, norm
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/seaborn/relational.py", line 260, in numeric_to_palette
    norm(np.asarray(data.dropna()))
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/matplotlib/colors.py", line 942, in __call__
    self.autoscale_None(result)
  File "/home/user/anaconda3/envs/py3torch/lib/python3.6/site-packages/matplotlib/colors.py", line 988, in autoscale_None
    self.vmin = A.min()
[...]

PR #1676, opened 2019-02-13
Fixes issue with string array input to hue and size in relplot

rovyko:
> Treats all non-numeric array inputs to hue and size as categorical. This prevents a string array from being passed to `numeric_to_palette` and causing the error in issue #1653 

When an array (`data`) is passed to the `hue` or `size` argument of  a `_RelationalPlotter` object, it is passed on to the method [`_semantic_type`](https://github.com/mwaskom/seaborn/blob/fe172c44777f87159e9e2b00e919bbd055d6f696/seaborn/relational.py#L523), which determines the array's data type (either `'categorical'` or `'numeric'`). If the array is made up of numeric-like strings (those that can be converted into floats), then the `'numeric'` type is returned, but the original `data` array variable is not converted to a numeric type and is passed to [`numeric_to_palette`](https://github.com/mwaskom/seaborn/blob/fe172c44777f87159e9e2b00e919bbd055d6f696/seaborn/relational.py#L230), which is expecting an array of numeric type.

It is possible to overwrite the `data` variable within the scope of `parse_hue` and `parse_size` with a numeric-converted array returned by `_semantic_type`, but this change would need to be propagated outside of the method to the object for use by other methods.

I think an easier solution is to treat only array types that can be parsed by `numeric_to_palette` as `'numeric'` and the rest as categorical. This change is made by checking the dtype of the passed `data` array against the options in `numpy.sctypes`.

This shouldn't contradict the documentation, which specifies that the array to the `hue` or `size` argument:
>Can be either categorical or numeric, although size mapping will behave differently in latter case.

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -526,7 +526,7 @@ class _RelationalPlotter(object):
             return "categorical"
         else:
             if data.dtype in [bool] + np.sctypes['int'] + np.sctypes['uint'] + \
-                    np.sctypes['float'] + np.sctype['complex']:
+                    np.sctypes['float'] + np.sctypes['complex']:
                 try:
                     float_data = data.astype(np.float)
                     values = np.unique(float_data.dropna())
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -525,13 +525,17 @@ class _RelationalPlotter(object):
         if self.input_format == "wide":
             return "categorical"
         else:
-            try:
-                float_data = data.astype(np.float)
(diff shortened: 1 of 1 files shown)


mwaskom:
Sorry, but the current behavior (treating object typed arrays that can cast to float without exception) is intentional, so I am going to reject this change.

ivirshup:
@mwaskom would you mind elaborating on why this is the case? My intuition was that by providing values as pandas categoricals or as strings they would be treated as categorical.

PR #1683, opened 2019-02-28
[Doc] swarmplot update

agamemnonc:
Update the `swarmplot` documentation to include a description of the `kwargs` argument which is currently missing.


mwaskom:

PR #1708, opened 2019-03-28
Add 'mean' option to 'inner' for violin plots

tom-a-horrocks:
This allows a single line to be drawn at the location of the mean in a violin plot, rather than the median as when inner='quart'. This would make the figure more versatile.

Line dashing currently follows that for the median line when inner='quart'. I have tried it without dashing (i.e., a solid line), however, the end of the solid line sometimes overruns the pdf boundary.


mwaskom:
Thanks — but I think I'm against adding too many other options here. It's easy to show the mean in other ways, e.g. with a `pointplot`.

PR #1711, opened 2019-04-04
removing unnecessary semicolons from docs

guilhermesilveira:
Just a simple cleanup of extra ;s in the documentation


mwaskom:
Thanks but these aren’t unnecessary. They prevent the return value from ending up in the REPL and making things confusing.

guilhermesilveira:
Thanks Michael. Would it make sense to remove the first one when in two sequential lines of code, such as https://github.com/mwaskom/seaborn/pull/1711/commits/fb30f1ea006ed34ab28ce539a4c4798561107a14#diff-9ecd3b20dfaca4c486f27a680847a37eL449
If so, there are only 4 occurrences of such cases.

PR #1726, opened 2019-04-14
Format numeric legend entries in relational plots

### LINKED ISSUE #1653: Incorrect legend in lineplot when using numeric values in hue

### LINKED ISSUE #1703: Too many decimal places in legend labs for relplot with hue or size keywords
I ran into this problem with a particular dataset. Calling `replot` with either `hue` or `size` causes extremely long labels in the legend because of what looks like floating point arithmetic in the labeling logic.

Example:

```python
test = pd.DataFrame({
 'foo' :[1,2,3,4,5,6,7],
 'bar' :[4,6,8,3,5,4,6],
    'baz':[0.4, 0.19, 0.35, 0.22, 0.35, 0.31, 0.09]
}
)
sns.relplot(
data=test.head(7),
x='foo',
y='bar',
size='baz')
```
Selection_006

Is there some way that the legend labels could be rounded or otherwise formatted to prevent this?

MaozGelbart:
Issues #1653 and #1703 describe relational plots legend entries that look affected by some floating point calculations. This is due to non-rounded values returned by matplotlib's ``ticker`` used for obtaining the hue/size legend entries. The proposed PR fixes this behavior by using matplotlib formatters on the obtained tick values to get a nice string representation.
1703-a

Fixes #1653 
Fixes #1703

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -271,7 +271,7 @@ class _RelationalPlotter(object):
         """Return nice legend entries for numeric hue/size variables."""
         levels = ticker.tick_values(*limits).astype(dtype)
         if len(levels) > 1 and levels[1]-levels[0] < 1:
-            sig_digits=int(1-(floor(log10(levels[1]-levels[0]))))
+            sig_digits = int(1-(floor(log10(levels[1]-levels[0]))))
             levels = np.around(levels, decimals=sig_digits)
         return levels
 
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -2,6 +2,7 @@ from __future__ import division
 from itertools import product
 from textwrap import dedent
 from distutils.version import LooseVersion
+from math import floor, log10
 
 import numpy as np
 import pandas as pd
@@ -265,6 +266,14 @@ class _RelationalPlotter(object):
         # palette = {l: cmap(norm([l, 1]))[0] for l in levels}
 
         return levels, palette, cmap, norm
+
+    def ticker_to_legend_entries(self, ticker, limits, dtype):
+        """Return nice legend entries for numeric hue/size variables."""
+        levels = ticker.tick_values(*limits).astype(dtype)
+        if len(levels) > 1 and levels[1]-levels[0] < 1:
+            sig_digits=int(1-(floor(log10(levels[1]-levels[0]))))
+            levels = np.around(levels, decimals=sig_digits)
+        return levels
 
     def color_lookup(self, key):
         """Return the color corresponding to the hue level."""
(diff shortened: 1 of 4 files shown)


Evidlo:
Is there anything blocking this?

mwaskom:
I think I'd prefer if all the entries had the same number of digits. That matches how matplotlib ticks the axes. Actually, how do they avoid floating point weirdness seeping into the axes labels?

MaozGelbart:
Within matplotlib formatting of ticks on axes is done with [ticker.Formatter](https://matplotlib.org/3.1.1/api/ticker_api.html#matplotlib.ticker.Formatter) objects. However these objects are dependent on the axis they control, in the sense that they query the axis for the view interval. One can inject a mock axis with the desired view interval to recover nice formatting. Consider the following example (specific values in the example are taken from #1703 ):
```python
>>> from matplotlib import ticker
>>> 
>>> data_interval=(0.09,0.4)
>>> tck = ticker.MaxNLocator(nbins=3)
>>> tck.tick_values(*data_interval)
['0.0', '0.15000000000000002', '0.30000000000000004', '0.45000000000000007']
>>>
>>> #adding formatter
>>> class dummy_ax:
>>>     def get_view_interval(self):
>>> 	    return data_interval
>>> 
>>> fmt = ticker.ScalarFormatter(useOffset=False)
>>> fmt.axis=dummy_ax()
>>> fmt.format_ticks(tck.tick_values(*data_interval))
['0.00', '0.15', '0.30', '0.45']
```
For sure it's better to use the matplotlib formatting method (with another benefit of auto-fallback to scientific notations on smaller/larger quantities), but the dummy axis injection exploits the intended use. What would be the preferred way? I tend to favor the matplotlib way over implementing a custom tick formatter in seaborn.

mwaskom:
I agree that we don't want to re-implement the formatting code in seaborn. Using a dummy axis aligns with the idea of using a ticker to get values for the numerical legend entries.  It may also either fix, or lay the groundwork for fixing, the problem where numerical legends can have values that are out of the range of the data, which matplotlib would otherwise clip when drawing the axes. So I would be in favor.

MaozGelbart:
I've updated the code and it now uses the matplotlib formatters, so for the given example the number of significant figures is equal for all legend entries. Examples:
1653a

It works also for plots that use LogNorm, this one is taken from the example on [lineplot page](https://seaborn.pydata.org/generated/seaborn.lineplot.html) (slightly differs from the current formatting):
1653b

mwaskom:
Thanks @MaozGelbart

PR #1727, opened 2019-04-15
 Consider binary hue data with one unique value categorical (#1722)

### LINKED ISSUE #1722: Scatterplot with single-valued boolean column for Hue gives faded plot & bad legend
Using a Boolean column for hue in a scatterplot gives a plot with faded points and a legend with the same color for True and False.

image

```
import pandas as pd
import numpy as np
import seaborn as sns

sns.set()

# Create example DataFrame of stock values
df = pd.DataFrame({
    'ticker':np.repeat( ['aapl','goog','yhoo','msft'], 25 ),
    'date':np.tile( pd.date_range('1/1/2011', periods=25, freq='D'), 4 ),
    'price':(np.random.randn(100).cumsum() + 10) })

df['split'] = False
df['split 2'] = "No"


plt.figure()
sns.scatterplot(x=df.date.dt.strftime('%Y-%w'), y='price', data=df, hue='split')
plt.figure()
sns.scatterplot(x=df.date.dt.strftime('%Y-%w'), y='price', data=df, hue='split 2')
```

MaozGelbart:
The current behavior of relational plots considers binary values as categorical for ``hue`` or ``size`` purposes, but only if both 0 (or ``False``) and 1 (or ``True``) values present in the data (see #1722). This PR makes relational plots consider single-valued binary data in the ``hue`` and ``size`` columns of relational plots as categorical.

Fixes #1722

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -530,7 +530,7 @@ class _RelationalPlotter(object):
             try:
                 float_data = data.astype(np.float)
                 values = np.unique(float_data.dropna())
-                if np.array_equal(values, np.array([0., 1.])):
+                if np.all(np.isin(values, np.array([0., 1.]))):
                     return "categorical"
                 return "numeric"
             except (ValueError, TypeError):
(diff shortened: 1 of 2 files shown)


mwaskom:
Feels like it's a little surprising, but I think your argument that it is more consistent with existing practice (and the fact that it will avoid plots where all the points are on the extreme of the numeric colormap) wins the day.

PR #1728, opened 2019-04-18
Update stripplot docs and examples with jitter=True default

MaozGelbart:
PR #1490 made ``jitter=True`` the default for ``stripplot()``. This PR incorporates the change into the docs and examples.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2810,11 +2810,11 @@ stripplot.__doc__ = dedent("""\
     {categorical_data}
     {order_vars}
     jitter : float, ``True``/``1`` is special-cased, optional
-        Amount of jitter (only along the categorical axis) to apply. This
-        can be useful when you have many points and they overlap, so that
-        it is easier to see the distribution. You can specify the amount
-        of jitter (half the width of the uniform random variable support),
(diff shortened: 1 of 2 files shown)


seaborn/categorical.py
```diff
@@ -2810,11 +2810,11 @@ def stripplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
     {categorical_data}
     {order_vars}
     jitter : float, ``True``/``1`` is special-cased, optional
-        Amount of jitter (only along the categorical axis) to apply. This
-        can be useful when you have many points and they overlap, so that
-        it is easier to see the distribution. You can specify the amount
-        of jitter (half the width of the uniform random variable support),
-        or just use ``True`` for a good default.
+        Amount of jitter (only along the categorical axis) to apply. Defaults
```
mwaskom: Doc string parameter information (generally) should not specify the default value. The defaults are apparent in the function signature, and defaults specified in the doc string test are easy to forget to update if the values change, making the documentation wrong.

MaozGelbart: Thanks! Reverted the jitter docstring to its previous description.

PR #1742, opened 2019-05-20
Meaningless change to re-trigger CI with matplotlib 3.1.0

shoyer:
My testing suggests that Seaborn's test suite fails with matplotlib 3.1.0. This will reveal is this is true on CI or merely a quirk of our local setup.


shoyer:
Nevermind, matplotlib 3.1.0 isn't available in conda yet.

Issue #1364, opened 2018-01-25
Weighting the samples of a kdeplot


arvoelke:
Related: 
 - #783 
 - https://stackoverflow.com/questions/31703149/weights-option-for-seaborn-distplot
 - https://stackoverflow.com/questions/30344568/how-to-pass-weights-argument-to-seaborns-jointplot-or-the-underlying-kdeplot

It would be nice to be able to create a `kdeplot` where a weight can be assigned to each sample. Even though there is a `hist_kws` argument, it is not taken into account with the kernel density estimate.

Here's a hypothetical use-case: you have data recording the `(income, donation)` of a number of people to some charity, and now you want to visualize the univariate distribution of `donation` (y-axis) against `income` (x-axis). If you try to do this with a `kdeplot`, currently, the amount donated cannot be included in the density, i.e., each donation is treated uniformly.

My work-around has been to use the weights+observations to generate a million-or-so samples according to that distribution, and then pass those samples to the `kdeplot`. This tends to be slow/noisy, and awkward to implement.

By the way... thank you to the community for the very useful toolkit. I have used it in a recent publication (http://ieeexplore.ieee.org/document/7378880/) and cited seaborn accordingly. Incidentally, I had to use the above work-around of generating proxy samples for that paper in order to have the `kdeplot` weighted correctly.

mwaskom:
The KDE is fit with scipy, which [doesn't seem to support it](https://github.com/scikit-learn/scikit-learn/issues/4394).

Your hypothetical use-case seems like something you want to be doing with a lowess regression, not a weighted KDE, anyway.

cstich:
It looks like scipy will with the 1.2.0 release support weights for KDE. The relevant pull request is [here](https://github.com/scipy/scipy/pull/8991). Any chance of opening this issue again?

MaxGhenis:
Agree I think this should be reopened, given it's being worked on in #1747.

mwaskom:
Closed in #2104 with the addition of `weights` to `kdeplot` as part of a very thorough rewrite.

PR #1747, opened 2019-05-23
Add the ability to weight samples when using kdeplot or joint plot

hockyg:
For jointplot, additional arguments must be given like:
    jointplot(...,joint_kws={'weights':weights}, marginal_kws={'weights':weights},...)
For kdeplot, it is simply an extra argument:
    kdeplot(...,weights=weights,...)

This pertains to/resolves issue  #1364


MaxGhenis:
Hi - I'd love to see this feature if you're able to complete the PR! Looks like it's mostly just missing tests?

mwaskom:
Given that this is version-dependent and adds some complexity w/r/t whether scipy or statsmodels is used (does statsmodels really not support weights?) I think this should be an explicit parameter in seaborn with docs that are clear about the implications, avoiding the current roundabout-logic and error messages.

PR #1755, opened 2019-06-05
handled bug in regress_out function when using {x,y}_partial argument…

divyanshusrivastava:
… in lmplot

Motivated by this post  -> https://stackoverflow.com/questions/56419714/series-object-has-no-attribute-reshape-when-using-x-partial-argument-in-l


mwaskom:
While you'd think that calling `.values` is the proper way to convert a pandas object to a numpy object, it's not guaranteed to work, but `np.asarray()` should. Also it would be best if this could get a test (one that fails on the status quo but passes with this change).

Issue #1552, opened 2018-08-30
Add a parameter to lineplot which would specify the treatment of Nones


SteveSmirnoff:
I've tried to visualize my data with Seaborn line plot. My series had lots of `Nans ` and a generated line plot was quite misleading. 

![image](https://user-images.githubusercontent.com/40000494/44867601-f6c25880-ac90-11e8-9ccd-ceedda0c4fed.png)

using code below I managed to solve this issue:
```
fig, ax = plt.subplots(figsize=(10, 5))

plot = sns.lineplot(
    ax=ax
    , data=df, x="Date", y="Data"
    , hue=df["Data"].isna().cumsum()
    , palette=["blue"]*sum(df["Data"].isna())
    , legend=False, markers=True
)

ax.set_xticklabels([])
plt.show()
```

![image](https://user-images.githubusercontent.com/40000494/44867731-5b7db300-ac91-11e8-9559-62ff68573436.png)

However - it would be very nice to add a parameter to line plot which would make it easy to specify the desired way to treat Nones.

Thank you for your attention.

mwaskom:
I tried to solve this in #2273 but hit some snags and may revisit it outside of a narrow bugfix context.

Notes for future thought ...

In the simplest case of using `lineplot` like `plt.plot`, this is relatively straightforward, but the more general case is tricky. First, we do want to exclude NA when aggregating, because the presence of some NAs should not erase datapoints that do have valid data. So the challenge is identifying situations where we want to preserve a "missing" datapoint. The second challenge is that the default behavior is to sort by `x`, but doing that will lose track of the NAs as they exist in the original data.

Gabriel-Kissin:
Firstly - much appreciation to @mwaskom for the awesome seaborn!

A quick workaround which seems to work, exploiting the `units` parameter of `sns.lineplot`:

```
def lineplot_breaknans(data, break_at_nan=True, break_at_inf=True, **kwargs):
    '''sns.lineplot by default doesn't break the line at nans or infs, 
        which can lead to misleading plots.
    See https://github.com/mwaskom/seaborn/issues/1552 
        and https://stackoverflow.com/questions/52098537/avoid-plotting-missing-values-on-a-line-plot
    
    This function rectifies this, and allows the user to specify 
        if it should break the line at nans, infs, or at both (default).
    
    Note: using this function means you can't use the `units` argument of sns.lineplot.'''

    cum_num_nans_infs = np.zeros(len(data))
[...]

PR #1790, opened 2019-08-05
Remove nan/inf-containing data points from relational plots data

### LINKED ISSUE #1761: relplot uses rows with missing data (which are not plotted) to determine scale for marker sizes
I have a dataset (exaggerated for effect) where one row has a large value in a particular column (foo) and missing data in another:

```python
test = pd.DataFrame(
{
    'foo' : [0, 7, 1, 0, 6, 8, 2, 6, 9, 100],
    'bar' : [8, 7, 7, 6, 2, 2, 4, 0, 2, np.NaN],
    'baz' : [1, 9, 6, 3, 6, 8, 9, 8, 7, 4],
})
test

   foo  bar  baz
0    0  8.0    1
1    7  7.0    9
2    1  7.0    6
3    0  6.0    3
4    6  2.0    6
5    8  2.0    8
6    2  4.0    9
7    6  0.0    8
8    9  2.0    7
9  100  NaN    4
```

I want to investigate the relationships between foo, bar and baz with `relplot` using `size`. When I use my `bar` as X (or Y) everything is fine; because one of the values is missing it isn't plotted so the large value doesn't affect the axis:

```python
sns.relplot(
    data=test, 
    x='foo', 
    y='bar',
    size='baz',
    sizes=(2,200)
)
```
Selection_011

But, when I map my `bar` column to `size`, it looks like the logic for scaling the points includes the last row - even though it won't be plotted - which makes the scale too big for the remaining points which are rendered way too small:

```python
sns.relplot(
    data=test, 
    x='baz', 
    y='bar',
    size='foo',
    sizes=(2,200)
)
```

Selection_012

The same thing happens when mapping `foo` to the `hue` variable:

Selection_013

The largest `foo` that will actually be plotted is 9, but the colour scale has room for 100, so the majority of the scale isn't used.

The workaround is to `dropna()` before plotting, but it seems for consistency that if points with missing data don't affect the axes limits, they shouldn't affect the size or hue limits. I'm quite fond of the fact that `relplot` deals sensibly with missing data, I think this is a good design choice for rapid data exploration, and I often have datasets with patterns of sparse missing data which cause this problem.

MaozGelbart:
Issue #1761 describes a behavior where relational plots continuous numeric semantics (``hue``/``size``) are calculated based on all available data points for the semantic, even if for points that are not plotted (i.e. nan- or inf-containing points). This PR removes such points from participation in relational plots, so the ``hue`` and ``size`` ranges will be calculated based on valid data points only. 

Fixes #1761

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -167,8 +167,12 @@ class _RelationalPlotter(object):
 
         # ---- Post-processing
 
+        # Make sure rows with nan-containing semantics are not used #GH1761
+        plot_data.dropna(axis=1, how='all', inplace=True)
+        plot_data.dropna(axis=0, how='any', inplace=True)
+
         # Assign default values for missing attribute variables
-        for attr in ["hue", "style", "size", "units"]:
+        for attr in ["x", "y", "hue", "style", "size", "units"]:
             if attr not in plot_data:
                 plot_data[attr] = None
 
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -167,9 +167,12 @@ class _RelationalPlotter(object):
 
         # ---- Post-processing
 
-        # Make sure rows with nan-containing semantics are not used #GH1761
-        plot_data.dropna(axis=1, how='all', inplace=True)
-        plot_data.dropna(axis=0, how='any', inplace=True)
+        # Make sure nan/inf-containing semantics are not used #GH1761
+        plot_data = (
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks — I think I'm going to reject this one because it will make it impossible to enable the behavior requested in https://github.com/mwaskom/seaborn/issues/1552, which I think should happen. #1761 needs some thinking but can probably be solved closer to where the actual issue occurs.

PR #1796, opened 2019-08-09
Fix documentation on distplot bins

tonyyzy:
Distplot bins documentation is misleading. If Freedman-Diaconis rule gives a value greater than 50, 50 will be used as the bin size.


MarcoGorelli:
In case it's not clear, you need to remove the trailing whitespace from line 60 :)

(see https://travis-ci.org/mwaskom/seaborn/jobs/569911602 :+1: 
> seaborn/distributions.py:60:78: W291 trailing whitespace
)

tonyyzy:
Ah, sorry, didn't see the notification that testing failed

PR #1809, opened 2019-08-20
Avoid mapping color codes for non-string palettes in sns.set()

### LINKED ISSUE #1546: Using set with color_palette() output fails
When creating a color palette using the color_palette() function, an error is thrown in 0.9.0 when trying to pass the output to sns.set(). MWE:

```
import seaborn as sns

pal = sns.color_palette("husl")
sns.set(palette=pal)
```

This appears to be for all named palettes

MaozGelbart:
Running ``sns.set(palette=pal)`` with ``pal`` created through ``color_palette()`` fails without specifying ``color_codes=False``(#1546 ). The proposed PR uses ``color_codes=False`` in case of a non-string palette input.

Fixes #1546

### DIFF
--- a/seaborn/rcmod.py
+++ b/seaborn/rcmod.py
@@ -4,6 +4,7 @@ import functools
 import matplotlib as mpl
 import warnings
 from . import palettes, _orig_rc_params
+from .external.six import string_types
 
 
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks — I ended up moving the error handling a little further down but the thorough test was very helpful.

PR #1819, opened 2019-08-29
added "fixed" option for the scale parameter for violinplot

iriet:
For anyone who wants to make multiple violin plots that share the same scaling parameter, but want the appearance to follow scale="counts" mode, I have added a "fixed" option that allows for overriding the maximum counts determination that is made in "counts" mode, since this latter value cannot be transferred to other plots.


mwaskom:
Hi — thanks, but I think this adds too much additional complexity.

Issue #655, opened 2015-07-27
Feat Req: Accessing statistical values computed for plots


jat255:
Sometimes it is desirable to get the values that are plotted for reference (using seaborn as a calculation package, as well as plotting). 

Currently, it appears (please correct me if I'm just oblivious to it) that there is no easy way to access the values that are calculated for the plots (such as the mean, confidence intervals, etc.). In order to access these, I had to explicitly create my own `_BarPlotter` (for instance) to get at it's parameters.

This is something that other people have expressed interest in as well (see [this](https://stackoverflow.com/questions/26192974/confidence-interval-information-in-figure-legend) Stack Overflow question and [my answer](http://stackoverflow.com/a/31643869/1435788)). Is this functionality available, or are there any plans to make it accessible?

mwaskom:
It is not available, and it will not be made available.

You should use statsmodels if you want to do statistical testing.

jat255:
Right. Thanks for the consideration. 

Is it that unreasonable to output some information that is already calculated (could be a switch that defaults to 'off')? I don't think it adds much complexity, and these are values that people are interested in, clearly. It's not really statistical testing, so much as getting a quantitative value out of your plot...

olgabot:
When I need to do something like that, I end up writing the code myself. Here's some examples for a `count`-style `FacetGrid` and a heatmap annotated with values other than the values provided, e.g. here I wanted to plot percentages but annotate with absolute values. The code for the plot below is [here](http://nbviewer.ipython.org/gist/olgabot/1a8423256419188f6238)

![image](https://cloud.githubusercontent.com/assets/806256/8918557/92d7fb16-346d-11e5-9fcb-1327db4a96b2.png)

mwaskom:
It would be a large change to the function of the package, and it is out of scope because seaborn is a library for visualization, not for statistics (statsmodels) or data munging (pandas).

qbilius:
I wanted to provide additional motivation why it is important to provide access to the statistics that were computed while generating plots in seaborn. These figures often end up in scientific reports, where it is important to be precise about the analysis that was performed. Without an access to the details how a particular statistic (e.g., a regression line) was computed, it becomes more difficult to establish that the statistic shown in the figure is exactly the same as the one computed externaly via pandas / statsmodels. In other words, to use seaborn-generated figures in my reports, I need to rerun statistical analyses myself, visualize them with matplotlib and make sure they match seaborn's output exactly.

[...]

PR #1837, opened 2019-09-18
ENH: Store references to regression outputs in plotter._data and to plotter in ax.__dict__['_plotter']

westurner:
I notice that #655 suggests that it would be infeasible to store `yhat`, `yhat_boots`, `grid`, and `err_bands` and hereby propose a simple underscore-prefixed `plotter._data` dict and also an `ax.__dict__['plotter']` in order that the parameters of these excellent charts are accessible to curious chart-implementers.

Note the commented `self._data['func']`. Also, I've run the test suite locally but haven't written any tests for this pending feedback regarding its merit.

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -125,6 +125,9 @@ class _RegressionPlotter(_LinearPlotter):
         # Save the range of the x variable for the grid later
         self.x_range = self.x.min(), self.x.max()
 
+        # Cache regression data in _data
+        self._data = dict()
+
     @property
     def scatter_data(self):
         """Data where each observation is a point."""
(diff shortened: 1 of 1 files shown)


mwaskom:
I think I’m going to pass here, thanks.

westurner:
Do you think it's somewhat unresponsible to publish charts without any of the underlying parameters as calculated by the apparently-unique for-charting-only analysis routines?

~"Just use statsmodels" [ https://github.com/mwaskom/seaborn/issues/655#issuecomment-125073496 ]?

What's the point of visualization-only? Why do we all need to use statsmodels differently from how seaborn generates these then-somewhat-irrelevant charts? Why do we need to do visual regression testing when we could just compare the parameters?

The responsible, reproducible thing to do would be to provide an option to also publish the calculated parameters.

Pretty pictures, tho.

mwaskom:
No, but feel free not to use the software.

PR #1877, opened 2019-11-24
Fix default color codes.

jrnold:
The default colors in `set_color_codes()` (used for `palette = "reset"`) uses the wrong color code for m (magenta) and instead duplicates yellow. Fixing this only requires changing the RGB numbers for the array entry associated with magenta.


mwaskom:
Thanks — good catch.

PR #1892, opened 2019-12-14
included parametric CIs and updated help

mmbannert:
Dear MIchael,

Thanks for creating seaborn - it's been of tremendous help in my daily work.

However, I have not found an easy way to create confidence intervals parametrically based on the standard error of the mean (sem): "ci" let's you specify "sd" in order to plot standard deviations or perhaps "68" to plot something >>like<< sem via one (of numerous possible) bootstrap methods.

There has been an increasing interest in quantifying measurement error in the form of confidence intervals (CI) over the last few years ("The New Statistics: Why and How", Cumming, 2014) and it is probably advisable to be consistent across studies about the way confidence intervals are constructed (or - if there is variability across fields - to have more flexibility).

Since most research surely uses parametric confidence intervals, I have added the possibility to (1) easily plot sem CIs by setting the "ci" parameter to "sem" (similar to the "sd" option) and (2) to calculate CIs parametricelly (and deterministically) for any desired level of estimation error simply from the sems without having to run any bootstraps. In order to use this functionality, the n_boot parameter simply has to be None. This results in confidence intervals that in my experience are just as "good" as the previous ones while preserving consistency across studies, which commonly rely on parametric CIs as well. 

I have updated the function help accordingly.

Cheers,
Michael

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2134,12 +2134,11 @@ _categorical_docs = dict(
     stat_api_params=dedent("""\
     estimator : callable that maps vector -> scalar, optional
         Statistical function to estimate within each categorical bin.
-    ci : float or "sd" or "sem" or None, optional
+    ci : float or "sd" or None, optional
         Size of confidence intervals to draw around estimated values.  If
-        "sd" or "sem", skip bootstrapping and draw the standard deviation 
-        or standard error of the mean of the observations, respectively.
-        If ``None``, no bootstrapping will be performed, and error bars will
-        not be drawn.
+        "sd" , skip bootstrapping and draw the standard deviation of the 
+        observations. If ``None``, no bootstrapping will be performed, and
+        error bars will not be drawn.
     n_boot : int, optional
         Number of bootstrap iterations to use when computing confidence
         intervals. If ``None``, confidence intervals will be calculated
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -16,6 +16,7 @@ from .external.six.moves import range
 from . import utils
 from .utils import iqr, categorical_order, remove_na
 from .algorithms import bootstrap
+from .algorithms import stats as sbstats
 from .palettes import color_palette, husl_palette, light_palette, dark_palette
 from .axisgrid import FacetGrid, _facet_docs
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Hi, thanks for the PR but I’m not going to incorporate it.  If you need to be very specific about what kind of error bars you show, I would recommend using marplotlib directly.

mwaskom:
Partially reconsidering .

I remain resistant to the idea of allowing `ci="sem"` (if you search, you will find this issue has been discussed), which doesn't exactly make sense semantically. Yes there is `ci="sd"`, which is even worse and I regret adding because it muddied the API, but at least that is a necessary hack in the sense that it allowed something genuinely new. In contrast, I don't really understand the argument in favor of showing the SEM rather than confidence intervals.

That said, setting `n_boot=None` to trigger analytic CIs is kind of an interesting idea. It's sensible, and also non-obvious enough that most people will continue to use what I would consider the superior approach of bootstrapping.

But this PR would need a lot of work and I don't have the time to provide guidance. Any change to how confidence intervals are computed would need to be cohesive throughout the library. And that isn't going to be easy because their computation is not fully generic, so some refactoring would have to  be done (one of the reasons I have remained resistant thus far). It would need to be very carefully tested. And thought would need to be put into what happens when, for instance,  the user asks for an analytic confidence interval but uses an estimator that is not the mean (remember, this can happen).

mmbannert:
Hi Michael,

I agree that the "sem" option is not strictly necessary as one could easily get the same results by setting ci=68 (more or less) in the case of analytic CIs. I just thought I'd put it in there because I saw that "sd" seemed to be an accepted option for ci.
 
True, I only took into account the possibility that the user would be interested in using the np.mean estimator. So now the code checks if "np.mean" is indeed the estimator that the user is asking for.

I think that parametric CIs are still useful because parametric statistics are widely used and easy to communicate in articles. Bootstrapping is at least as fine but it requires an extra paragraph about how exactly data have been bootstrapped due to the abundance of possible bootstrapping techniques.

I understand if you're too busy to provide more guidance. Sorry, I can't be more helpful. 

Best,
mb

mwaskom:
Closed with #2407

PR #1895, opened 2019-12-15
add the ability to change the figure dpi.

orena1:
The default is 100:
https://matplotlib.org/3.1.1/users/dflt_style_changes.html
so it would not change the behavior. This is important for jupyter-notebook and jupyterlab


mwaskom:
This is something that can be handled through the matplotlib rc params, or on the returned figure object. I don’t think it needs to be a parameter of the grid object.

PR #1896, opened 2019-12-17
Prevent multiple regplots from adding expanding x margins to the plot

### LINKED ISSUE #1814: Range of axes become too large when using hue and kind="reg" together in a pairplot.
Hello!  When I specify kind="reg" and a hue, the ranges of the axes in every plot in my pairplot become incredibly large.  For example, instead of ranging from 0 to 100 on the y-axis and 0 to 30 on the x-axis, the axes are now -100 to 200 and -100 to 100, respectively.  Before/after images included.

before

after

Thanks!

mwaskom:
Fixes #1814

### DIFF
--- a/doc/releases/v0.9.1.txt
+++ b/doc/releases/v0.9.1.txt
@@ -4,23 +4,31 @@ v0.9.1 (Unreleased)
 
 This is a minor release, comprising mostly bug fixes and compatibility with dependency updates.
 
+New features
+~~~~~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


mwaskom:
This works with multiple plots on a single axes but doesn't work across multiple axes. I believe that is a bug in matplotlib (https://github.com/matplotlib/matplotlib/issues/15967). For lack of a clean solution to work around the mpl issue, I'm going instead to opt for making truncation of the regression line the default behavior, which has other arguments in its favor.

Issue #1908, opened 2019-12-30
Inconsistent distplot() behavior between cumulative=True and cumulative=False


TheTravellingSalesman:
I noticed what I assume to be a bug in the behavior of `seaborn.distplot()`. 

(Where `sparse_data` is a dataset containing null values):

If you try to plot the PDF of data which contains null values using `seaborn.distplot(sparse_data, hist=False, kde_kws=dict(cumulative=False))`, the resulting plot will display the PDF of the data, excluding the null values.

If you try to plot the CDF of the same data using `seaborn.distplot(sparse_data, hist=False, kde_kws=dict(cumulative=True))`, the resulting plot will display nothing at all, although you'd expect it to plot the CDF of the data, excluding the null values.

Here is a minimal reproducible example:

```
import seaborn as sns
import pandas as pd
df = pd.DataFrame({'feature_1': [1,2,3,None,None,None,4,5,6]})

# this will plot the PDF of the data, ignoring null values
sns.distplot(df['feature_1'], hist=False, kde_kws=dict(cumulative=False))

# this will display an empty plot
sns.distplot(df['feature_1'], hist=False, kde_kws=dict(cumulative=True))
```

The attached .zip file contains the Jupyter-Lab Notebook I used to run this example, and an environment.yml file which can be used to re-create the conda environment I ran this example in.

[seaborn_distplot_bug.zip](https://github.com/mwaskom/seaborn/files/4010540/seaborn_distplot_bug.zip)

mwaskom:
Trying the evaluate the kernel density and CDF using statsmodels directly 

```python
import statsmodels.nonparametric.kde as smnp
kde = smnp.KDEUnivariate(df["feature_1"])
kde.fit()
print(kde.density)
print(kde.cdf)
```

Returns a valid array for the density and an array of nans for the cdf.

There's also a warning printed when the CDF is evaluated:

> IntegrationWarning: The occurrence of roundoff error is detected, which prevents the requested tolerance from being achieved.  The error may be  underestimated.

So the inconsistency, such as it is, arises from statsmodels...

TheTravellingSalesman:
Ah, I see. 

So in summary, the `statsmodels` subdependency of `seaborn` is running into a roundoff error when attempting to fit the KDE for the CDF of the data, which results in an array of NaNs, which `seaborn` plots, which results in a blank plot. 

Thank you for clearing that up. I'll dive into `statsmodels` to find the root of this inconsistent behavior. :-)

[...]

PR #1910, opened 2019-12-30
Remove NaN values from input array before passing it into statsmodels…

TheTravellingSalesman:
... for KDE fitting. Avoids roundoff error in CDF fitting.

This pull request modifies 
```
seaborn.distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
             hist_kws=None, kde_kws=None, rug_kws=None, fit_kws=None,
             color=None, vertical=False, norm_hist=False, axlabel=None,
             label=None, ax=None)
```
such that if there are any NaN values the input array, `a`,  then those values are removed prior to `a` being passed to `statsmodels` for KDE fitting. This avoids roundoff error in the CDF KDE fitting, which results in `statsmodels` returning an array of NaNs, and a blank CDF plot. 

This is the proposed solution to issue #1908

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -178,6 +178,11 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
     if a.ndim > 1:
         a = a.squeeze()
 
+    # Drop null values from array
+    mask = np.isnan(a)
+    if mask.any():
+        a = a[~mask]
+
     # Decide if the hist is normed
     norm_hist = norm_hist or kde or (fit is not None)
 


TheTravellingSalesman:
Hi @mwaskom,

I'm looking into a way I might be able to test my changes as part of it's own unit test, or perhaps as part of an existing unit test, but I'm not sure there's a cohesive way to do so. I'm also noticing that the effect on the total code coverage is minuscule, so I'm not sure it's even worth integrating into testing. 

If you have any thoughts or advice, please let me know.

mwaskom:
Thanks @TheTravellingSalesman; see #1913 which built on this with tests  and a change within `kdeplot`.

TheTravellingSalesman:
Absolutely, happy to help! 
Thank you for the quick responses, and I'll look forward to the next release!

PR #1920, opened 2020-01-10
Correct typo

habi:
At least I think that `a.namel` should be `a.name`...


mwaskom:
A "namel" is a name that is also a label :)

PR #1923, opened 2020-01-12
DOC: Fix broken/misleading links and add catplot cross-refs

MaozGelbart:
* corrected a few links under catplot examples in boxplot and pointplot
* corrected residplot "see also" link to jointplot
* added some cross-refs to catplot ("see also") 
* added kwargs description for stripplot, as this is directly shown in the code examples below. Notably, several categorical plots accept kwargs in their signature but it is not described below them, perhaps because they don't use the kwargs input - this may needs to be looked at (examples: boxenplot, violinplot, pointplot).

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2219,7 +2219,7 @@ _categorical_docs = dict(
                 glyphs.\
     """),
     catplot=dedent("""\
-    catplot : Combine a categorical plot with a class:`FacetGrid`.\
+    catplot : Combine a categorical plot with a :class:`FacetGrid`.\
     """),
     boxenplot=dedent("""\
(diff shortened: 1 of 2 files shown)


mwaskom:
I have no idea what's up with that test failure ... it looks like `conda` is broken?

MaozGelbart:
@mwaskom the difference from previous successful builds is that ``conda update -q`` on travis CI now results in conda 4.8.0, as opposed to 4.6.14 previously. See [conda/conda#9583](https://github.com/conda/conda/issues/9583).

mwaskom:
Yep. But is seabon doing something weird here? I would have assumed that conda fundamentally breaking would precipitate a large number of bug reports, whereas that issue was posted 4 days ago and has no response.

PR #1925, opened 2020-01-14
Use a persistant internal random number generator for boostrapping

### LINKED ISSUE #1924: Nondeterministic regplot due to unset bootstrap random_seed
I've had some random non-determinism in my regplot error bands (was particularly annoying as I was keeping plots under git to make sure they are reproducible).

I've tracked it down to [this line](https://github.com/mwaskom/seaborn/blob/065d3c1ed58b029d61913cca8fc315cc985ffc91/seaborn/regression.py#L235), when it calls [bootstrap](https://github.com/mwaskom/seaborn/blob/065d3c1ed58b029d61913cca8fc315cc985ffc91/seaborn/algorithms.py#L32). Bootstrap falls back to numpy random (reads from /dev/urandom, I believe), so currently there is no way to make plots deterministic.

I see two possible solutions here:

1. propagate `random_seed` argument all the way up to `regplot` function. 

2. set `random_seed` argument `bootstrap`, e.g. to 0. If someone genuinely wants nondeterminism, they can pass None explicitly.
   
    Are there any downsides? I've grepped usage in seaborn and it seems that all usages of `bootstrap` would suffer from similar nondeterminism, so that would fix it once and for all.

Happy to open a PR once we figure out what can be done. Thanks!

mwaskom:
Fixes #1924 

This provides part of a stop-gap solution for deterministic bootstrapping.

As written, it affords control over the random state for those who are willing to use a hack by assigning a `RandomState` object to `seaborn.algorithms._rng`.

The open question is whether expose a proper mechanism for setting the seed in the public API. I think this is kind of a hacky approach either way, so I'm not wild about doing so. There's not much precedent that I am aware of for libraries having their own private but internally-global random state. Punting would allow me to decide later on to add a parameter to every function that does bootstrapping such that users could pass a `RandomState` there, which is "best practices".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
   - hash -r
   - conda config --set always_yes yes --set changeps1 no 
   - conda config --add channels conda-forge
-  - conda update -q conda
+  # - conda update -q conda  TODO Re-enable on resolution to conda/conda#9583
   - conda config --set channel_priority false
   - conda info -a
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no 
   - conda config --add channels conda-forge
   # - conda update -q conda  TODO Re-enable on resolution to conda/conda#9583
+  - conda install conda=4.6.14  # TODO remove after re-enabling above line
   - conda config --set channel_priority false
   - conda info -a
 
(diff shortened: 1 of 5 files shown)


MaozGelbart:
The python2.7 failures are due to [https://github.com/conda-forge/setuptools-feedstock/issues/126](https://github.com/conda-forge/setuptools-feedstock/issues/126) and should be resolved soon if I understand correctly.

mwaskom:
That issue is now closed, but re-triggered python27 builds continue to fail...

mwaskom:
Closing in favor of #1926

PR #1926, opened 2020-01-14
Expose RNG seed where bootstrapping is used

### LINKED ISSUE #1924: Nondeterministic regplot due to unset bootstrap random_seed

mwaskom:
Alternative to #1925 that exposes a seed argument on a per-function basis and has no global state.

Closes #1924

### DIFF
--- a/doc/releases/v0.9.1.txt
+++ b/doc/releases/v0.9.1.txt
@@ -19,6 +19,8 @@ New features
 
 - Added the ability to pass hierarchical label names to the :class:`FacetGrid` legend, which also fixes a bug in :func:`relplot` when the same label appeared in diffent semantics.
 
+- Added the ability to seed the random number generator for the bootstrap used to define error bars in several plots. Relevant functions now have a ``seed`` parameter, which can take either fixed seed (typically an ``int``) or a numpy random number generator object (either the newer ``numpy.random.Generator`` or the older ``numpy.random.RandomState``).
+
 Bug fixes and adaptations
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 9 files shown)


mwaskom:
@karlicoss  I think this covers everything but if you want to do some testing, that would be helpful.

karlicoss:
@mwaskom wow, that was really fast, sorry I didn't get to test it before your merged.
I just checked against master branch it and it works! Thanks for your investigation and fix :heart: 

I also had to add `truncate=False` (due to [this change](https://github.com/mwaskom/seaborn/pull/1896), I believe), and now scaling is even slicker and plots look better. Thanks again :)

PR #1929, opened 2020-01-15
MNT: Fix Travis-CI tests with latest conda

MaozGelbart:
The proposed PR allows for conda4.8 to install and run, including tests on python3.5. Builds for python3 are also faster by ~15% by downloading the proper miniconda. The only test mode that requires conda4.6.14 is python2.7 with pinned dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,10 @@ before_install:
   - conda config --add channels conda-forge
   # Pinned deps aren't happy when conda>4.6.14
   - if [[ "$DEPS" == "PINNED" ]]; then
-      conda update -q conda=4.6.14
+      conda update -q conda=4.6.14;
     else
-      conda update -q conda
-  fi
+      conda update -q conda;
+    fi
   - conda config --set channel_priority false
   # Workaround py2.7/conda4.8 tqdm absence 
   - if [[ "$PYTHON" == "2.7" ]]; then pip install tqdm; fi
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks @MaozGelbart. #1926 has a couple of fixes that get travis green. My plan is to wrap up a couple remaining issues, make an 0.9.1 release, and then do a 0.10.0 release that drops support for Python 2 (and 3.5). Hopefully that will make these annoying issues go away.

MaozGelbart:
Thanks @mwaskom . The proposed PR allows for conda4.8 to install and run, including tests on python3.5. Builds for python3 are also faster by ~15% by downloading the proper miniconda. The only test mode that requires conda4.6.14 is python2.7 with pinned dependencies. I don't know how to pick the setuptools commit from your branch so I ended up copying it.

Thanks for sharing your plans regarding upcoming releases. I also anticipate maintaining python 2.7 in seaborn will only become harder as dependent packages drop support.

mwaskom:
Thanks, this ended up a better fix than what I had in #1926.

MaozGelbart:
Thanks for merging. It seems that [https://github.com/mwaskom/seaborn/pull/1929/commits/4bf5fd0a29cb0f38367c9e0864b14419dc53e611](https://github.com/mwaskom/seaborn/pull/1929/commits/4bf5fd0a29cb0f38367c9e0864b14419dc53e611) may be reverted as conda-forge setuptools45 for py27 is marked "broken" in the conda-forge channel - but this needs to be verified.

PR #1941, opened 2020-01-22
Drop Python 2 support

mwaskom:
- Removes support for Python 2
- Removes official support for Python 3.5
- Increases minimal versions of dependencies to versions from circa 2017.


.travis.yml
```diff
@@ -21,25 +18,13 @@ env:
...
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes --set changeps1 no 
   - conda config --add channels conda-forge
-  # Pinned deps aren't happy when conda>4.6.14
-  - if [[ "$DEPS" == "pinned" ]]; then
-      conda install -q conda=4.6.14;
-    else
-      conda update -q conda;
```
MaozGelbart: this line is required to update conda to the latest version. it used to fail on pinned and py27 but should be ok with py27 removal

PR #1944, opened 2020-01-23
Make dogplot more tolerant

janpipek:
I really like this easter egg :-)

However, when you suggest someone to use `dogplot` instead of `catplot`, they will receive a `TypeError` explaining that they provided too many arguments. This kind of spoils the fun. What about making it more tolerant?


mwaskom:
Woof.

Issue #1309, opened 2017-10-22
Heatmap with custom center doesn't preserve vmin/vmax


dlmackin:
When I set a non-default value for center on a heatmap I expected the ends of the range to stay mapped to the same colours, 

For example, here's some data with a range from -10 to 30:

```
uniform_data = np.random.randint(low=-10, high=30, size=(10, 12))
sns.heatmap(uniform_data, cmap="PuOr", annot=True)
```
![image](https://user-images.githubusercontent.com/1126755/31864539-81701a72-b72c-11e7-9282-baa98b7ff491.png)

I would like to be able to set the center to 0 and have values -10 and 30 remain mapped to the ends of the colour map, like this:

![image](https://user-images.githubusercontent.com/1126755/31864541-8974f238-b72c-11e7-9909-d3411c55dfc5.png)

Instead, I get colours mapped to a range of -30 to 30:

```
sns.heatmap(uniform_data, cmap="PuOr", annot=True, center=0)
# Adding explicit vmin/vmax doesn't help:
# sns.heatmap(uniform_data, cmap="PuOr", annot=True, center=0, vmin=-10, vmax=30)
```
![image](https://user-images.githubusercontent.com/1126755/31864543-8d7502d8-b72c-11e7-9466-d5505b79b39c.png)

This seems to be because the normalization step in  [_determine_cmap_params](https://github.com/mwaskom/seaborn/blob/0beede57152ce80ce1d4ef5d0c0f1cb61d118375/seaborn/matrix.py#L227) can only do a linear interpolation between 2 points. I made the second example using a custom colour map normalizer, like this:

```
class MidpointNormalize(matplotlib.colors.Normalize):
    def __init__(self, vmin=None, vmax=None, midpoint=None, clip=False):
        self.midpoint = midpoint
        matplotlib.colors.Normalize.__init__(self, vmin, vmax, clip)

    def __call__(self, value, clip=None):
        x, y = [self.vmin, self.midpoint, self.vmax], [0, 0.5, 1]
        return np.ma.masked_array(np.interp(value, x, y))

sns.heatmap(uniform_data, cmap="PuOr", annot=True, norm=MidpointNormalize(midpoint=0))
```

If this behaviour makes sense generally and not just for me, I can make a pull request updating the normalization code in [_determine_cmap_params](https://github.com/mwaskom/seaborn/blob/0beede57152ce80ce1d4ef5d0c0f1cb61d118375/seaborn/matrix.py#L227) to do the same.

[...]

PR #1948, opened 2020-01-27
Copy props from old cmap when creating new cmap in heatmap

### LINKED ISSUE #1729: Heatmap with center, cmap and mask does not respect cmap.set_bad color
I want to display some data with a custom center and a custom mask color. However, `center`, `cmap` and `mask` together drop some cmap properties that one can set with `cmap.set_bad` for bad/missing data.

It seems like `_HeatMapper` creates another cmap from the colors of the passed cmap when center is specified: https://github.com/mwaskom/seaborn/blob/b9551aff1e2b020542a5fb610fec468b69b87c6e/seaborn/matrix.py#L230

Demo:
```python
import matplotlib as mpl
import numpy as np
import seaborn as sns


np.random.seed(42)
sns.set(rc={"figure.figsize": (12, 9)})
fig, (ax1, ax2) = mpl.pyplot.subplots(nrows=1, ncols=2)
cmap = mpl.cm.get_cmap("BrBG")
cmap.set_bad("black")

corr = np.corrcoef(np.random.randn(10, 200))
mask = np.zeros_like(corr)
mask[np.triu_indices_from(mask)] = True

heatmap_kwargs = dict(mask=mask, vmin=-0.3, vmax=0.3, square=True, cmap=cmap)
sns.heatmap(corr, ax=ax1, **heatmap_kwargs)
sns.heatmap(corr, ax=ax2, center=0.0, **heatmap_kwargs)

fig.savefig("mwe.png")
```

mwe

MaozGelbart:
When ``heatmap`` is fed with both ``cmap`` and ``center`` arguments, it creates a new colormap but the props of the input cmap are not copied (colors for under/over/bad values). The proposed PR copies the props of the input cmap to the new cmap internally used within the heatmapper. 

Fixes #1729

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -216,20 +216,29 @@ class _HeatMapper(object):
 
         # Recenter a divergent colormap
         if center is not None:
-            # Copy under/over/bad values
+
+            # Copy bad values
             # in mpl<3.2 only masked values are honored with "bad" color spec
             # (see https://github.com/matplotlib/matplotlib/pull/14257)
             bad = self.cmap(np.ma.masked_invalid([np.nan]))[0]
+
+            # under/over values are set for sure when cmap extremes
+            # do not map to the same color as +-inf
             under = self.cmap(-np.inf)
             over = self.cmap(np.inf)
+            under_set = under != self.cmap(0)
+            over_set = over != self.cmap(self.cmap.N - 1)
+
             vrange = max(vmax - center, center - vmin)
             normlize = mpl.colors.Normalize(center - vrange, center + vrange)
             cmin, cmax = normlize([vmin, vmax])
             cc = np.linspace(cmin, cmax, 256)
             self.cmap = mpl.colors.ListedColormap(self.cmap(cc))
             self.cmap.set_bad(bad)
-            self.cmap.set_under(under)
-            self.cmap.set_over(over)
+            if under_set:
+                self.cmap.set_under(under)
+            if over_set:
+                self.cmap.set_over(over)
 
     def _annotate_heatmap(self, ax, mesh):
         """Add textual labels with the value in each cell."""
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -216,11 +216,20 @@ class _HeatMapper(object):
 
         # Recenter a divergent colormap
         if center is not None:
+            # Copy under/over/bad values
+            # in mpl<3.2 only masked values are honored with "bad" color spec
+            # (see https://github.com/matplotlib/matplotlib/pull/14257)
+            bad = self.cmap(np.ma.masked_invalid([np.nan]))[0]
+            under = self.cmap(-np.inf)
+            over = self.cmap(np.inf)
             vrange = max(vmax - center, center - vmin)
             normlize = mpl.colors.Normalize(center - vrange, center + vrange)
             cmin, cmax = normlize([vmin, vmax])
             cc = np.linspace(cmin, cmax, 256)
             self.cmap = mpl.colors.ListedColormap(self.cmap(cc))
+            self.cmap.set_bad(bad)
+            self.cmap.set_under(under)
+            self.cmap.set_over(over)
 
     def _annotate_heatmap(self, ax, mesh):
         """Add textual labels with the value in each cell."""
--- a/seaborn/tests/test_matrix.py
+++ b/seaborn/tests/test_matrix.py
@@ -1,5 +1,6 @@
 import itertools
 import tempfile
+import copy
 
 import numpy as np
 import matplotlib as mpl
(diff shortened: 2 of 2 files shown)


mwaskom:
Hmm I'm not sure this produces behavior we'd generally want.

For example:

```python
x = np.random.uniform(-.5, 1, (10, 10))
sns.heatmap(x, center=0, vmin=-.1, cmap="bwr")
```
image

The problem is the presence of the bright blue cells, which previously would have been a very pale blue.

The problem is that, if a matplotlib colormap doesn't have an explicit `_rgba_over` or `_rgba_under` set, it defaults to using the first and last color in the spectrum. So the approach here is assigning the original "lowest" color to the new colormap.

I guess the best way to check whether the colormap intends to have over/under colorts that are different from just the ends of the spectrum is to compare `cmap(-inf)` versus `cmap(0)` for the original colormap and only do `set_under` on the new object if they differ.

Does that make sense? Am I missing anything?

MaozGelbart:
I'm quite new to this code so I may got this wrong - please correct me if that's the case. What you describe looks like a different issue that relates to the ``center`` part (probably with/without ``vmin/vmax``) but not to ``set_under/over/bad``. Consider this code, which uses ``pcolormesh`` calls with the current seaborn centering logic:
```python
import matplotlib
from matplotlib.colors import ListedColormap, Normalize, DivergingNorm
import numpy as np
import copy

# make these smaller to increase the resolution
dx, dy = 0.05, 0.05

# generate 2 2d grids for the x & y bounds
y, x = np.mgrid[slice(1, 5 + dy, dy),
                slice(1, 5 + dx, dx)]

z = np.sin(x)**10 + np.cos(10 + y*x) * np.cos(x)

# x and y are bounds, so z should be the value *inside* those bounds.
# Therefore, remove the last value from the z array.
z = z[:-1, :-1]

fig, axs = plt.subplots(figsize=(10,8), nrows=2, ncols=2)
(ax0, ax1 ,ax2, ax3) = (axs.flat[0], axs.flat[1], axs.flat[2], axs.flat[3])

#no vmin or center
cmap = copy.copy(plt.get_cmap('PiYG'))
im = ax0.pcolormesh(x, y, z, cmap=cmap)
fig.colorbar(im, ax=ax0)
ax0.set_title('normal')

# vmin only
cmap = copy.copy(plt.get_cmap('PiYG'))
vmin=.6
im = ax1.pcolormesh(x, y, z, cmap=cmap, vmin=vmin)
fig.colorbar(im, ax=ax1)
ax1.set_title('vmin='+str(vmin))

# vmin+center (copied from _heatmapper)
(vmin, vmax) = (.6, 1)
cmap = copy.copy(plt.get_cmap('PiYG'))
center=.65
vrange = max(vmax - center, center - vmin)
normlize = Normalize(center - vrange, center + vrange)
cmin, cmax = normlize([vmin, vmax])
cc = np.linspace(cmin, cmax, 256)
cmap = ListedColormap(cmap(cc))
im = ax2.pcolormesh(x, y, z, cmap=cmap, vmin=vmin)
fig.colorbar(im, ax=ax2)
ax2.set_title('current code, vmin=' + str(vmin) + ', center='+str(center) )

# vmin+center, using correct normalization
(vmin, vmax) = (.6, 1)
cmap = copy.copy(plt.get_cmap('PiYG'))
center=.65
norm = DivergingNorm(vmin=vmin, vmax=vmax, vcenter=center)
im = ax3.pcolormesh(x, y, z, cmap=cmap, norm=norm, vmin=vmin)
fig.colorbar(im, ax=ax3)
ax3.set_title('fix, vmin=' + str(vmin) + ', center='+str(center) )
```


If I understand correctly, the bottom left facet is the problematic one, and it uses the seaborn logic to get the new erroneously centered cmap. Non-equal centering a cmap is now very easy with ``DivergingNorm`` but it was only introduced in matplotlib=3.1 (and was renamed in matplotlib=3.2 to TwoSlopeNorm). I think that a fix for such an issue (if it is indeed an issue) should be introduced in a different PR.

mwaskom:
No, the different behavior from `DivergingNorm`/`TwoSlopeNorm` is intentional and working correctly.

My comment was about how your changes here alter the appearance of cells with values below/above vmin/vmax. Previously, the behavior was correct when the supplied colormap does not have user-set under/over values and wrong when it does. Now, the behavior is wrong when the supplied colormap has user-set under/over values and correct when it does.

MaozGelbart:
Thanks! I misunderstood the behavior of ``center`` when it's not symmetrical with regards to vmin/vmax. This behavior is also emphasized in [https://github.com/mwaskom/seaborn/issues/1309#issuecomment-338498308](https://github.com/mwaskom/seaborn/issues/1309#issuecomment-338498308) but not so much in the heatmap docstring. I now see how this PR changes things when under/over were not set on the input cmap. So following your comment above, seems like testing for ``input_cmap(min(data))==input_cmap(-inf)`` is enough to infer that under color was not set?

mwaskom:
> So following your comment above, seems like testing for input_cmap(min(data))==input_cmap(-inf) is enough to infer that under color was not set?

No, because `matplotlib.colors.Colormap.__call__` expects a value normalized in [0., 1.]. (Note floats — `int` values are treated as an index by `ListedColormap` which many seaborn functions return.

MaozGelbart:
Thanks for the guidance. So unless under/over were explicitly set to the defaults, the code should now handle cmap under/over values properly.

PR #1951, opened 2020-01-31
Use the correct platform integer type

toddrme2178:
`np.array.take` requires that the index be the platform integer type (`int32` for 32bit and `int64` for 64bit), but `np.random.RandomState.randint` returns an `int64` on any platform by default.  This makes sure it returns the right type of integer for the current platform.


mwaskom:
I assume this should be closed in favor of #1952?

PR #1952, opened 2020-01-31
Use the correct platform integer type

toddrme2178:
`np.array.take` requires that the index be the platform integer type (`int32` for 32bit and `int64` for 64bit), but `np.random.RandomState.randint` returns an `int64` on any platform by default.  This makes sure it returns the right type of integer for the current platform.

Fixes mwaskom#1950

### DIFF
--- a/seaborn/algorithms.py
+++ b/seaborn/algorithms.py
@@ -81,7 +81,7 @@ def bootstrap(*args, **kwargs):
 
     boot_dist = []
     for i in range(int(n_boot)):
-        resampler = integers(0, n, n)
+        resampler = integers(0, n, n, dtype=np.int_)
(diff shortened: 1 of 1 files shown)


mwaskom:
Can you link to some documentation on this? It's a much leakier abstraction than one usually sees in numpy...

toddrme2178:
The relevant numpy issue is numpy/numpy#11696. 

> Arguably this is by design. If you're indexing with a type bigger than np.intp, then it's possible to overflow, and you don't want that.
>
> From that, it follows that you should really just use np.intp to store indices. Internally, this may save a bunch of copies in numpy anyway

mwaskom:
Thanks @toddrme2178

PR #1958, opened 2020-02-11
FIX: avoid squeezing singleton vectors in regression plotter

### LINKED ISSUE #1949: `lmplot` fails if any group has only one record
For example:

```
import numpy as np
import pandas as pd
import seaborn as sns

df = pd.DataFrame({
    'col': np.append(np.random.choice(np.array(['a', 'b', 'c']), 10), ['d']),
    'x': np.random.normal(size = 11),
    'y': np.random.normal(size = 11),
})

sns.lmplot(x = 'x', y = 'y', col = 'col', data = df)
```

It raises `IndexError: invalid index to scalar variable` which is not very intuitive. I think this case can occur with any data coming from the real world, and would be much nicer to exclude this group automatically or simply plot the single dot without regression line.

MaozGelbart:
As reported in #1949 , ``regplot`` fails when its array-like input is single-valued. This may be pandas series, python list or numpy array (although different errors may appear according to the input type). For example, the following code fails on master with recent ``numpy`` and ``pandas`` packages:
```python
import seaborn as sns
sns.regplot([1.5], [2])
```
with ``IndexError: too many indices for array`` raised in ``_RegressionPlotter.dropna()``. This is because the input vector is reduced by ``np.squeeze`` and singletons lose that dimension and no longer support select operations. 
The proposed fix avoids squeezing singletons, so "array-like" properties are retained. This means that the single point will be plotted and (if ``fit_reg==True``) a regression will be fitted (although, no line will be plotted).

Fixes #1949

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -48,7 +48,7 @@ class _LinearPlotter(object):
                 vector = np.asarray(val)
             else:
                 vector = val
-            if vector is not None:
+            if vector is not None and vector.shape != (1,):
                 vector = np.squeeze(vector)
             if np.ndim(vector) > 1:
                 err = "regplot inputs must be 1d"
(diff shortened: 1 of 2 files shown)


MaozGelbart:
Oh, a line will be plotted if ``truncate==False``. Somewhat misleading. Perhaps would be best not to fit and draw a line if ``x.min()==x.max()``?

mwaskom:
Incorporated and merged into #1969

PR #1961, opened 2020-02-14
DOC: update truncate parameter string to reflect new default

qwhelan:
The docstring for `truncate` is out of date given the new default is `True`

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -536,9 +536,8 @@ _regression_docs = dict(
     """),
     truncate=dedent("""\
     truncate : bool, optional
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks  — this is a good example of why parameter descriptions shouldn't mention default values. Too easy to get out of sync. In that spirit, I'd prefer if it changed to

> If ``True``, the regression line is bounded by the data limits. If ``False``,  it extends to the x axis limits.

qwhelan:
Updated with your proposed changes

mwaskom:
Thanks @qwhelan

Issue #1668, opened 2019-02-07
[Feature request] [boxplot] Option to have boxplot widths according to the number of observations


RalfG:
Hi,

What I would really like as a feature in Seaborn's boxplot function, is the option to set the boxplot's widths according to the size of the dataset visualized in each box. This can be done pretty easily in matplotlib.pyplot with the boxplot `widths` argument:

```python
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns

# Create five datasets with different numbers of sample sizes
sample_sizes = [5, 84, 95, 45, 10]
data = [np.random.ranf(s) for s in sample_sizes]

# Define widths for boxplots as 0.8 of the normalized sample size
widths = [0.8 * s / max(sample_sizes) for s in sample_sizes]

plt.boxplot(data, widths=widths)
plt.show()
```
![seaborn_feature_request](https://user-images.githubusercontent.com/26458971/52412355-cd7ae380-2ade-11e9-8a17-2c939c3a5061.png)
However, then I'm loosing the awesome `data=..., x=…, y=…, hue=…` functionality and the nice styling that Seaborn provides:
```python
# Function to make flat list out of list of lists
flatten = lambda l: [item for sublist in l for item in sublist]

# Construct DataFrame for seaborn boxplot
data_df = pd.DataFrame([flatten([[s] * s for s in sample_sizes]), flatten(data)]).T
data_df.rename(columns={0: 'sample_size', 1: 'value'}, inplace=True)

sns.boxplot(data=data_df, x='sample_size', y='value', order=sample_sizes)
plt.show()
```
![seaborn_feature_request_2](https://user-images.githubusercontent.com/26458971/52412484-2f3b4d80-2adf-11e9-81ed-dc6fdd479b10.png)

I have tried to pass `widths` as a keyword argument in `sns.boxplot`, but then an error is returned saying the argument `widths` received multiple values. So it would be really cool to be able to pass the `widths` argument to pyplot, or even cooler if there was an argument in sns.boxplot that automatically sets the widths according to the respective sizes if it is set to True.

Thanks a lot!

datachico:
I arrived here looking for the same exact thing.  Would be super useful!

buhtz:
Because this Issue and the linked PR (#1962) are pending I would like to ask if I can access the drawn boxes as a workaround.

Something like this:

```
[...]

PR #1962, opened 2020-02-14
Add support for per-category width on boxplot.

elmer-garduno:
Support per-category width on boxplot. As discused on https://github.com/mwaskom/seaborn/issues/1668, this is done by passing an array of widths on the sns.boxplot(width) argument.

Please let me know if this makes sense. I can add further testing.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1,4 +1,4 @@
-from __future__ import division
+/from __future__ import division
 from textwrap import dedent
 import colorsys
 import numpy as np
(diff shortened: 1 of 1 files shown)


mwaskom:
I think that if this were going to be supported, it would have to be similar to how that option works in violinplot.

elmer-garduno:
Thanks, you mean in terms of preserving the area? Or in terms of
automatically using the number of samples to determine the relative width?

elmer-garduno:
I see, do you mean using the scale parameter instead and preserve the
semantics of weight as they are?

If so that makes sense, let me try it.

mwaskom:
This is now obviated (without explicit effort to enable it) by #3294:

```python
sns.boxplot(data=tips, x="day", y="total_bill", widths=[.2, .4, .6, .8])
```


It doesn't work with `hue`; enabling that would be rather difficult and I lean not worth it.

Thanks for your interest!

PR #1965, opened 2020-02-18
MNT: Remove unused code paths

### LINKED ISSUE #1942: Remove code conditional on unsupported dependencies
#1941 updates the minimal version of the dependencies and obviates several instances in the main codebase where there are multiple paths for older versions of libraries (especially matplotlib).

I may not include this in #1941, but this code should eventually get cleaned up.

MaozGelbart:
With the update of minimal supported versions, some code paths may be removed. This PR removes such code paths for unsupported ``pandas``, ``matplotlib`` and ``scipy`` versions. It also modifies some tests to ensure ``pandas==1.0.0`` compatibility. 

Fixes #1942

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -8,7 +8,6 @@ import matplotlib.pyplot as plt
 import matplotlib.transforms as tx
 from matplotlib.collections import LineCollection
 import warnings
-from distutils.version import LooseVersion
 
 try:
     import statsmodels.nonparametric.api as smnp
@@ -216,10 +215,7 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
         if bins is None:
             bins = min(_freedman_diaconis_bins(a), 50)
         hist_kws.setdefault("alpha", 0.4)
-        if LooseVersion(mpl.__version__) < LooseVersion("2.2"):
-            hist_kws.setdefault("normed", norm_hist)
-        else:
-            hist_kws.setdefault("density", norm_hist)
+        hist_kws.setdefault("density", norm_hist)
 
         orientation = "horizontal" if vertical else "vertical"
         hist_color = hist_kws.pop("color", color)
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -1,5 +1,4 @@
 """Plotting functions for visualizing distributions."""
-from __future__ import division
 import numpy as np
 from scipy import stats
 import pandas as pd
--- a/seaborn/rcmod.py
+++ b/seaborn/rcmod.py
@@ -1,13 +1,8 @@
 """Control plot style and scaling using the matplotlib rcParams interface."""
-from distutils.version import LooseVersion
 import functools
 import matplotlib as mpl
 import warnings
 from . import palettes, _orig_rc_params
-
-
-mpl_ge_150 = LooseVersion(mpl.__version__) >= '1.5.0'
-mpl_ge_2 = LooseVersion(mpl.__version__) >= '2.0'
 
 
 __all__ = ["set", "reset_defaults", "reset_orig",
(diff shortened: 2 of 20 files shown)


mwaskom:
Thanks @MaozGelbart, do you think you've made all the changes you intended to make here?

MaozGelbart:
I think I'm done, thanks!

MaozGelbart:
Sorry, just added removal of ``__future__`` imports as well.

seaborn/distributions.py
```diff
@@ -216,10 +214,7 @@ def distplot(a, bins=None, hist=True, kde=True, rug=False, fit=None,
         if bins is None:
             bins = min(_freedman_diaconis_bins(a), 50)
         hist_kws.setdefault("alpha", 0.4)
-        if LooseVersion(mpl.__version__) < LooseVersion("2.2"):
```
mwaskom: The [minimal matplotlib version](https://github.com/mwaskom/seaborn/blob/master/setup.py#L32) is 2.1.2 so this path needs to stick around a little while longer.

seaborn/rcmod.py
```diff
@@ -44,23 +39,21 @@
 
     ]
 
-if mpl_ge_2:
+_style_keys.extend([
```
mwaskom: With the removal of the conditional, this code can be simplified further by just defining one list.

seaborn/rcmod.py
```diff
@@ -540,12 +533,9 @@ def set_palette(palette, n_colors=None, desat=None, color_codes=False):
 
     """
     colors = palettes.color_palette(palette, n_colors, desat)
-    if mpl_ge_150:
-        from cycler import cycler
-        cyl = cycler('color', colors)
-        mpl.rcParams['axes.prop_cycle'] = cyl
-    else:
-        mpl.rcParams["axes.color_cycle"] = list(colors)
+    from cycler import cycler
```
mwaskom: Since this doesn't need to be under a conditional, pleas move it to the top of the module.

seaborn/tests/test_utils.py
```diff
@@ -368,59 +364,61 @@ def test_locator_to_legend_entries():
...
-            # yield check_load_dataset, name
-            check_load_cached_dataset(name)
+def check_load_dataset(name):
+    ds = load_dataset(name, cache=False)
+    assert(isinstance(ds, pd.DataFrame))
+
+
+def check_load_cached_dataset(name):
+    # Test the cacheing using a temporary file.
+    # With Python 3.2+, we could use the tempfile.TemporaryDirectory()
```
mwaskom: Implementing this would fit well with these changes.

seaborn/utils.py
```diff
@@ -568,18 +549,7 @@ def get_view_interval(self):
 
 def get_color_cycle():
     """Return the list of colors in the current matplotlib color cycle."""
-    try:
-        cyl = mpl.rcParams['axes.prop_cycle']
-        try:
-            # matplotlib 1.5 verifies that axes.prop_cycle *is* a cycler
```
mwaskom: Can you double check that this issue is addressed in later matplotlib versions? It is written ambiguously and I'm not sure off the top of my head whether matplotlib enforces the presence of `color` in the `axes.prop_cycle`.

seaborn/distributions.py
MaozGelbart: ["density" kw exists in mpl==2.1.2](https://matplotlib.org/2.1.2/api/_as_gen/matplotlib.pyplot.hist.html) as well (and has the same meaning), so this works the same across all supported versions.

seaborn/utils.py
MaozGelbart: This is never validated and is actually as bug on master since the fallback rcparam ``axes.color_cycle`` has been deprecated a while ago. So in this sense this PR doesn't change current code state. I'll open a separate issue for this bug.

seaborn/rcmod.py
MaozGelbart: Thanks, updated

seaborn/tests/test_utils.py
MaozGelbart: Added this as well, thanks!

seaborn/rcmod.py
MaozGelbart: Simplified, thanks!

seaborn/utils.py
mwaskom: OK good, I agree that's best handled independently.

seaborn/distributions.py
mwaskom: OK great, they were both around for a while.

mwaskom:
Thanks!

PR #1969, opened 2020-02-22
Don't fail in regplot on singletons, don't fit regression either

mwaskom:
Follow up to #1958 

cc @MaozGelbart

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -536,9 +536,8 @@ _regression_docs = dict(
     """),
     truncate=dedent("""\
     truncate : bool, optional
-        By default, the regression line is drawn to fill the x axis limits
-        after the scatterplot is drawn. If ``truncate`` is ``True``, it will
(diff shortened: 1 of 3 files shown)


seaborn/regression.py
```diff
@@ -127,8 +127,13 @@ def __init__(self, x, y, data=None, x_estimator=None, x_bins=None,
         else:
             self.x_discrete = self.x
 
+        # Disable regression in case of singleton inputs
+        if len(self.x) <= 1:
+            self.fit_reg = False
```
MaozGelbart: Perhaps warn here if fit_reg was initially set to True (current default)?

mwaskom: I considered that but I am generally not a fan of warnings being raised as a consequence of default arguments. In most cases, I would not think people would expect to draw a regression line for singleton inputs. (I think the main situation you'll encounter this code path is in a faceted context, where the warning would be somewhat confusing.)

PR #1979, opened 2020-03-01
MNT: simplify to_utf8

MaozGelbart:
Within python3 all strings are unicode strings, so this function can be simplified with the removal of python2 support. Note that this PR also removes unicode-specific grid labels tests, because they test that unicode strings are supported and now all strings are unicode strings. See [here](https://docs.python.org/3/library/stdtypes.html#str) for information on how str(obj) is handled in python3.


MaozGelbart:
Reopened as an attempt to trigger CI.

MaozGelbart:
CI passed but is not shown for some reason. CI run logs can be found [here](https://travis-ci.org/mwaskom/seaborn/builds/656929497?utm_medium=notification&utm_source=github_status).

PR #1985, opened 2020-03-09
Make all links possible use https://

hartwork:
Related to #1984

PS: For colorbrewer2 https is not yet available, so I raised awareness upstream at https://github.com/axismaps/colorbrewer/issues/28 .

Issue #1984, opened 2020-03-09
Project description link points to http://seaborn.pydata.org/ without TLS


hartwork:
Hi!

While browsing the seaborn docs I noticed the cross-out lock in the address bar...

![seaborn_lack_ssl](https://user-images.githubusercontent.com/1577132/76224728-13dbce00-621c-11ea-843b-28ad2410af22.png)

I find that https://seaborn.pydata.org/ with TLS works fine so I'd suggest to add an "s" to the link in the main project description:

![seaborn_proj_desc](https://user-images.githubusercontent.com/1577132/76224871-1f2ef980-621c-11ea-8466-b271e776b82d.png)

I'm happy to provide a PR for any more https fixes needed in Git once this issue is fixed.

Thanks and best, Sebastian

mwaskom:
Thanks for catching.

PR #1988, opened 2020-03-11
Make remaining links use https:// (issue #1984, follow-up to #1985)

hartwork:
As a result of https://github.com/axismaps/colorbrewer/issues/28 URL https://colorbrewer2.org/ is working now :tada: so here's one last update regarding https. Best! :beers: 

Related to:
- issue #1984
- pull request #1985


mwaskom:
Thanks

Issue #1994, opened 2020-03-15
ValueError if set saturation or light to 100 in diverging_palette


scratchmex:
In the docs of [seaborn.diverging_palette](https://seaborn.pydata.org/generated/seaborn.diverging_palette.html) says you can set the light or saturation in [0,100]. Doing so throws an exception from matplotlib.

Code to reproduce:
```python
import seaborn as sns

sns.diverging_palette(0, 1, l=100) # ValueError: RGBA values should be within 0-1 range
sns.diverging_palette(0, 1, s=100) # ValueError: RGBA values should be within 0-1 range
```

<details>
<summary>Both drop the exact same exception:</summary>

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-a2fdf7756e52> in <module>
----> 1 sns.diverging_palette(0, 1, l=100)

~\Anaconda3\lib\site-packages\seaborn\palettes.py in diverging_palette(h_neg, h_pos, s, l, sep, n, center, as_cmap)
    744     palfunc = dark_palette if center == "dark" else light_palette
    745     n_half = int(128 - (sep // 2))
--> 746     neg = palfunc((h_neg, s, l), n_half, reverse=True, input="husl")
    747     pos = palfunc((h_pos, s, l), n_half, input="husl")
    748     midpoint = dict(light=[(.95, .95, .95)], dark=[(.133, .133, .133)])[center]

~\Anaconda3\lib\site-packages\seaborn\palettes.py in light_palette(color, n_colors, reverse, as_cmap, input)
    636     """
    637     color = _color_to_rgb(color, input)
--> 638     light = set_hls_values(color, l=.95)  # noqa
    639     colors = [color, light] if reverse else [light, color]
    640     return blend_palette(colors, n_colors, as_cmap)

~\Anaconda3\lib\site-packages\seaborn\utils.py in set_hls_values(color, h, l, s)
    169     """
    170     # Get rgb tuple representation
--> 171     rgb = mplcol.colorConverter.to_rgb(color)
    172     vals = list(colorsys.rgb_to_hls(*rgb))
    173     for i, val in enumerate([h, l, s]):

~\Anaconda3\lib\site-packages\matplotlib\colors.py in to_rgb(c)
    298 def to_rgb(c):
    299     """Convert *c* to an RGB color, silently dropping the alpha channel."""
--> 300     return to_rgba(c)[:3]
    301 
    302 

[...]

PR #1995, opened 2020-03-16
FIX: migration from husl to hsluv

### LINKED ISSUE #1361: husl color system now called HSLuv
It looks like the husl color system has been renamed to HSLuv.

Here's the new link - http://www.hsluv.org/

This old link from the color palette's docs is dead - http://www.husl-colors.org/

scratchmex:
As said in docs 

>  seaborn provides an interface to the husl system (since renamed to HSLuv)

and in #1361, it is prudent to migrate all references from husl to hsluv maintaining backwards compatibility. This PR resolves #1361.

I tried to maintain all backwards compatibility possible. Any advice please comment down.


scratchmex:
scratchmex@2147adf fixes #1994 and scratchmex@0d1d364 test against it

mwaskom:
(Copying from your repo, which showed up in my notifications first).

Thanks — but I am not convinced that this is a necessary change.

Seaborn already packages husl.py internally with appropriate acknowledgement of the original source.

As far as I can tell, hsluv.py has the same code as seaborn's husl.py, just a different name.

This name has been used for many years in seaborn, and the change here will require people to change their code for no reason. I would like to avoid that, unless there is a compelling reason for it.

scratchmex:
Well, definitely the speed and precision of the implementation has changed. Also, there is full backwards compatibility. The previous file still there intact. 

I think one of the reasons why have changed name is because of visibility, at the begining I had a really trouble finding what "husl" was, I even thought all the colors where wrong because of that file. If we use this color schema for beautiful plots I think at least we should give the name more visiblity, even add it as a dependency instead of source file, the package is really lightweight, I'm not an expert but I don't see any problem with that. The possible fixes for that implementation could be useful for other libraries as well, that's the point of open source isn't it?
I think the change should be done later or sooner and if the change if fully backward compatible, why not?

mwaskom:
Because, like I said, it’s disruptive to existing code.

hartwork:
If I may jump in as a mediator here, it seems reasonable to me that:

- if HUSL has been progressed to HSLuv, users can expect to use that colour model (revision 4 or later) as "hsluv" in the future.  I consider [1](https://github.com/mwaskom/seaborn/pull/1995#issuecomment-599683108) [2](https://github.com/mwaskom/seaborn/issues/1361) and [3](https://github.com/vispy/vispy/issues/1776) indicators for that.

- seaborn's offers colour model "husl" today and users can expect proper deprecation

- bundled `husl.py` and `hsluv.py` have already diverged and HSLuv upstream will continue to evolve.  I don't see any strong reason for bundling code here over plain importing and depending on [PyPI hsluv](https://pypi.org/project/hsluv/) and [multiple reasons against bundling](https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies).

So I think seaborn needs:

- one pull request to add colour model "husl" (based on importing, not bundling, [PyPI hsluv](https://pypi.org/project/hsluv/))

- one pull request to mark "husl" as deprecated, after "hsluv" has become available

- one pull request to remove "husl" some time in the future

What do you think?

mwaskom:
I think I've been very clear. Absent a compelling reason to break existing code, I am not going to do it.

scratchmex:
> I think I've been very clear. Absent a compelling reason to **break existing code**, I am not going to do it.

I don't get why
```python
def husl_palette(*args, **kwargs):
    import warnings
    warnings.warn("husl changed name to hsluv. "
                  "You should use 'hsluv_palette' instead.",
                  FutureWarning)

    return hsluv_palette(*args, **kwargs)


def hsluv_palette(n_colors=6, h=.01, s=.9, l=.65):  # noqa
    """Get a set of evenly spaced colors in HSLuv hue space.
```

```python
    elif input in ("husl", "hsluv"):
        if input == "husl":
            import warnings
            warnings.warn("husl changed name to hsluv. "
                          "You should use 'hsluv' instead.",
                          FutureWarning)
        color = hsluv.hsluv_to_rgb(color)
```

```python
        elif palette in ("husl", "hsluv"):
            if palette == "husl":
                import warnings
                warnings.warn("husl changed name to hsluv. "
                              "You should use 'hsluv' instead.",
                              FutureWarning)
```

and mantaining the current `husl.py` file breaks the existing code. The changes **clearly** assure the previous references to *husl* keep sense. 
Furthermore this test shows the compatibility:

```python
    def test_husl_deprecation(self):

        nt.assert_warns(FutureWarning,
                        palettes._color_to_rgb, (0, 0, 0), input="husl")
        nt.assert_warns(FutureWarning, palettes.color_palette, "husl")
        nt.assert_warns(FutureWarning, palettes.husl_palette)
```

If I'm not seeing it, could you tell me please how this changes break existing code?.

PR #1999, opened 2020-03-17
Fix `add_legend` to always populate `_legend`

bennguvaye:
It gets populated when `legend_out` is `True`, but not when it is `False`.
This patch fixes this consistency problem.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -128,6 +128,7 @@ class Grid(object):
 
             leg = ax.legend(handles, labels, **kwargs)
             leg.set_title(title, prop={"size": title_size})
+            self._legend = leg
 
         return self
 


mwaskom:
Do you have an example of a problem that this caused? Remember, user code should never rely on attributes/methods that start with a single underscore.

bennguvaye:
No, indeed I hit this by directly accessing the `_legend` attribute.
It seems it's only ever used in tests, so it can't cause bugs.
I still think the fix is legit.

Indeed the real problem I hit is it looks like there is no other convenient way to manipulate the legend. A possibility could be to expose methods to get the default handles and labels for the grid.

Thank you

mwaskom:
It's an innocuous change so I'll merge it, but don't complain in the future if code that accesses `._legend` breaks, that's what the underscore is for!

bennguvaye:
Thank you, I won't complain !

PR #2010, opened 2020-03-30
Add optional argument showfliers for boxenplot

lukasberbuer:
If a big data set has a few outliers, the 

Same API as matplotlib boxplot to show/hide markers beyond the box(es)

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1808,7 +1808,8 @@ class _LVPlotter(_CategoricalPlotter):
 
     def __init__(self, x, y, hue, data, order, hue_order,
                  orient, color, palette, saturation,
-                 width, dodge, k_depth, linewidth, scale, outlier_prop):
+                 width, dodge, k_depth, linewidth, scale, outlier_prop,
+                 showfliers):
 
         # TODO assigning variables for None is unneccesary
         if width is None:
@@ -1834,6 +1835,10 @@ class _LVPlotter(_CategoricalPlotter):
         self.scale = scale
 
         self.outlier_prop = outlier_prop
+
+        if showfliers is None:
+            showfliers = True
+        self.showfliers = showfliers
 
         self.establish_variables(x, y, hue, data, orient, order, hue_order)
         self.establish_colors(color, palette, saturation)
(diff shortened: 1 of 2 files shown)


mwaskom:
To be honest, I’ve never understood why matplotlib has that parameter. If people make conscious choices during data preprocessing/cleaning to exclude certain observations, that can be reasonable. But its strange to have a visualization parameter that says “ignore data points that make my plot look bad.” Maybe I’m missing something....

lukasberbuer:
I get your point. Personally, I like to use boxenplots (actually through catplot) to get a rough understanding of big data sets before even starting preprocessing and cleaning. In that step, I don't want to know about a few extreme outliers. For me, the beauty of boxenplots is it's level of detail - somewhere between boxplots and histograms/distplots...

seaborn/categorical.py
```diff
@@ -1835,6 +1836,10 @@ def __init__(self, x, y, hue, data, order, hue_order,
 
         self.outlier_prop = outlier_prop
 
+        if showfliers is None:
```
mwaskom: Is this needed? The function signature can have `showfliers=True` as the default parameter.

seaborn/categorical.py
```diff
@@ -2664,6 +2673,9 @@ def boxenplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
         Proportion of data believed to be outliers. Used in conjunction with
         k_depth to determine the number of percentiles to draw. Defaults to
         0.007 as a proportion of outliers. Should be in range [0, 1].
+    showfliers : bool, optional
+        If True, outliers beyond ``outlier_prop`` are plotted as markers
```
mwaskom: I would change this to `If False, suppress the plotting of outliers`. I want to emphasize that hiding the outliers requires an explicit choice. Also, it's not quite right to say "outliers beyond `outlier_prop`" because it's not a threshold on the continuous value but rather a proportion of observations.

seaborn/tests/test_categorical.py
```diff
@@ -2631,7 +2631,8 @@ class TestBoxenPlotter(CategoricalFixture):
                        orient=None, color=None, palette=None,
                        saturation=.75, width=.8, dodge=True,
                        k_depth='proportion', linewidth=None,
-                       scale='exponential', outlier_prop=None)
+                       scale='exponential', outlier_prop=None,
+                       showfliers=False)
```
mwaskom: This should default to `True`.

lukasberbuer:
Thank you for your review. I'm not sure about the test. Do you think it is sufficient to just count the elements in the PathCollection with showfliers `True` and `False`?

PR #2025, opened 2020-04-15
ENH: Handle`legend.title_fontsize` rcparam

MaozGelbart:
`legend.fontsize` rcparam is a part of seaborn context. However the legend title isn't, and the title font size is taken as 85% of `axes.labelsize` rcparam. `legend.title_fontsize` rcparam [has been added](https://github.com/matplotlib/matplotlib/pull/11172) in 
`matplotlib==3.0`. This PR incorporates `legend.title_fontsize` rcparam into seaborn, for both plotting functions and rcmod context (which was set to 12, one point above `legend.fontsize` which currently equals 11). 

The change, if accepted, affects categorical plots, `lmplot` and `pairplot` but not relational plots, as they currently use an empty handle to draw a title-like line.

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1,6 +1,7 @@
 from itertools import product
 import warnings
 from textwrap import dedent
+from distutils.version import LooseVersion
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


mwaskom:
Thanks @MaozGelbart

Issue #2028, opened 2020-04-17
Grid color doesn't change with string variable but changes with categorical variable


paulrougieux:
When using string (character) variables the colour doesn't change
```
import seaborn
iris = seaborn.load_dataset("iris")
g = seaborn.FacetGrid(iris, col="species")
g.map(seaborn.scatterplot,'petal_length','petal_width','species')
```
![Screenshot from 2020-04-17 16-32-26](https://user-images.githubusercontent.com/5647397/79580463-4e614200-80c9-11ea-836e-dc6d5d51a840.png)

It does change when changing that to a categorical variable
```
iris['species'] = iris['species'].astype('category')
g.map(seaborn.scatterplot,'petal_length','petal_width','species')
```
![Screenshot from 2020-04-17 16-32-50](https://user-images.githubusercontent.com/5647397/79580431-45707080-80c9-11ea-829e-3157f1db4885.png)

The behaviour with string variables can be misleading, especially if the data frame contains  combinations of character variables which appear in one facet and are slightly different in another facet, you end up with colors which are not comparable between facets. and the legend only matches to one of the facet. 

```
iris = seaborn.load_dataset("iris")
iris['petal_length_below_4'] = iris['petal_length'] < 4
g = seaborn.FacetGrid(iris, col="petal_length_below_4")
g.map(seaborn.scatterplot,'petal_length','petal_width','species')
g.add_legend()
```
![Screenshot from 2020-04-17 16-44-18](https://user-images.githubusercontent.com/5647397/79581459-bcf2cf80-80ca-11ea-8252-4b2664c2807a.png)

mwaskom:
Yes, this is an unavoidable consequence of the way that `FacetGrid` works. That's why the docs for `scatterplot` have an example that says:

> Use relplot() to combine scatterplot() and FacetGrid: This allows grouping within additional categorical variables. Using relplot() is safer than using FacetGrid directly, as it ensures synchronization of the semantic mappings across facets.

```python
g = sns.relplot(x="total_bill", y="tip", col="time", hue="day", style="day", 
                kind="scatter", data=tips)
```

paulrougieux:
Thank you, I prefer flowers to tippers (because petals are nicer than smokers), the example then becomes:
```
import seaborn
iris = seaborn.load_dataset("iris")
g = seaborn.relplot(x='petal_length', y='petal_width', col='species', hue='species', data=iris)
```
[...]

PR #2030, opened 2020-04-19
Examples of ways to have the same mapping between facets

paulrougieux:
Following discussion in issue https://github.com/mwaskom/seaborn/issues/2028


mwaskom:
Thanks @paulrougieux

PR #2036, opened 2020-04-22
Update algorithms.py

DestinedGamerz:
mine doesn't work


mwaskom:
Hi, I'm not sure what you're getting at here, but this change fundamentally breaks the function.

If you have a bug, please submit a reproducible report to the issue tracker, thanks.

Issue #1455, opened 2018-06-14
[Boxplot] Gap Inside Grouped Boxes


schulzp:
Hi,

thanks for the seaborn library, it's a great tool!

I'd like to suggest an aesthetic improvement: A configurable gap between boxes inside a group of boxes: ![Boxplot with Gap](https://image.ibb.co/eqK2sy/seaborn_boxplot_gap.png)

Since I'm currently writing my thesis, I'm a bit short on time, all I've got for now is a patch. If you consider this feature worthwhile, I'd create a proper pull request, of course.

```patch
Index: categorical.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- categorical.py	(date 1528960332000)
+++ categorical.py	(date 1528960332000)
@@ -366,6 +366,7 @@
         n_levels = len(self.hue_names)
         if self.dodge:
-            each_width = self.width / n_levels
+            each_width = (self.width - (self.gap * (n_levels - 1))) / n_levels
             offsets = np.linspace(0, self.width - each_width, n_levels)
             offsets -= offsets.mean()
         else:
@@ -436,13 +437,14 @@
 
     def __init__(self, x, y, hue, data, order, hue_order,
                  orient, color, palette, saturation,
-                 width, dodge, fliersize, linewidth):
+                 width, gap, dodge, fliersize, linewidth):
 
         self.establish_variables(x, y, hue, data, orient, order, hue_order)
         self.establish_colors(color, palette, saturation)
 
         self.dodge = dodge
         self.width = width
+        self.gap = gap
         self.fliersize = fliersize
 
         if linewidth is None:
@@ -2203,12 +2205,12 @@
 
 def boxplot(x=None, y=None, hue=None, data=None, order=None, hue_order=None,
             orient=None, color=None, palette=None, saturation=.75,
-            width=.8, dodge=True, fliersize=5, linewidth=None,
+            width=.8, gap=0.1, dodge=True, fliersize=5, linewidth=None,
             whis=1.5, notch=False, ax=None, **kwargs):
 
     plotter = _BoxPlotter(x, y, hue, data, order, hue_order,
                           orient, color, palette, saturation,
-                          width, dodge, fliersize, linewidth)
+                          width, gap, dodge, fliersize, linewidth)
 
     if ax is None:
         ax = plt.gca()
```

Please let me know what you think.

Cheers,
Peter

mwaskom:
[...]

PR #2042, opened 2020-04-29
Use dodge to scale offset between nested plots

anki-xyz:
Unfortunately, there is no way to scale the offset between nested, grouped plots. By using the existing ```dodge``` variable as a float, one can easily now change the distance between individual plots.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -365,6 +365,7 @@ class _CategoricalPlotter(object):
             each_width = self.width / n_levels
             offsets = np.linspace(0, self.width - each_width, n_levels)
             offsets -= offsets.mean()
+            offsets *= float(self.dodge)
         else:
             offsets = np.zeros(n_levels)
 
@@ -2171,9 +2172,10 @@ _categorical_docs = dict(
         elements for one level of the major grouping variable.\
     """),
     dodge=dedent("""\
-    dodge : bool, optional
+    dodge : bool or float, optional
         When hue nesting is used, whether elements should be shifted along the
-        categorical axis.\
+        categorical axis (bool). If a float is passed, the elements are shifted
+        with an offset sclaed by the float. \
     """),
     linewidth=dedent("""\
     linewidth : float, optional
@@ -2317,6 +2319,15 @@ boxplot.__doc__ = dedent("""\
 
         >>> ax = sns.boxplot(x="day", y="total_bill", hue="smoker",
         ...                  data=tips, palette="Set3")
+
+    Draw a boxplot with nested grouping by two categorical variables
+    with a larger offset to each other:
+
+    .. plot::
+        :context: close-figs
+
+        >>> ax = sns.boxplot(x="day", y="total_bill", hue="smoker",
+        ...                  data=tips, palette="Set3", dodge=1.4)
 
     Draw a boxplot with nested grouping when some bins are empty:
 
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2327,7 +2327,7 @@ boxplot.__doc__ = dedent("""\
         :context: close-figs
 
         >>> ax = sns.boxplot(x="day", y="total_bill", hue="smoker",
-        ...                  data=tips, palette="Set3", dodge=1.4)
+        ...                  data=tips, palette="Set3", dodge=1.1)
 
     Draw a boxplot with nested grouping when some bins are empty:
 


mwaskom:
I'm not convinced this is a good addition.

Here is what the new example you have added looks like:
image

This is a very confusing image: it groups together visual elements that do not go together (the nonsmokers from one day and the smokers from the next day).

A change like this would also need to be applied to other functions with a similar API, and there would need to be tests.

But again, I'm not convinced: "this wasn't possible, now it is" is not a compelling argument.

anki-xyz:
hi, i do agree, the example is not pleasing, the ```dodge``` value was too high. If one uses 1.1 instead of 1.4, however, it is visually more pleasing:

image

compared to the original one:

image

My argument would be that I adjust each time the figure afterwards in Illustrator or Inkscape as the stacked barplots are visually not pleasing. Further, it is basically a single line of code that is changed in the ```_CategoricalPlotter``` base class.

mwaskom:
I would not recommend arguing for your change with general statements about what is or isn't "visually pleasing". The example plot you showed was not ugly, it was *wrong*. The less dodged plots are less actively incorrect, but still harder to parse correctly.

Just because something is a small change does not mean it's a good idea.

mwaskom:
I'm not completely opposed to this and have suggested it in the past (https://github.com/mwaskom/seaborn/issues/1455#issuecomment-397635877). But getting a change in requires thinking hard about all of the implications and ensuring consistency and accuracy, even if it's a small change.

mwaskom:
Another problem: `pointplot` already accepts a float for `dodge`, but applies it differently. Therefore, the following example doesn't work:

```
sns.boxplot(x="sex", y="total_bill", hue="smoker",
            data=tips, palette="pastel", dodge=1.1)
sns.pointplot(x="sex", y="total_bill", hue="smoker",
              data=tips, palette="dark", dodge=1.1)
```
image

anki-xyz:
I will test and follow this up more closely, and for the time being close this PR, thanks for checking!

mwaskom:
Let's leave the PR open. The idea has merit but it does require a bit more thorough thinking about the various implications.

I think that implementing a continuous `dodge` the way it works in `pointplot` would be annoying: it requires the user to compute the widths of the elements to get things right, which isn't something they need to know about. `pointplot` also treats `dodge=True` differently from `dodge=1` which isn't very Pythonic.

My main point is that absence of a feature can be better than presence of two conflicting features. But the best outcome of course is presence of two consonant features.

robince:
> Another problem: `pointplot` already accepts a float for `dodge`, but applies it differently. 

I came here looking for numerical dodge option for `boxplot`. I would like it to work exactly the same as `pointplot`. My usecase is replicating [grouped violin and boxplots](https://psyteachr.github.io/introdataviz/representing-summary-statistics.html#grouped-violin-boxplots).

For ungrouped they look great:
Screenshot 2022-05-02 at 08 09 40

But with grouping I can't get them to line up as I can't modify the boxplot dodge independently of the width (I can get them to align if the boxplots are wide). 

```
ax = sns.violinplot(data=dat_long, x='Condition', y='RT', hue='Language',
                    inner=None, linewidth=0, saturation=0.5,legend=False)
plt.setp(ax.collections, alpha=.5)
sns.boxplot(data=dat_long, x='Condition', y='RT', hue='Language',
            saturation=0.5, width=0.2,
            boxprops={'zorder': 2}, ax=ax, notch=True, medianprops={'linewidth': 0})
sns.pointplot(data=dat_long, x='Condition', y='RT', hue='Language', ci=68,
              join=False, dodge=0.4)
```
Screenshot 2022-05-02 at 08 12 16

I'm also stuck on removing excess legend entries (a `legend=False` option would be useful).

mwaskom:
With the new objects interface, the `Dodge` transform has a `gap` option:

```python
import seaborn.objects as so
from seaborn import load_dataset
tips = load_dataset("tips")
so.Plot(tips, "day", color="sex").add(so.Bar(), so.Hist(), so.Dodge(gap=.1))
```
image

In the medium-term, it's likely that the plotting functions will be refactored to use the objects behind the scenes and that will present the opportunity to introduce a `gap` parameter and get this behavior. I think that moots this PR, so I'm going to close it for now.

luispc:
It would be great to be able to add a small gap in nested boxplot. This possibility is already offered in ggplot and other chart oriented toolboxes.

PR #2050, opened 2020-05-05
Disable autoscaling on categorical axes. Closes #2049

### LINKED ISSUE #2049: stripplot has too much jitter and you cannot change it using the jitter parameter
Seems this broke over an update of a package. 

These are the versions of my packages on Python 3.7:

```
Package         Version
--------------- -------
cycler          0.10.0 
kiwisolver      1.2.0  
matplotlib      3.2.1  
numpy           1.18.4 
pandas          1.0.3  
pip             18.1   
pkg-resources   0.0.0  
pyparsing       2.4.7  
python-dateutil 2.8.1  
pytz            2020.1 
scipy           1.4.1  
seaborn         0.10.1 
setuptools      40.8.0 
six             1.14.0 
```

When I run the first example from the stripplot I get this. 

stripplot

So, too much jitter, much more than shown in the docs, and if try to change it with the jitter parameter it does not change.

hwalinga:
As discussed in issue. 

This PR changes the `auto=None` to `auto=False` in the set_{x,y}lim calls for the stripplot. This will prevent matplotlib rescaling the image to limits that just contain data, but instead will not rescale and keep the provided limits. 

Previously `auto=None` did not rescale as well, but this changed in matplotlib version 3.2.2. As this might be considered a bug this is possible reverted in next releases. See matplotlib/matplotlib#17331

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -400,10 +400,10 @@ class _CategoricalPlotter(object):
 
         if self.orient == "v":
             ax.xaxis.grid(False)
-            ax.set_xlim(-.5, len(self.plot_data) - .5, auto=None)
+            ax.set_xlim(-.5, len(self.plot_data) - .5, auto=False)
         else:
             ax.yaxis.grid(False)
-            ax.set_ylim(-.5, len(self.plot_data) - .5, auto=None)
+            ax.set_ylim(-.5, len(self.plot_data) - .5, auto=False)
 
         if self.hue_names is not None:
             leg = ax.legend(loc="best", title=self.hue_title)


mwaskom:
Thanks. Consensus at matplotlib seems to be that the change in 3.2.x is a genuine bug (unintended consequence of a related change). I think it will get fixed in a 3.2.3 release.

This PR does fix the issue, although it changes what happens when you draw a categorical plot in seaborn. It means (1) that additional plots added onto the same axes might have their data hidden if the data lie outside the limits that are set, and (2) adding data inside the limits won't cause them to snap back to "tight" margins around the categorical elements.

In my view, (2) is a desirable change, and (1) is potentially undesirable. But adding data onto a categorical plot using matplotlib methods directly, or a non-categorical seaborn function, already has substantial pitfalls that I see new users encounter with some regularity.

I think I am going to leave this PR open for now but include it in a planned broader effort to put more safeguards around mixing categorical and non-categorical plots.

mwaskom:
Would you please fill out the PR description a bit more and link to the relevant matplotlib issue so that this makes sense when revisited in the future? Thanks.

hwalinga:
Fixed in the latest release of matplotlib.

PR #2052, opened 2020-05-06
Require keyword arguments for most parameters

mwaskom:
This PR is a major and long-desired change that leverages Python 3 functionality to require most parameters to be defined with keyword arguments.

Most of the functions in the seaborn plotting API have a signature that looks roughly like:

```python
def plot(x, y, [semantic_vars, ...], data=None, [other_params, ...], **passthrough_kwargs):
    ...
```

Where `semantic_vars` are `hue`, `style`, etc.

This PR will enforce that ~~`data` and all subsequent params~~ most parameters, aside from those that specify data positions, be specified with explicit keyword arguments.

The motivation for requiring keyword arguments is that, when functions have long signatures, one would like to order the parameters in ways that are more meaningful than chronology. This is very much the case in seaborn, where a single function call has subsets of parameters that, e.g., control statistical aggregation or change the aesthetics of the plot.  In theory, inserting new parameters or reordering existing ones to achieve this has the potential to break user code, potentially in damaging ways (i.e., an argument might get passed to a parameter it is technically valid for but that produces an incorrect plot).

Seaborn has in the past sometimes inserted new parameters towards the end of the `other_params` under the sincere hope that people are not writing out 10+ positional arguments; this change will let us do that more safely in the future. But the real benefit of this change is the ability to define new `semantic_vars` while making it obvious what arguments can or cannot be used to extract values from the ``data`` object.

With this change, we will be able to

- standardize the signatures of the oldest functions that predated this convention (e.g., `lmplot`, where `hue` comes after `data`)
- add core semantics to older functions (i.e, `hue` or `style` for `kdeplot` or `regplot`)
- add new semantics going forward (e.g., `style` in categorical plots).

## Outstanding questions

- Should there be a grace period with a deprecation warning? scikit-learn has taken that approach and, we could borrow [their decorator](https://github.com/scikit-learn/scikit-learn/pull/13311/files) to do so. **A: Yes. 0.11 will warn; 0.12 will error.**

- Where should the `*` go? **A: For vector-oriented function, it goes after `x, y` (so before semantic variables and data. For multi-axes grid and matrix-oriented plotting functions, it goes after `data`).**

- What should the signature for the `*map` functions (`heatmap` and `clustermap` be?). **A: It will remain `func(data, ...)`, but this change will make it easy to add both location and semantic arguments to these functions (i.e. to pass long-form data with arguments to pivot into a matrix; #2054 ).** 

- What should the signature be for the `*Grid` objects? These are naturally long-form but don't have "leading semantics".  **A: ``data, ...``**.

- Are there other places we want to enforce keyword arguments? Maybe in `palettes.py`? **A: Perhaps, but this can be deferred**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ seaborn.egg-info/
 .cache/
 .coverage
 cover/
-.idea
+.idea/
+.vscode/
 .pytest_cache/
--- a/doc/releases/v0.11.0.txt
+++ b/doc/releases/v0.11.0.txt
@@ -2,6 +2,8 @@
 v0.11.0 (Unreleased)
 --------------------
 
+- Enforced keyword-only arguments for most parameters of most functions and classes.
+
 - Added an explicit warning in :func:`swarmplot` when more than 2% of the points are overlap in the "gutters" of the swarm.
 
 - Added the ``axes_dict`` attribute to :class:`FacetGrid` for named access to the component axes.
--- a/seaborn/_decorators.py
+++ b/seaborn/_decorators.py
@@ -1,0 +1,46 @@
+from inspect import signature, Parameter
+from functools import wraps
+import warnings
+
+
+# This function was adapted from scikit-learn
+# github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/validation.py
+def _deprecate_positional_args(f):
+    """Decorator for methods that issues warnings for positional arguments.
+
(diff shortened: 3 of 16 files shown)


.gitignore
```diff
@@ -7,5 +7,6 @@ seaborn.egg-info/
 .cache/
 .coverage
 cover/
-.idea
+.idea/
+.vscode/
```
dfm, line 11: 💯

PR #2071, opened 2020-05-12
Refactor variable processing

mwaskom:
### Summary and new features

This PR adds centralized and abstracted logic for processing both "long-form" and "wide-form" data inputs. The new code is currently used by the relational plotting functions (`relplot`, `scatterplot`, and `lineplot`). It will be extended to other modules in future PRs.

The relational plot capabilities are a superset of previous behavior. Several new things are possible:

- Named variables for long-form data can refer to named levels of a pandas `Index` or `MultiIndex`. Previously it was necessary to call `reset_index` before plotting.
- The `data` parameter can be a `dict` for wide-form data. Long-form `dict` was previously supported but not documented.
- A `DataFrame` used in wide-form mode can have a mixture of types; the non-numeric types will be removed before plotting. Previously this caused an error.
- Only one of `x` or `y` can be specified with long-form data, although this will produce an empty plot. Previously it produced an error.
- `relplot` now accepts all of these formats too, whereas previously it was limited to a long-form dataframe and named variables.

---

### Wide-form data handling

A major innovation here is writing abstracted code for handling wide-form data, where the semantics relate to the dimensions of the data object rather than the variables collected within it. That will make it easy to leverage this code in modules that handle wide-form data differently, and it will potentially allow control over how wide-form data is mapped to the vectors that define a plot.

Here is an example showing the new behavior of relational plots with "wide-form" inputs:

image

<details> <summary>Code for generating plot</summary>

```python
import numpy as np
import pandas as pd

df = pd.DataFrame(np.arange(12).reshape(3, 4), columns=list("abcd"), index=[0, 2, 4])
s = pd.Series([0, 10, 20], index=[1, 3, 5], name="s")

data_objects = {
    "wide_df": df,
    "wide_array": df.values,
    "list_of_series": [df["a"], s],
    "list_of_array": [df["a"].values, s.values],
    "list_of_list": [df["a"].tolist(), s.tolist()], 
    "list_of_dict": [df["a"].to_dict(), s.to_dict()],
    "dict_of_series": {"a": df["a"], "s": s},
    "dict_of_array": {"a": df["a"].values, "s": s.values},
    "dict_of_list": {"a": df["a"].tolist(), "s": s.tolist()},
    "dict_of_dict": {"a": df["a"].to_dict(), "s": s.to_dict()},
    "flat_series": s,
    "flat_array": s.values,
    "flat_list": s.tolist(),
    "flat_dict": s.to_dict(),
}

def show_wide_to_long(data_objects, module):

    fig, axes = plt.subplots(
        nrows=2, ncols=len(data_objects) // 2,
        figsize=(2 * len(data_objects) // 2, 2 * 2),
        constrained_layout=True
    )

    plot_func = {"rel": sns.lineplot}
    plot_manager = {"rel": sns.relational._RelationalPlotter}
    
    func = plot_func[module]
    p = plot_manager[module]()
    
    for ax, (kind, data) in zip(axes.T.flat, data_objects.items()):

        plot_data, variables = p.establish_variables(data)
        func(x="x", y="y", hue="hue", data=plot_data, ax=ax)
        ax.set_title(kind)

        
show_wide_to_long(data_objects, "rel")
```

</details>

---

### Future changes

Once keyword-only arguments are strictly enforced, the common API may become `func(data=None, *, x=None, y=None, ...)`. The advantage is that calling `func(data)` should now do something useful for many different ways that `data` can be defined.

Once the refactoring is complete, it will be necessary to rewrite the API documentation for the impacted functions. I would also like to write a more narrative document in the user guide/tutorial explaining how seaborn thinks about long-form and wide-form data.

Note that the relational plot tests very thoroughly exercise the new core code, but I haven't migrated those tests to the `test_core` module itself. I will rethink this while refactoring the categorical plots.


seaborn/core.py
```diff
@@ -0,0 +1,213 @@
...
+                for i, var in enumerate(data):
+                    key = getattr(var, "name", i)
+                    # TODO is there a safer/more generic way to ensure Series?
+                    # sort of like np.asarray, but for pandas?
+                    data_dict[key] = pd.Series(var)
+
+                data = data_dict
+
+            # Pandas requires that dict values either be Series objects
+            # or all have the same mapping
```
mwaskom, line 95: s/mapping/length

seaborn/conftest.py
```diff
@@ -12,3 +14,144 @@ def close_figs():
...
+
+@pytest.fixture
+def long_df():
+
+    n = 100
+    rs = np.random.RandomState()
+    df = pd.DataFrame(dict(
+        x=rs.randint(0, 20, n),
+        y=rs.randn(n),
+        a=np.take(list("abc"), rs.randint(0, 3, n)),
```
mwaskom, line 112: I think we can use `choice` now

seaborn/core.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def establish_variables_longform(self, data=None, **kwargs):
+        """Define plot variables given long-form data and/or vector inputs.
+
+        Parameters
+        ----------
+        data : dict-like collection of vectors
+            Input data where variable names map to vector values.
+        kwargs : variable -> data mappings
+            Keys are seaborn variables (x, y, hue, ...) and values are vectors
```
mwaskom, line 144: Values can also be column names.

seaborn/core.py
```diff
@@ -0,0 +1,213 @@
...
+            index = data.index.to_frame()
+        except AttributeError:
+            index = {}
+
+        # The caller will determine the order of variables in plot_data
+        for key, val in kwargs.items():
+
+            if isinstance(val, str):
+                # String inputs trigger __getitem__, first from data itself
+                try:
```
mwaskom, line 182: Use `key in data` rather than nested try/except?

PR #2086, opened 2020-05-19
Boxenplot bug fixes, small features and style changes

### LINKED ISSUE #803: Confusing lvplot results with small n
Paging @drewokane 

My understanding is that each feature in the letter value plot should correspond to an actual observation in the dataset. However, this seems to break down for small ns:

``` python
np.random.seed(0)
x = np.random.randn(6)
sns.lvplot(x)
sns.stripplot(x)
```

image

Is this correct behavior?

### LINKED ISSUE #2063: boxenplot with outlier_prop=0 shows 1 outlier
The comment and code here do not match and may explain it:
https://github.com/mwaskom/seaborn/blob/master/seaborn/categorical.py#L1884

### LINKED ISSUE #2085: boxenplot(x=x, showfliers=False) isn't autoscaling
Possibly related to #2049. When plotting a boxenplot with a single series and using the new `showfliers=False`, the result is not autoscaled and the plot is centered around the median. Explicit autoscaling after plotting fixes the problem. Consider the following code:
```python
import seaborn as sns; sns.set_style("darkgrid")
fig, axs = plt.subplots(3)
sns.boxenplot(x=tips["total_bill"], ax=axs[0], showfliers=True)
sns.boxenplot(x=tips["total_bill"], ax=axs[1], showfliers=False)
sns.boxenplot(x=tips["total_bill"], ax=axs[2], showfliers=False)
axs[2].autoscale_view(True)
for i, title in enumerate(["showfliers=True","showfliers=False","showfliers=False with autoscale"]):
    axs[i].set_title(title, loc="left")
plt.tight_layout()
```
Results with:

boxen_shrink

### LINKED ISSUE #2118: boxenplot "trustworthy" method uses `outlier_prop`
Originally posted here: [https://github.com/mwaskom/seaborn/pull/2086#issuecomment-631452922](https://github.com/mwaskom/seaborn/pull/2086#issuecomment-631452922) . Calling `boxenplot` with `k_depth="trustworthy"` uses `outlier_prop` to obtain the number of boxes but this is not documented. It may be better to introduce a new argument to control "trustworthy" parameterization.
See the code:
[https://github.com/mwaskom/seaborn/blob/dd40fd6885f9d164d6925f5b4571d69f1c134b4b/seaborn/categorical.py#L1865-L1868](https://github.com/mwaskom/seaborn/blob/dd40fd6885f9d164d6925f5b4571d69f1c134b4b/seaborn/categorical.py#L1865-L1868)
`p` equals `outlier_prop`.

MaozGelbart:
This PR incorporates several bug fixes, small features and style improvements to `boxenplot`:

- [x] Fixes a bug ([https://github.com/mwaskom/seaborn/pull/2086#issuecomment-630914580](https://github.com/mwaskom/seaborn/pull/2086#issuecomment-630914580)) where an extra box was erroneously added to `boxenplot`.
- [x] When only a single box is drawn, `boxenplot` uses the input color and not just white (revealed after resolving the previous bug).
- [x] `boxenplot` now autoscales the axes to make sure that all boxes are visible in the absence of outliers (#2085)
- [x] Allows `k_depth` to take a number and not only one of the predefined methods. 
- [x] Defaults `outlier_prop=0.007` also for the public `boxenplot` and not only for the private `_LVPlotter`. As a result this will be shown in the documentation page.
- [x] Specifying `k_depth="full"` now generates `int(log(n))+1` boxes (presumably covers all data points). Perhaps sufficient to address #2063 ? Due to the nature of order statistics this won't cover exactly all points for some datasets. 
- [x] `boxenplot` style now better matches `boxplot`. This includes the following changes: 
  - Use `self.gray` for plotting outliers instead of the color of first boxes.
  - Use `self.gray` for plotting boxes lines. This is also important since `Patch` objects are styled by seaborn default with white edges and when plotting on white background it obscures some boxes. This also matches the legend. 
  - The boxes are sequentially colored from the input color all the way to white. Now they use at least some color from their original color to distinguish between different hues. 
  - Use `"butt"` capstyle for medians to prevent the caps from extending beyond box edges.
- [x] Adds some arguments validation and remove internal unused arguments when class members are used. 
- [x] Avoids calling `ax.scatter` when there are no outliers to show.
- [x] Deprecates `lvplot`.
- [x] Adds tests to make sure all use cases are covered.
- [x] Improve example plot that shows swarmplot on top of boxenplot.
- [x] `"trustworthy"` method uses `outlier_prop` but this is not documented and probably misuse ([https://github.com/mwaskom/seaborn/pull/2086#issuecomment-632042987](https://github.com/mwaskom/seaborn/pull/2086#issuecomment-632042987)). Opened #2118 to track this. This is now fixed and also correctly parameterized with the addition of `trust_alpha` new parameter.
- [x] Change default `k_depth` to ~~`trustworthy` or `auto`~~ `tukey`. (#803 )
- [x] Improve docstring to emphasize the relationship between `k_depth` and `outlier_prop`.
- [x] Restrict `outlier_prop` to be in the range of (0,1]. Closes #2063 
- [x] Draft a release notes entry. 

Example of the new style, compare with the current one [https://seaborn.pydata.org/_images/seaborn-boxenplot-2.png](https://seaborn.pydata.org/_images/seaborn-boxenplot-2.png)
2086d

Will be happy for any comment!

Fixes #2086 (comment)
Fixes #2085 
Fixes #803 
Fixes #2118 
Closes #2063


mwaskom:
Thanks for taking a crack at this.

I'm increasingly concerned that there is either a bug in the boxenplot code or that the documentation of this parameter needs to be clarified.

This example suggests that the number of outliers drawn differs from what is specified in `outlier_prop` by roughly a factor of 2 and then a factor of 10.
```python
f, ax = plt.subplots()
n = 10000
x = np.random.standard_t(10, n)
props = [0, .005, .01, .02, .04, .08, .16]
for p in props:
    sns.boxenplot(x=x, y=np.full(n, p), outlier_prop=p, orient="h", order=props)
    ax.collections[-1].set_linewidth(0)
    print(p, len(ax.collections[-2].get_offsets()) / n)
```
```
0 0.002
0.005 0.001
0.01 0.002
0.02 0.004
0.04 0.008
0.08 0.0158
0.16 0.0314
```

I think the factor of 2 can be explained by a distinction between "per distribution" and "per tail" but I don't know where the factor of 10 is coming from.

I also think it's a misconception that setting `outlier_prop` to 0 is meant to omit outliers, and the approach of just hiding the outlier glyphs is wrong. If we are going to use 0 to mean "don't show them", then the boxes should be extended all the way out to the extremes. But based on skimming the paper, I think it might make more sense to constrain this value to be positive.

Either way, I need to read the letter-value plot paper closely and then go through this code again (which was originally and external contribution) to make sense of what's happening better.

MaozGelbart:
> I think the factor of 2 can be explained by a distinction between "per distribution" and "per tail" but I don't know where the factor of 10 is coming from.

The `range` clause here is generating an extra level and I think this is the source of the factor 2 you observe:
[https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1860-L1864](https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1860-L1864)

> I also think it's a misconception that setting `outlier_prop` to 0 is meant to omit outliers, and the approach of just hiding the outlier glyphs is wrong. 

I agree, if outliers are calculated this is erroneous. The problem in this case are these lines:
[https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1837-L1840](https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1837-L1840)
Because `outlier_prop = 0` evaluates to `False` and causes fallback to default of 0.007. I will fix that but I think the approach of avoiding `ax.scatter` calls when outliers plotting is suppressed is desirable. 

> If we are going to use 0 to mean "don't show them", then the boxes should be extended all the way out to the extremes. But based on skimming the paper, I think it might make more sense to constrain this value to be positive.

From reading into the paper I understand that `outliers_prop=0` means "no observation may be considered an outlier" (or if to use the paper's jargon, "no stopping rule"). Therefore as you write k should be increased -> boxes should be added to reach the extremes. I will fix this. I also think having this possibility is an important option, just as boxplot whiskers may cover all extreme observations if they're close enough to IQR. Anyway `outliers_prop` parameter has an interplay with `k_depth` parameter and this perhaps needs a little bit more explanation in their description.

I will put more effort on this later, and hopefully will also address #803 as well and a few visual improvements.

mwaskom:
One other thing I noticed when I was looking at this yesterday is that the code implementing the "trustworthy" method used `outlier_prop` too, but a quick read through the equations suggested that's wrong ... that method is parameterized by a precision measure that is different from `outlier_prop`. Can you check whether those are correct, while you're taking a quick look?

mwaskom:
> Because outlier_prop = 0 evaluates to False and causes fallback to default of 0.007.

That sounds right. This should just be a default value in the function signature rather than `None` with a conditional that sets a default anyway, if the default isn't inferred using any other information.

MaozGelbart:
> One other thing I noticed when I was looking at this yesterday is that the code implementing the "trustworthy" method used `outlier_prop` too, but a quick read through the equations suggested that's wrong ... that method is parameterized by a precision measure that is different from `outlier_prop`. Can you check whether those are correct, while you're taking a quick look?

Indeed, it looks like an overloading of the `outlier_prop` parameter into "trustworthy" method. The value represents a confidence interval of a two-tailed box which is defined as `(1-alpha)/2`. The authors suggest the usual alpha=0.05 which I think should be a constant here instead:
[https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1849-L1850](https://github.com/mwaskom/seaborn/blob/2d3502b05d91343bd37d8f8074874cfbae9f1e68/seaborn/categorical.py#L1849-L1850)

mwaskom:
Just a heads up that #2090 just landed with a lot of changes/reorganization. I didn't touch the categorical module, but you'll probably want to avoid straying outside of it here or else we might end up with a real merge headache.

mwaskom:
I will try to go through the code later today, but my first impression is that the new plot looks *really* good. Great job!

MaozGelbart:
> Fix `"trustworthy"` method to use constant instead of taking `outlier_prop` into account ([#2086 (comment)](https://github.com/mwaskom/seaborn/pull/2086#issuecomment-632042987))

According to the paper, properly parameterizing this method would require a new input variable if we don't want to abuse "outlier_prop". If we use a "default" 0.05 parameterization, according to the paper this equals to the "tukey" method already implemented. With this PR an explicit number of boxes may be achieved using `k_depth=scalar` so I'm not sure we need to keep the "trustworthy" method as it is today. In my opinion, it may be changed to "tukey" synonym and possibly deprecated.

MaozGelbart:
> Change default `k_depth` to `trustworthy` or `auto`. Should this get a FutureWarning before the change? (#803 )

A general problem is that each series has its own calculation of boxes, and if we use "auto" this may result in an inconsistent ways of boxes calculations between different series. Perhaps figure this out in a separate PR? I'd be +1 for setting "tukey" as a default method (5-8 outliers for each tail) but this may require a deprecation period. Also for some examples this may show only 2-3 boxes (which will cover 75-87% of the data points).

mwaskom:
> A general problem is that each series has its own calculation of boxes, and if we use "auto" this may result in an inconsistent ways of boxes calculations between different series.

That's a good argument against "auto".

I'm probably leaning towards `tukey` as a default and deferring a decision on adding a parameter for trustworthy. That would let us switch to trustworthy in the future in a non-breaking way. I'm ambivalent about doing a release with a `FutureWarning` ... that makes sense for library code where it can be fixed once, but for interactive usage, the warning will be very annoying (or require adding a parameter ever time).

Given that default behavior will be changing any way with the bug fix (and that this is a relatively underused function to begin with), a clean break might be best.

> Also for some examples this may show only 2-3 boxes (which will cover 75-87% of the data points).

It probably makes sense to change the examples to use a larger dataset — `diamonds` would be a good candidate, but all of its categorical variables have a large number of levels, making for very busy plots with hue mapping. This can be handled separately.

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
-        if saturation is None:
-            saturation = .75
         self.saturation = saturation
 
-        if k_depth is None:
-            k_depth = 'proportion'
+        if k_depth not in ('proportion', 'tukey', 'trustworthy') and (
+            not isinstance(k_depth, numbers.Number)
+        ):
+            raise ValueError(f'k_depth "{k_depth}" not recognized')
```
mwaskom, line 1809:
```suggestion
k_depth_methods = ["proportion", "tukey", "trustworthy"]
if k_depth not in k_depth_methods and not isinstance(k_depth, Number):
    msg = (f"k_depth must be one of {k_depth_methods} or a number, "
           f"but {k_depth} was passed.")
    raise ValueError(msg)
```

mwaskom: (Requires direct import of `Number` I guess)

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
 
+        # TODO seems this member is only used to frame the legend artists
         if linewidth is None:
             linewidth = mpl.rcParams["lines.linewidth"]
         self.linewidth = linewidth
 
-        if scale is None:
-            scale = 'exponential'
+        if scale not in ('linear', 'exponential', 'area'):
+            raise ValueError(f'Scale "{scale}" not recognized')
```
mwaskom, line 1818: Similar suggestion as above.

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
-                                - np.log2(2 * stats.norm.ppf((1 - p)) ** 2))
-        }
-        k = k_dict[k_depth]
-        try:
-            k = int(k)
-        except ValueError:
-            k = 1
-        # If the number happens to be less than 0, set k to 0
-        if k < 1.:
+        if p == 0:
```
mwaskom, line 1840: Do I understand correctly that `outlier_prop` is only relevant when `k_depth == 'proportion`' *or* if it is set to 0? That might be ok, but it does seem a little convoluted.

Maybe `k_depth="full"` would be cleaner?

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
+        elif self.k_depth == 'tukey':
+            # This results with 5-8 points in each tail
+            k = int(np.log2(n)) - 3
+        elif self.k_depth == 'proportion':
+            k = int(np.log2(n)) - int(np.log2(n * p)) + 1
+        elif self.k_depth == 'trustworthy':
+            k = int(np.log2(n / (2 * stats.norm.ppf((1 - p)) ** 2))) + 1
+        else:
+            k = int(self.k_depth)  # allow having k as input
+        # If the number happens to be less than 1, set k to 1
```
mwaskom, line 1852: What circumstances lead to this?

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
+        if k < 1:
             k = 1
+
         # Calculate the upper box ends
-        upper = [100 * (1 - 0.5 ** (i + 2)) for i in range(k, -1, -1)]
+        upper = [100 * (1 - 0.5 ** (i + 1)) for i in range(k, 0, -1)]
```
mwaskom, line 1857: Is there any reason not to add 1 to `k` before doing this math and the math in `_lv_outliers`? (while changing `range(k, 1, -1)`?

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
         # Calculate the lower box ends
-        lower = [100 * (0.5 ** (i + 2)) for i in range(k, -1, -1)]
+        lower = [100 * (0.5 ** (i + 1)) for i in range(k, 0, -1)]
```
mwaskom, line 1859: See above

seaborn/categorical.py
```diff
@@ -1797,78 +1799,72 @@ def __init__(self, x, y, hue, data, order, hue_order,
...
         # Stitch the box ends together
         percentile_ends = [(i, j) for i, j in zip(lower, upper)]
         box_ends = [np.percentile(vals, q) for q in percentile_ends]
         return box_ends, k
 
     def _lv_outliers(self, vals, k):
         """Find the outliers based on the letter value depth."""
-        perc_ends = (100 * (0.5 ** (k + 2)), 100 * (1 - 0.5 ** (k + 2)))
+        perc_ends = (100 * (0.5 ** (k + 1)), 100 * (1 - 0.5 ** (k + 1)))
```
mwaskom, line 1867: See above

seaborn/categorical.py
```diff
@@ -1942,41 +1936,47 @@ def horz_perc_box(x, b, i, k, w):
             hex_color = mpl.colors.rgb2hex(color)
 
             if vert:
-                boxes = [vert_perc_box(x, b[0], i, k, b[1])
-                         for i, b in enumerate(zip(box_ends, w_area))]
+                box_func = vert_perc_box
+                xs_median = [x - widths / 2, x + widths / 2]
+                ys_median = [y, y]
+                xs_outliers = np.repeat(x, len(outliers))
```
mwaskom, line 1942: Not a big deal but `np.full(len(outliers), x)` might be more idiomatic.

seaborn/categorical.py
```diff
@@ -2032,14 +2027,13 @@ def draw_letter_value_plot(self, ax, kws):
...
-                                 vert=vert,
                                  widths=self.nested_width,
-                                 k_depth=self.k_depth,
                                  ax=ax,
-                                 scale=self.scale,
-                                 outlier_prop=self.outlier_prop,
                                  **kws)
 
+        # Autoscale the values axis to make sure all patches are visible
+        ax.autoscale_view(scalex=self.orient == "h", scaley=self.orient == "v")
```
mwaskom, line 2035: Is this sufficient or do we need to autoscale on both dimensions now? I can't remember which of the recent changes in matplotlib were intentional and which were bugs.

mwaskom: There is a relevant issue in swarmplot I am thinking of.

seaborn/categorical.py
```diff
@@ -2663,21 +2645,22 @@ def boxenplot(
     {saturation}
     {width}
     {dodge}
-    k_depth : "proportion" | "tukey" | "trustworthy", optional
+    k_depth : "proportion", "tukey", "trustworthy" or scalar, optional
```
mwaskom, line 2648: I think the official numpydoc standard is to enclose options in curly braces. You need to double up the curly braces to not mess with the string formatting.

seaborn/categorical.py
```diff
@@ -2663,21 +2645,22 @@ def boxenplot(
...
         The number of boxes, and by extension number of percentiles, to draw.
         All methods are detailed in Wickham's paper. Each makes different
         assumptions about the number of outliers and leverages different
-        statistical properties.
+        statistical properties. If `"proportion"`, draw no more than
+        `outlier_prop` extreme observations.
     {linewidth}
-    scale : "linear" | "exponential" | "area"
+    scale : "linear", "exponential" or "area", optional
```
mwaskom, line 2655: Same as above.

seaborn/categorical.py
```diff
@@ -2663,21 +2645,22 @@ def boxenplot(
...
         results visually. "linear" reduces the width by a constant linear
         factor, "exponential" uses the proportion of data not covered, "area"
         is proportional to the percentage of data covered.
     outlier_prop : float, optional
-        Proportion of data believed to be outliers. Used in conjunction with
-        k_depth to determine the number of percentiles to draw. Defaults to
-        0.007 as a proportion of outliers. Should be in range [0, 1].
+        Proportion of data believed to be outliers. Used to determine the
+        number of boxes to plot when `k_depth="proportion"`. If
+        `outlier_prop=0`, the plot will contain `log(n)+1` boxes.
```
mwaskom, line 2663: See comment above about the resolution of `k_depth` and `outlier_prop`

mwaskom:
Also can you add a draft of a release notes entry to the TODO?

seaborn/categorical.py
MaozGelbart: Using "tukey" with less than 8 observation, for example.

MaozGelbart:
> It probably makes sense to change the examples to use a larger dataset — `diamonds` would be a good candidate, but all of its categorical variables have a large number of levels, making for very busy plots with hue mapping. This can be handled separately.

Opened #2114 .

seaborn/categorical.py
MaozGelbart: I reckon that calling `ax.plot` for plotting the medians takes care of the width dimension.

MaozGelbart:
> I'm probably leaning towards `tukey` as a default and deferring a decision on adding a parameter for trustworthy.

Opened #2118 for tracking the trustworthy parameterization issue.

mwaskom:
> Opened #2118 for tracking the trustworthy parameterization issue.

Thanks, but I think it would be best to fix the trustworthy computation here. It's both technically incorrect and also a breaking change, so better to have all the disruptive fixing happening in one step.

If you want to punt on adding a parameter, it's fine to have trustworthy be redundant for now. That can be fixed in the future in a non-breaking way.

I'm also happy to add `alpha` as a parameter with 0.05 as a default. We can live with "tukey" being redundant, or we can deprecate tukey here.

seaborn/categorical.py
MaozGelbart: Thanks, will change this.

seaborn/categorical.py
MaozGelbart: For this enumeration, it seems more natural to use `k` since it refers to the number of boxes; the outer edge of box `i` is the beginning of box `i+1` hence the `i+1`. I will note this in the comment above this line. For `_lv_outliers` I will add 1 to k.

seaborn/categorical.py
MaozGelbart: Thanks, will do this.

seaborn/categorical.py
MaozGelbart: Thanks, will fix this.

seaborn/categorical.py
MaozGelbart: Thanks, will do.

seaborn/categorical.py
MaozGelbart: Thanks, will change this.

seaborn/categorical.py
MaozGelbart: Indeed. You suggestion is much better, will add "full". Note that this will not imply last box goes all the way towards the outliers; just having log(n)+1 boxes. It will be clearly documented.

seaborn/categorical.py
MaozGelbart: Will be fixed with `k_depth="full"`.

MaozGelbart:
> I'm also happy to add `alpha` as a parameter with 0.05 as a default. We can live with "tukey" being redundant, or we can deprecate tukey here.

I propose adding a new parameter `box_alpha` to control this. Since "tukey" is mentioned in the paper it may stay for convenience (and with this PR also the default).

seaborn/categorical.py
MaozGelbart: Thanks, updated the code here.

mwaskom:
It occurs to me that `box_alpha` is probably no less ambiguous than `alpha`, as to whether it's about computation or appearance. Perhaps `trust_alpha` would be better?

MaozGelbart:
> It occurs to me that `box_alpha` is probably no less ambiguous than `alpha`, as to whether it's about computation or appearance. Perhaps `trust_alpha` would be better?

Thanks, renamed to `trust_alpha`.

seaborn/categorical.py
```diff
@@ -1942,41 +1953,47 @@ def horz_perc_box(x, b, i, k, w):
...
-                # Plot the medians
-                ax.plot([y, y], [x - widths / 2, x + widths / 2],
-                        c='.15', alpha=.45, **kws)
+            boxes = [box_func(x, b[0], i, k, b[1])
+                     for i, b in enumerate(zip(box_ends, w_area))]
 
-                ax.scatter(outliers, np.repeat(x, len(outliers)),
-                           marker='d', c=hex_color, **kws)
+            # Plot the medians
+            with axes_style(rc={"lines.solid_capstyle": "butt"}):
```
mwaskom, line 1973: This parameter can go directly to `plot`
e.g.
```python
f, ax = plt.subplots()
ax.plot([1, 1], lw=10, solid_capstyle="butt")
```
Unless I'm missing something.

seaborn/categorical.py
```diff
@@ -1795,80 +1797,91 @@ class _LVPlotter(_CategoricalPlotter):
...
+            msg = f'scale must be one of {scales}, but {scale} was passed.'
+            raise ValueError(msg)
         self.scale = scale
 
+        if ((outlier_prop > 1) or (outlier_prop <= 0)):
+            msg = f'outlier_prop {outlier_prop} not in range (0, 1]'
+            raise ValueError(msg)
         self.outlier_prop = outlier_prop
 
+        if not 0 < trust_alpha < 1:
```
mwaskom, line 1829: This says `< 1` but the range in the error message is inclusive at 1.

seaborn/categorical.py
```diff
@@ -1795,80 +1797,91 @@ class _LVPlotter(_CategoricalPlotter):
...
+            # This results with 5-8 points in each tail
+            k = int(np.log2(n)) - 3
+        elif self.k_depth == 'proportion':
+            k = int(np.log2(n)) - int(np.log2(n * p)) + 1
+        elif self.k_depth == 'trustworthy':
+            k = (
+                int(
+                    np.log2(
+                        n
+                        / (2 * stats.norm.ppf((1 - self.trust_alpha / 2)) ** 2)
```
mwaskom, line 1861: Could clean this code up by defining `2 * stats.norm.ppf((1 - self.trust_alpha / 2)) ** 2` as an intermediate variable, then the rest will fit on one line.

seaborn/categorical.py
```diff
@@ -2663,21 +2662,27 @@ def boxenplot(
...
-        Proportion of data believed to be outliers. Used in conjunction with
-        k_depth to determine the number of percentiles to draw. Defaults to
-        0.007 as a proportion of outliers. Should be in range [0, 1].
+        Proportion of data believed to be outliers. Must be in the range
+        (0, 1]. Used to determine the number of boxes to plot when
+        `k_depth="proportion"`.
+    trust_alpha : float, optional
+        Confidence level for a box to be plotted. Used to determine the
+        number of boxes to plot when `k_depth="trustworthy"`. Must be in the
+        range (0, 1).
```
mwaskom, line 2685: See note about inclusive/exclusive range here. I'm not sure whether the code or comments need to be fixed.

mwaskom:
This is just about done. I pulled the code and played around, and the aesthetics of the plots are definitely better, and I also think they look much more reasonable with small/medium data. Great job seeing these extensive improvements through.

seaborn/categorical.py
MaozGelbart: Will change this. 
My original intent was allowing the user overriding this through kwargs, but since `plot` and `scatter` kwargs are shared this is not possible anyway. And perhaps we don't even want users to change this?

seaborn/categorical.py
MaozGelbart: Thanks, will change the error message here.

seaborn/categorical.py
MaozGelbart: For sure, thanks!

seaborn/categorical.py
MaozGelbart: Will fix the code, thanks!

seaborn/categorical.py
mwaskom: Indeed, lack of full-customization via specific kwargs is a problem throughout the categorical module. Could be improved, but doesn't need to happen here.

PR #2090, opened 2020-05-22
Refactor semantic mapping operations

mwaskom:
This PR moves the semantic mapping operations originally developed for the relational plots down to the core level and substantially changes how those operations are implemented. Much of the logic has been moved out of the `Plotter` classes and into new `Mapping` classes. This makes the body of the `*plot` functions a bit more verbose (but only a bit) while significantly streamlining the `Plotter` classes. (It also hints at the makings of an expressive alternate external API. Currently all `_core` functionality remains private.)

I don't believe there are any intentional changes to the external API vis a vis the relational plots. Once other modules are moved onto this infrastructure, there may be changes in behaviors that are inconsistent across the library.

There is a significant challenge in totally decoupling the `Plotter` and `Mapping` classes that arises with the `size` semantic. Whereas hue values are artist-independent and style values are specified in terms of specific artist attributes, the size values are externally abstract but internally map onto different attributes depending on what artist ends up being used. Currently this means that the `SizeMapping` needs to access default size information from the `Plotter` object it is attached to. This is not ideal, because we'd like the `Plotter` to become generic and have all artist-specific logic live downstream from the semantic mappings. But the downstream abstraction is not currently well-defined, so we have to live with this for now.

Possible solutions include:

- Internally storing a normalized range that gets resolved at draw time. This proved very confusing.
- Work more like `StyleMapping` where the customization API involves specific attributes (`width`, `radius`, etc.) while maintaining all relevant defaults at the `Mapping` level.
- Some as-yet-to-be determined interactions at draw time, e.g. the lookup is lazy and doesn't happen until the Mapping is attached to some drawing-related class that knows what defaults to use.

This PR also does not refactor or improve the legend creation, which remains challenging.

Also note that the PR started with an attempt to rewrite the distribution plots on top of the new core infrastructure and some of the products from that effort are included in the changeset, but not as live code.


mwaskom:
Maybe all three semantics should work like `style` currently does, where they internally manage multiple matplotlib artist attributes and it's up to the code that is engaging with matplotlib to ask for the right ones?

- hue: facecolor, edgecolor
- size: linewidth, radius, width etc.
- style: markers, dashes, hatches?, etc.

mwaskom:
I'm kind of liking that last idea. Maybe the syntax for assigning attributes to be mapping variables could be something like

```
sns.violinplot(data=data, x=x, y=y, hue=group, facecolor=".8", edgecolor="@hue")
```

(Or some special seaborn function/class if we want to avoid special string weirdness).

Anyway, not a blocker for this PR.

PR #2098, opened 2020-05-24
DOC: small fixes

MaozGelbart:
This PR contains two small fixes to seaborn docs:
* Update kdeplot example to use keyword-only arguments, avoiding a warning
* Update intersphinx path to new numpy docs location


mwaskom:
Thanks!

Issue #796, opened 2015-12-10
KDE plot behaves inconsistently with and without statsmodels


jwdebelius:
I'm using KDE plots to generate a kernel smoothed plot in Seaborn 0.6.0. I used the following code to generate a plot with a vector of data:

``` python
import numpy as np
import seaborn as sn
import matplotlib.pyplot as plt

sn.set_style('ticks', {'axes.facecolor': 'none'})
ax = plt.axes()
sn.kdeplot(data, ax=ax, legend=False)
sn.kdeplot(data, ax=ax, legend=False, clip=[5, 10])
```

The first time I used the function, I had a conda environment with seaborn installed, but no statsmodels. It rendered a plot where clip limited the shaded area in the second plot, but the shaded area had the same kernel as the parent function. This is what I'd like.

![no_statsmodels](https://cloud.githubusercontent.com/assets/5534846/11707721/d282601c-9ebc-11e5-82a2-5f96e10e98a7.png)

I then conda installed statsmodels (the latest version via conda install) because I wanted to use some of that functionality in that package. With statsmodels in the environment, the kernel smoothing on my clipped area changed, and it fit the data for only that subset, creating a new smoothed kernel.

![statsmodels](https://cloud.githubusercontent.com/assets/5534846/11707622/b94610ea-9ebb-11e5-8cce-53e23840e190.png)

It would be really nice to have a consistent behavior. For my application, shading a clipped area of the same plot is handy.

mwaskom:
FWIW the second behavior is probably the "right" one, so these could be made more consistent, but I'm not sure it would help you.

mwaskom:
A robust, but more verbose, approach would be

```

JWDebelius:
Thanks!

mwaskom:
Closed in #2104 with the removal of the statsmodels backend

PR #2104, opened 2020-05-27
Modernize kdeplot and rugplot

### LINKED ISSUE #61: kdeplot should account for prior proportions when drawing multiple densities
Currently, if plotting separate densities for the levels of some factor, the distributions are not scaled by the prior proporition

mwaskom:
`kdeplot` and `rugplot` are two of the oldest parts of the library and predate the common API adopted for almost all other functions. This limits them and causes inconsistencies. They also were not well-tested or organized, making it difficult to add or modify even small features.

With the core refactoring done in #2090 and elsewhere, it's now time to bring these functions in line with the rest of the library. This PR accomplishes that. Doing so is necessarily disruptive. I have tried to minimize breakage, while still implementing a number of substantial changes that are important to understand.

I hope that, on balance, both the better API and numerous new features outweigh the costs of dealing with the API breaks.

# Standardized API

The most obvious new feature is that the functions can now be invoked with the same syntax as in the rest of seaborn, including drawing densities and rugs for subsets of the data in different colors using `hue` mapping. For example:

```python
import seaborn as sns
semantics = dict(x="total_bill", y="time")
sns.kdeplot(data=tips, **semantics)
sns.rugplot(data=tips, **semantics)
```
image

Note how we have assigned variables from a long-form data structure to multiple semantic roles in each plot: `x` and `hue`.

To draw the density on the y axis, rather than use the `vertical` parameter, simply change the assignment of variables:

```python
semantics = dict(y="total_bill", hue="time")
sns.kdeplot(data=tips, **semantics)
sns.rugplot(data=tips, **semantics)
```
image

Passing both x and y estimate and plot a bivariate kernel density estimate:

```python
iris = sns.load_dataset("iris")
sns.kdeplot(data=iris, x="sepal_width", y="petal_width", hue="species")
```
image

`rugplot` can also now draw a "bivariate" rug in a single function call:

```python
semantics = dict(x="total_bill", y="tip", hue="time")
sns.scatterplot(data=tips, **semantics)
sns.rugplot(data=tips, **semantics)
```
image

It's also possible to plot wide-form data, which draws a separate distribution for each column in the dataframe, using hue-mapping.

```python
sns.kdeplot(data=iris)
```
image

There is a deprecation layer that should catch the old keyword argument names and related parameters.

Currently the function signatures are disorganized, because all new parameters have been added at the end. When keyword-only arguments are enforced (after the 0.11 release), these parameters can be reorganized.

Positional `x` is allowed because `kdeplot(x=x)` does the same thing as `kdeplot(data=x)`. In 0.12, `data` will become the first and only valid positional argument.

Closes https://github.com/mwaskom/seaborn/issues/595

### API changes

In `rugplot`:
```diff
+ x, y, hue, data
-a, axis, vertical
```

In `kdeplot`:
```diff
+ x, y, hue, data
- data, data2, vertical
```
(Previously, `data` was the name for the `x` vector in `kdeplot`. Because of how "flat" data are processed by the distribution plots, there is no difference between `kdeplot(data=x)` and `kdeplot(x=x)`. But `data` is now interpreted differently in the context of any additional semantics.

# Dropping the statsmodels KDE backend

`kdeplot` has always been able to use both `statsmodels` and `scipy` to compute the density estimate. The `statsmodels` backend would be used if it was installed, with `scipy` serving as a fallback.

There are a number of differences between the `statsmodels` and `scipy` interfaces that caused subtle issues for non-default plots in seaborn. Additionally, statsmodels offered three primary advantages (1) non-gaussian kernel (2) FFT-based computation (3) cumulative densities. For (1), it does not seem that non-gaussian kernels are especially useful for visualization. I looked through the github search results for `kdeplot kernel` and it seemed this feature was used almost exclusively in a pedagogical context. ("These are what the different kernels look like"). For (2) the FFT computation can be a *lot* faster on a dense grid, but there is minimal difference between the two methods on the size of the default grid used in seaborn. For (3), it was not difficult to implement cumulative densities on top of what scipy gives you.

Therefore, I made the decision to drop the `statsmodels` backend; the new `kdeplot` uses only `scipy`. This has the effect of addressing several outstanding issues.

If this decision is going to cause substantial problems to downstream users/libraries, please open an issue to discuss.

Closes https://github.com/mwaskom/seaborn/issues/796
Closes https://github.com/mwaskom/seaborn/issues/938

### API changes

```diff
- kernel
```

# Other enhancements/changes

With the rewrite cleaning up the code, a number of new and long-desired features could be added. These mostly pertain to `kdeplot`

## Normalization of hue levels

The KDE is computed on each hue-based subset independently. But the default behavior is to scale the resulting densities by the proportion of each category in the full dataset, such that the total area under all of the curves sums to 1.

```python
sns.load_dataset("diamonds")
ax = sns.kdeplot(data=diamonds, x="carat", hue="cut")
```
image

This can be controlled with the `common_norm` parameter.

Fixes #61 (the oldest open issue in seaborn!)

### API changes

```diff
+ common_norm
```

## Multiple ways to visualize the hue subsets

The default behavior is to "layer" each curve, but it is also possible to stack them:

```python
ax = sns.kdeplot(data=diamonds, x="carat", hue="cut", multiple="stack')
```

image

and the stack can be normalized so that the curves give the conditional frequency at each `x` level:

```python
ax = sns.kdeplot(data=diamonds, x="carat", hue="cut", multiple="fill')
```

image

### API changes

```diff
+ multiple, common_grid
```

## Better control over bandwidth

The old version of `kdeplot` had a `bw` parameter which defaulted to the implementation in the compute backend when given a scalar: for `statsmodels` it set the actual width of the kernel, whereas `scipy` set the width of the kernel as a scaling factor multiplied by the data standard deviation.

Neither approach is especially helpful for visualization. Therefore, this PR deprecates the confusing `bw` and adds two new parameters

- `bw_method`, which is passed directly to `scipy`
- `bw_adjust`, which scales the bandwidth that `scipy` choses.

This means that you can let scipy use a reference rule but then nudge the kernel to be narrower or wider in a completely distribution-independent way.

### API changes
```diff
+ bw_method, bw_adjust
- bw
```

## Useful parametrization of bivariate contours

Previously the selection of contour level sets was delegated to the matplotlib contour algorithm, meaning they were evenly spaced densities, but essentially arbitrary.

There is a new external API for drawing contour at levels [p_0, p1, ... p_n] such that the proportion of the density lying below the ith contour level is p_i. The contours can be chosen by setting `thresh` (aka p_0) and passing an integer n to `levels`, which draws evenly spaced iso-proportion contours. Or you can pass a sequence of proportion values to `levels`.

```python
semantics = dict(x="sepal_length", y="sepal_width", hue="species")
sns.kdeplot(data=iris, **semantics, thresh=.5, levels=5)
```
image

This has the effect of obviating the `shade_lowest` parameter, because x, y values will not be shaded if they are below `thresh`. Set `thresh=0` to fill the available space with color:

```python
sns.kdeplot(data=iris, **semantics, fill=True, thresh=0, levels=100, cmap="mako")
```
image

Currently, if a colorbar is added, it will correspond to iso-densities and not iso-propotions, but I would like to improve this (possibly by dropping colorbar support and moving that information into a legend).

### API changes
```diff
+ levels, thresh
- shade_lowest
```

## Weighted kernel density estimate

A popular feature request, the refactoring and code cleaning made this very easy to add, even in the case of `hue` subsetting. It is now possible to pass weights to `weights`, which is processed with respect to `data=` like any other variable (i.e. it can be a vector of values or a name of a column in `data`).

This does require `scipy` >= 1.2, which is newer than the minimally supported version.

Closes https://github.com/mwaskom/seaborn/pull/1747
Fixes https://github.com/mwaskom/seaborn/issues/1364

### API changes

```diff
+ weights
```

## Fitting/evaluating the KDE in log space

This PR adds the ability for the KDE computation to be done in log space and then plotted correctly. The default behavior is to detect a log-scaled axis, but there is also a new parameter that sets the axis scale and then uses it during computation:

```python
x = np.random.lognormal(2, 2, size=1000)
sns.kdeplot(x=x, log_scale=True)
```
image

Fixes https://github.com/mwaskom/seaborn/issues/731

### API changes

```diff
+ log_scale
```

## Changes to color-handling in bivariate KDE plots

To ease the integration of `hue` mapping, unfilled bivariate plots default to showing all of the lines for each contour blob in the same color. It is still possible to pass a colormap object to `cmap`, but only when hue-mapping is not used. Additionally, the direction of the colormap ramp has been reversed for filled contours, to give better definition at the `thresh` level.

## Changes to the artist used to draw filled plots

Plots with multiple line and collection elements are not layered intuitively in matplotlib: all lines are drawn after all collections with the same zorder. Therefore, when ``fill=True``, *only* `fill_between` is used to draw the plot.

## Soft deprecation of `shade=` in favor of `fill=`

The `shade` parameter was not using that term properly, and the term `fill` is strictly superior. But unlike some of the other parameters affected by this rewrite, `shade` is very commonly used.

Therefore, this PR adds `fill`, and `shade` has become an alias. The old parameter does not currently issue a warning, but will likely be gradually deemphasized and then eventually formally deprecated. (There is no corresponding `fill_lowest`; see section on new bivariate API).

### API changes
```diff
+ fill
- shade
```

# Remaining tasks

- [x] Implement new bivariate `kdeplot`
- [x] Test new bivariate `kdeplot`
- [x] Refactor KDE estimation
- [x] Enhance tests for `rugplot`
- [x] Improve legend or clearly document roadmap for doing so
- [x] Implement legend for rugplot
- [x] Decision on terminology for `cut_by_hue`/`scale_by_hue`
- [x] Review and finalize other parameter names
- [x] Decision on `multiple` terminology
- [x] Decision on sticky edges for `multiple="fill"`
- [x] Set `fill=True` with `multiple="stack"` or `multiple="fill"`
- [x] Input checking on parameters that take ambiguous arguments
- [x] Change usage of `shade` to `fill` everywhere else in the library
- [x] Update `jointplot`/`pairplot`
- [x] Document inconsistencies with `violinplot`
- [x] Draft release notes
- [x] Write new docstrings
- [x] Add one or more new gallery examples
- [x] Remove old code and tests


mwaskom:
Punting decision on conceptual category for ``weights`` variable (#2111)

mwaskom:
Inconsistencies with `violinplot` noted in #2112

mwaskom:
Incomplete legend functionality noted in #2113

PR #2108, opened 2020-05-30
Add custom formatting of hue and size legend labels in `scatterplot`, `lineplot` and `relplot`

romanwerpachowski:
Sometimes the default formatting of hue or size legend labels in `scatterplot`, `lineplot` and `relplot` leaves something to be desired, for example in some cases I was getting a label "120.000000001", while it would obviously be better to have just "120". One might also want to format size labels as percentages, for example.

This PR adds two new parameters to the above plot functions:
- `hue_fmt`: New-style formatting string (e.g. '.2f') for legend tick levels of the `hue` variable when it is numeric. Not relevant if it is categorical.
- `size_fmt`: New-style formatting string (e.g. '.2f') for legend tick levels of the `size` variable when it is numeric. Not relevant if it is categorical.

I wanted to use the same way of specifying custom formatting as with the `fmt` parameter in `heatmap` function.

Unit tests are included. I also cleaned up some linter warnings (unused variables) and fixed 1 test which seemed to be [testing the wrong variable](https://github.com/mwaskom/seaborn/commit/92de1d7ded921f322365c7fe483c6939edc5ce72#diff-aa765aac3f5ebabf440c888c666886caL282) because of a copy-paste error, probably.


mwaskom:
Hi — thanks for submitting, but I think I'm -1 on this.

I realize that the relational plot legends are less flexible than would be ideal. The root problem is that matplotlib legends are more difficult to modify post-hoc than are other matplotlib artists, placing the sole burden on seaborn to offer a configuration API. Therefore, the API for legend control is something I've spent a lot of time thinking about and, while I haven't converged on a working implementation yet, these new parameters are far enough away from my current thinking that I believe they would constrain future progress.

mwaskom:
With that said, while the legend formats aren't configurable, floating point noise should have been fixed by #1726, so if you're seeing them. on the latest version a reproducible example would be helpful.

romanwerpachowski:
I understand, and thanks for the explanation. There are some unrelated test code fixes in the PR (see above), are you interested in those?

mwaskom:
A separate PR for the dud test would be helpful.

What linter are you using? Seaborn is configured to use flake8 (see the travis script and the exceptions allowed in setup.cfg) which seems to only complain about F841 when the assignment uses literal values.

e.g.

```python
f, ax = plt.subplots()
```

is fine but

```python
f, ax = "figure", "axes"
```

Is a problem.

I'm not sure if this is intended behavior or what, but FWIW I would find the cognitive overhead of remembering to type `_, ax = plt.subplots` rather high.

romanwerpachowski:
I'm using whatever the Python extension for VS Code is using. I'll send a separate PR for the dud test.

PR #2109, opened 2020-05-31
Test docs during CI

### LINKED ISSUE #2061: Add doc build to travis/github actions
This will take a little effort to make sure errors in the test build register as a failure; the current machinery might just pass through nbconvert exceptions.

mwaskom:
Closes #2061


mwaskom:
The hope here had been to split out components of the docs (gallery, notebooks docstrings) into separate github action jobs, but it doesn't seem to be possible to run setup once and then execute multiple downstream jobs in parallel. So for the time being, I will keep the doc CI as a single step using the command we'd normally use to build the docs.

Note that the doc build doesn't currently fail if there is an error in the docstring examples, although those are covered by the doctests that get run on travis.

PR #2129, opened 2020-06-09
Adding first principal component to regplot instead of y ~ x regression

Spaak:
The first PC is more appropriate to plot than either a y ~ x or x ~ y regression line when there is no clear distinction between predictor and predicted variable.

See [this blog post](https://benediktehinger.de/blog/science/scatterplots-regression-lines-and-the-first-principal-component/) and [this eLife paper](https://elifesciences.org/articles/00638) explaining why this is the case.

This PR implements this option for `regplot`. The orientation of the PC1 line is more or less arbitrary when there is no actual correlation between x and y, which can look strange, especially with confidence intervals plotted. There might be a better way to compute CIs (they should be circular in the case of zero correlation, this is now not accounted for), but I think the code more or less does what it should do.

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -318,28 +318,28 @@ class _RegressionPlotter(_LinearPlotter):
                                     seed=self.seed).T
         yhat_boots = grid.dot(beta_boots).T
         return yhat, yhat_boots
-    
+
     def fit_pca(self, grid):
         """Use the first principal component as "regression" line."""
         def reg_func(_x, _y):
-            dat = np.stack((_x[:,1],_y), axis=1)
+            dat = np.stack((_x[:, 1], _y), axis=1)
             meandat = dat.mean(0)
             dat_demean = dat - meandat
-            cov = np.dot(dat_demean.T, dat_demean) / (dat.shape[0]-1)
+            cov = np.dot(dat_demean.T, dat_demean) / (dat.shape[0] - 1)
             v, w = np.linalg.eig(cov)
-            wmax = w[:,v.argmax()] # take eigenvector with max eigval
-            
+            wmax = w[:, v.argmax()]  # take eigenvector with max eigval
+
             dat_proj = np.dot(wmax, dat_demean.T)
             # get two points on PC1 line
             a, b = meandat + dat_proj.min() * wmax
             c, d = meandat + dat_proj.max() * wmax
-            
+
             # convert from line endpoint to slope/intercept representation
             # for nice integration with bootstrapping on gridded x
-            slope = (d-b)/(c-a)
-            intercept = d-slope*c
+            slope = (d - b) / (c - a)
+            intercept = d - slope * c
             return np.asarray((intercept, slope))
-        
+
         X, y = np.c_[np.ones(len(self.x)), self.x], self.y
         grid = np.c_[np.ones(len(grid)), grid]
         yhat = grid.dot(reg_func(X, y))
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -78,9 +78,9 @@ class _RegressionPlotter(_LinearPlotter):
     def __init__(self, x, y, data=None, x_estimator=None, x_bins=None,
                  x_ci="ci", scatter=True, fit_reg=True, ci=95, n_boot=1000,
                  units=None, seed=None, order=1, logistic=False, lowess=False,
-                 robust=False, logx=False, x_partial=None, y_partial=None,
-                 truncate=False, dropna=True, x_jitter=None, y_jitter=None,
-                 color=None, label=None):
+                 robust=False, pca=False, logx=False, x_partial=None,
+                 y_partial=None, truncate=False, dropna=True, x_jitter=None,
+                 y_jitter=None, color=None, label=None):
 
         # Set member attributes
         self.x_estimator = x_estimator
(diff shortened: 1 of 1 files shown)


Spaak:
Perhaps the ellipsoid described by PC1 and PC2 is a better "CI" than the typical line-based CI. I'll try to get that working, see how it turns out, and push to this PR if appropriate. (So please keep this as work in progress for now, don't merge yet, in case you were considering it ;) )

mwaskom:
What is the difference between drawing the first PC and doing [orthogonal distance regression](https://docs.scipy.org/doc/scipy/reference/odr.html)?

I am open to this feature in general but am -1 on adding any more modes to `regplot` in this fashion. In fact, `regplot` is on deck for a complete revamp to its API, which is limiting in various ways.

Spaak:
> What is the difference between drawing the first PC and doing [orthogonal distance regression](https://docs.scipy.org/doc/scipy/reference/odr.html)?

I hadn't heard of orthogonal distance regression, but based on a cursory glance (and the name!) it seems very similar and related. The first PC is specifically a line that minimizes residuals orthogonal to it, as opposed to ordinary regression which minimizes residuals parallel to the y-axis (or x-axis, if you flip the variables). (See Fig 2 in the [blog post](https://benediktehinger.de/blog/science/scatterplots-regression-lines-and-the-first-principal-component/) I linked for the different residuals being minimized.)

> I am open to this feature in general but am -1 on adding any more modes to `regplot` in this fashion. In fact, `regplot` is on deck for a complete revamp to its API, which is limiting in various ways.

Good to hear that you're open in principle. I fully agree that `regplot` is due a revamp, all these various mutually exclusive regression options are a clear indication that API improvements can be made. Should I wait for (/optionally contribute to) the revamp, and take it from there? And does it make sense to keep this PR open in the meanwhile (in case I make further changes to the functionality)? If not, feel free to close.

mwaskom:
Yes to be clear I know what PCA is, I've just never looked into the details of ODR (it comes from a more engineering tradition rather than statistics) but had always assumed that this is what it was basically doing.

I mention it because seaborn tries to offload computation to numpy/scipy/statsmodels wherever possible.

mwaskom:
We can leave this PR open. I expect that this functionality, if included in seaborn, would go in a different function that shares infrastructure with `regplot`, especially it uses a very different visual representation of uncertainty, which would take a lot of thinking (e.g., if we have elliptical CIs does it make sense only to use them in the context of PCA regression? Won't people expect that they can draw a scatterplot with ellipses but no lines? This is an example of how apparently simple new features can have repercussions throughout the library if not carefully considered).

Spaak:
> Yes to be clear I know what PCA is

I wasn't doubting that 😄 I emphasized the different types of residuals being minimized, which to me suggested that ODR was very similar.

> I mention it because seaborn tries to offload computation to numpy/scipy/statsmodels wherever possible.

I actually considered implementing the actual PCA algorithm using sklearn, but I did a check and saw that sklearn is now not a dependency of seaborn. Since PCA is extremely simple, I chose to implement using vanilla numpy instead (`numpy.linalg.eig` on the data covariance matrix, extremely fast). Would you prefer to offload the PCA to sklearn? Or perhaps statsmodels might be a good choice actually, I now see they also offer a PCA interface.

> I expect that this functionality, if included in seaborn, would go in a different function that shares infrastructure with regplot,

Yes, makes sense.

> if we have elliptical CIs does it make sense only to use them in the context of PCA regression? Won't people expect that they can draw a scatterplot with ellipses but no lines?

I think the case here is relatively analogous to what's now implemented for `logistic`. The funky CI I'm getting now is a consequence of recoding the PC line in slope/intercept form, like it's done now for the standard linear regression.

> This is an example of how apparently simple new features can have repercussions throughout the library if not carefully considered

Yes, I see that. I'll work a bit more on the CIs, and hopefully something useful will come out of it.

mwaskom:
I don't think seaborn should add a scikit-learn dep just for this. I do think it's worth investigating whether the scipy ODR module can be used to produce the same effect and, if so, whether it makes sense to use here rather than implementing PCA in seaborn.

mwaskom:
With the new objects interface, it becomes much easier to add functionality like this to seaborn. Here's a very basic adaptation of your code as a stat transform:

```python
class PCFit(so.Stat):
    def _fit(self, data):

        dat = data[["x", "y"]].to_numpy()
        meandat = dat.mean(0)
        dat_demean = dat - meandat
        cov = np.dot(dat_demean.T, dat_demean) / (dat.shape[0] - 1)
        v, w = np.linalg.eig(cov)
        wmax = w[:, v.argmax()]  # take eigenvector with max eigval
        dat_proj = np.dot(wmax, dat_demean.T)

        return pd.DataFrame([
            meandat + dat_proj.min() * wmax,
            meandat + dat_proj.max() * wmax,
        ], columns=["x", "y"])

    def __call__(self, data, groupby, *args, **kws):
        return groupby.apply(data, self._fit)
```

Now we can simply pass this object like any other stat:

```python
x, y = np.random.multivariate_normal([1, 3], [(1, .5), (.5, 1)], size=300).T
(
    so.Plot(x, y)
    .add(so.Dots())
    .add(so.Line(color="C2"), so.PolyFit(1))
    .add(so.Line(color="C1"), PCFit())
)
```


and because you can just plug it right in, it can be composed with all of the other features:

```python
(
    so.Plot(penguins, "bill_length_mm", "bill_depth_mm", color="species")
    .add(so.Dots())
    .add(so.Line(), PCFit())
)
```


So I think this is _probably_ something that will be added in the future and definitely something that could get incorporated as a third-party object either way. I'm not quite yet advertising / documenting the third-party plugin functionality as I think there's a few more aspects of the internal API that need to stabilize. But I do think it sufficiently moots this PR as the way forward on incorporating this functionality into seaborn, so I'm going to close it for now.

PR #2140, opened 2020-06-15
Clarified documentation of despine

e-dorigatti:
When `fig` and `ax` are both specified, `ax` is ignored and all axes in `fig` are despined. Took me a while to figure out, so I clarified the documentation.

I think in this case it would be better to despine `ax` and ignore `fig`, but I realize this is undesirable as it might break some code out there.

I also thought of adding an explicit warning, would this be excessive? For example:

```
    elif fig is not None:
        axes = fig.axes
        if ax is not None:
            warnings.warn('ax parameter ignored when fig is specified')
```

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -219,9 +219,10 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
     """Remove the top and right spines from plot(s).
 
     fig : matplotlib figure, optional
-        Figure to despine all axes of, default uses current figure.
+        Figure to despine all axes of, defaults to current figure.
     ax : matplotlib axes, optional
-        Specific axes object to despine.
+        Specific axes object to despine when a figure is not
+        specified, ignored otherwise.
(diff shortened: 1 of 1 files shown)


seaborn/utils.py
```diff
@@ -219,9 +219,10 @@ def despine(fig=None, ax=None, top=True, right=True, left=False,
     """Remove the top and right spines from plot(s).
 
     fig : matplotlib figure, optional
-        Figure to despine all axes of, default uses current figure.
+        Figure to despine all axes of, defaults to current figure.
     ax : matplotlib axes, optional
-        Specific axes object to despine.
+        Specific axes object to despine when a figure is not
```
mwaskom, line 224: I think this would be clearer as 

> Specific axes object to despine. Ignored if ``fig`` is provided.

e-dorigatti: Thank you for the suggestion, I updated the wording accordingly.

mwaskom:
Thanks @e-dorigatti

PR #2141, opened 2020-06-15
Add ecdfplot function

mwaskom:
This PR adds a new function in the distributions module, `ecdfplot`.

Simple example:

```python
sns.ecdfplot(diamonds, x="price", hue="cut")
```
ecdf

There's some motivation for it in #1536 
There was an initial attempt in #1621, but the overhaul to the distributions module made it unfortunately impossible to pull those commits into the history

There's really not too much to the implementation.

We could add more semantics than just `hue` fairly easily, but because the other distribution plots don't support them well, I have deferred for the moment. They can always be added later.

We could also do a bivariate ECDF plot. But I am not sure this is very useful. The main advantage of the ECDF over histogram/kde is that it lends itself better to comparing distributions. But that is less true in the bivariate case. Still nothing here blocks it from being added later.

## To do:

- [x] Unit tests for `ECDF` estimator
- [x] Unit tests for `ecdfplot` function
- [x] Docstring content
- [x] API examples
- [x] Release notes

### DIFF
--- a/seaborn/_statistics.py
+++ b/seaborn/_statistics.py
@@ -1,3 +1,29 @@
+"""Statistical transformations for visualization.
+
+This module is currently private, but is being written to eventually form part
+of the public API.
+
+The classes should behave roughly in the style of scikit-learn.
+
+- All data-independent parameters should be passed to the class constructor.
+- Each class should impelment a default transformation that is exposed through
+  __call__. These are currently written for vector arguements, but I think
+  consuming a whole `plot_data` DataFrame and return it with transformed
+  variables would make more sense.
+- Some class have data-dependent preprocessing that should be cached and used
+  multiple times (think defining histogram bins off all data and then counting
+  observations within each bin multiple times per data subsets). These currently
+  have unique names, but it would be good to have a common name. Not quite
+  `fit`, but something similar.
+- Alternatively, the transform interface could take some information about grouping
+  variables and do a groupby internally.
+- Some classes should define alternate transforms that might make the most sense
+  with a different function. For example, KDE usually evaluates the distribution
+  on a regular grid, but it would be useful for it to transform at the actual
+  datapoints. Then again, this could be controlled by a parameter at  the time of
+  class instantiation.
+
+"""
 from distutils.version import LooseVersion
 from numbers import Number
 import numpy as np
@@ -348,9 +374,17 @@ class Histogram:
 
 
 class ECDF:
-
+    """Univariate empirical cumulative distribution estimator."""
     def __init__(self, stat="proportion"):
+        """Initialize the class with its paramters
 
+        Parameters
+        ----------
+        stat : {{"proportion", "count"}}
+            Distribution statistic to compute.
+
+        """
+        # TODO add remove_duplicates
         _check_argument("stat", ["count", "proportion"], stat)
         self.stat = stat
 
--- a/seaborn/_statistics.py
+++ b/seaborn/_statistics.py
@@ -375,18 +375,20 @@ class Histogram:
 
 class ECDF:
     """Univariate empirical cumulative distribution estimator."""
-    def __init__(self, stat="proportion"):
+    def __init__(self, stat="proportion", complementary=False):
         """Initialize the class with its paramters
 
         Parameters
         ----------
         stat : {{"proportion", "count"}}
             Distribution statistic to compute.
+        complementary : bool
+            If True, use the complementary CDF (1 - CDF)
 
         """
-        # TODO add remove_duplicates
         _check_argument("stat", ["count", "proportion"], stat)
         self.stat = stat
+        self.complementary = complementary
 
     # Do we need bivariate ECDF?
 
@@ -395,7 +397,6 @@ class ECDF:
         sorter = x.argsort()
         x = x[sorter]
         weights = weights[sorter]
-
         y = weights.cumsum()
 
         if self.stat == "proportion":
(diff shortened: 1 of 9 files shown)


mwaskom:
Note to self: write a test against the statsmodels implementation, but protect it under a skipif decorator for builds without statsmodels.

lmcinnes:
Thanks for getting to this. I do believe having it easily available as a default in Seaborn will make a notable difference in helping ensure it gets used more often (as there are certainly many cases where it is one of the better options, particularly in comparing distributions).

It looks as if the distributions overhaul made this easier to just plug in -- sorry my early effort wasn't more help. Mostly just glad to see it included.

ericmjl:
@mwaskom thank you for this! I hope you have been taking care of yourself this while. Just wanted to echo @lmcinnes' comment, as convenience tooling will definitely help with encouraging best practices. I will be sure to celebrate this PR in a blog post, tweet it out, and share it via my newsletter once merged.

Meanwhile, I will go in now and see if I can help with anything via a review (if I'm able to leave a review comment, that is).

seaborn/_statistics.py
```diff
@@ -345,3 +345,39 @@ def __call__(self, x1, x2=None, weights=None):
...
+
+
+class ECDF:
+
+    def __init__(self, stat="proportion"):
+
+        _check_argument("stat", ["count", "proportion"], stat)
+        self.stat = stat
+
+    # Do we need bivariate ECDF?
```
ericmjl, line 357: I think the bivariate visualization will be difficult to explain, so I think a univariate is good enough :smile:.

seaborn/_statistics.py
```diff
@@ -345,3 +345,39 @@ def __call__(self, x1, x2=None, weights=None):
...
+
+    def _eval_univariate(self, x, weights):
```
ericmjl, line 359: Might a docstring help future contributors to the library?

I thought of the following as a strawman, though I am merely deciphering the implementation intent from the code:

```suggestion
    def _eval_univariate(self, x, weights):
        """
        Return x, y for plotting ECDF.
        
        X-values are the data values,
        while y-values are either the cumulative proportion (if self.stat is "proportion")
        or the cumulative number (if self.stat is "count").
        of values equal to or below that value,

        """
```

seaborn/distributions.py
```diff
@@ -1770,6 +1827,57 @@ def kdeplot(
 )
 
 
+def ecdfplot(
+    data=None, *,
+    # Vector variables
+    x=None, y=None, hue=None, weights=None,
```
ericmjl, line 1833: Should `y` be left in, given that `y` is not used below?

seaborn/distributions.py
```diff
@@ -1770,6 +1827,57 @@ def kdeplot(
...
+    # Computation parameters
+    stat="proportion",
+    # Hue mapping parameters
+    palette=None, hue_order=None, hue_norm=None, color=None,
+    # Axes information
+    log_scale=None, legend=True, ax=None,
+    # Other appearance keywords
+    **kwargs,
+):
```
ericmjl, line 1842: Suggested docstring:

```suggestion
):
    Plot ECDF of data.
    
    :param data: Tidy (“long-form”) dataframe where each column is a variable and each row is an observation.
    :param x: Column name of variable to plot ECDF for.
    :param hue: Grouping variable that will produce elements with different colors. Can be either categorical or numeric, although color mapping will behave differently in latter case.
    :param weights: Column that houses weighting for each sample.
    :param stat: Whether to plot proportion or raw counts. Defaults to "proportion".
```

Apologies, I have to drop off for a bit, so my review comment ends here.

seaborn/distributions.py
mwaskom: `y` is in the signature because you can draw a plot with the variable on the y axis (and the proportion/count on the y axis) by assigning to `y`.

seaborn/distributions.py
mwaskom: Thanks — yeah, filling in the docstrings etc. is on the TODO, just wanted to get the PR started. The seaborn docs use a kind of convoluted templating system for maintaining consistency across parameters in the shared API, so most of the doc are actually written, they just need to be stuck together.

anntzer:
I like the idea (well, I'm trying to get the same thing into matplotlib itself :-)).  One suggestion would be to also support complementary cumulative distributions (ccdfs) -- they can be useful e.g. in log scale when looking at distributions with exponential tails to the right.

mwaskom:
Ah yes — great suggestion!

anntzer:
Also, while we're at it, I have also found it useful to have a flag to draw the ec(c)df with swapped axes (i.e., the variable along y and the probabilities along x).

mwaskom:
Yeah, that works here. In seaborn it's not a flag, you just assign to the `y` variable:

```python
sns.ecdfplot(diamonds, y="price", hue="cut")
```

mwaskom:
@anntzer the eccdf should just be mirrored in y, right?

```python
obs = np.arange(10)
sns.ecdfplot(obs)
sns.ecdfplot(obs, complementary=True)
```
image

anntzer:
re: eccdf: yes.
re: y: that's nice (I have to admit I don't actually use seaborn very often, so I'm not too aware of its conventions, here this certainly works out well), but the plot in y looks slightly off (likely off-by-1 error)?  Compare
```
sns.ecdfplot(x=[1, 2, 3])
```
x
and
```
sns.ecdfplot(y=[1, 2, 3])
```
y
Clearly the plot in y is not the transpose of the plot in x; in fact the plot in y doesn't even reach proportion=0.

mwaskom:
Oh, good catch. Actually I don't know the answer to this: how do you get `plot` to step horizontally, not vertically?

mwaskom:
Fixed:

```python
sns.ecdfplot(y=[1, 2, 3])
```

image

PR #2144, opened 2020-06-20
MNT: fix some failures with matplotlib=3.3.0rc1

MaozGelbart:
Matplotlib 3.3.0rc1 has been released and results with several failing tests. This PR addresses some of them:

* FixedLocator now requires that `set_ticklabels()` inputs are equal to the number of ticks ([https://github.com/matplotlib/matplotlib/pull/17266](https://github.com/matplotlib/matplotlib/pull/17266)). This is an issue for `palplot` as well as categorical plots using unnamed groups, and fails several smoke tests and documentation examples. The proposed changes pass empty ticklabels in such cases (e.g. `[""]` instead of `[]`).

* `ax.get_aspect()` now returns `1` instead of `"equal"` when aspect was set to `"equal"` ([https://github.com/matplotlib/matplotlib/pull/16012](https://github.com/matplotlib/matplotlib/pull/16012)). "equal" aspect input is still valid. This affects matrix tests. The proposed change involves testing for both `"equal"` and `"1"` aspect values. 

Two remaining test failures are under investigation. While still unclear, these may not require seaborn code changes so I didn't include them here.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -357,12 +357,16 @@ class _CategoricalPlotter(object):
         if ylabel is not None:
             ax.set_ylabel(ylabel)
 
+        group_names = self.group_names
+        if not group_names:
+            group_names = ["" for _ in range(len(self.plot_data))]
+
         if self.orient == "v":
             ax.set_xticks(np.arange(len(self.plot_data)))
-            ax.set_xticklabels(self.group_names)
+            ax.set_xticklabels(group_names)
         else:
             ax.set_yticks(np.arange(len(self.plot_data)))
-            ax.set_yticklabels(self.group_names)
+            ax.set_yticklabels(group_names)
 
         if self.orient == "v":
             ax.xaxis.grid(False)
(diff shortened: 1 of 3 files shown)


mwaskom:
Thanks!

> This is an issue for palplot as well as categorical plots using unnamed groups, and fails several smoke tests and documentation examples. 

Is this likely to break a lot of existing code?

MaozGelbart:
`plt.boxplot` uses `FixedLocator` as its locator, for example. `set_ticklabels([])` is a popular way to remove tick labels (see [this stackoverflow thread](https://stackoverflow.com/questions/2176424/hiding-axis-text-in-matplotlib-plots)). I guess the absence of a deprecation period on this change is likely to break existing code, apart from breaking seaborn.

mwaskom:
Hm, is it worth kicking up a request for a deprecation cycle to matplotlib?

MaozGelbart:
Good idea, I'll open an issue on this.

mwaskom:
Thanks! Still good to have these changes of course.

PR #2146, opened 2020-06-23
TST: Increase testing speed by utilizing multiple cores

MaozGelbart:
pytest is capable of running multiple CPUs in parallel if `pytest-xdist` plugin is installed. This should improve tests speed on multi-core systems. Each Travis-CI process has 2 CPUs available and currently only one is utilized. This PR introduces multi-core testing, hopefully improving local and CI tests speed. On my 6-core system, the full test suit is completed in less than 60 seconds, compared to ~330 seconds using a single core.

To overcome race conditions on datasets pulls, I added a script to cache test datasets before actual running test suits. 

I also removed unnecessary qt pulls from non-interactive travis builds; this reduced ~1m from such travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
   - source activate testenv
   # Conda's matplotlib package pulls qt but this is not required for
   # non-interactive tests; use the compact matplotlib-base instead
-  - if [ $MPLBACKEND == "qtagg" ];
+  - if [ $MPLBACKEND == "qt5agg" ];
     then cat ci/deps_${DEPS}.txt ci/utils.txt > deps.txt;
     else cat ci/deps_${DEPS}.txt ci/utils.txt | sed 's/matplotlib$/matplotlib-base/' > deps.txt;
     fi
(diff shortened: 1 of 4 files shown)


mwaskom:
Cool — looks like it's hitting some a race condition when the doctests try and cache the example data.

MaozGelbart:
Thanks! So the longer Travis CI tests are only improved by ~10%-15% mainly since long setup time cannot be parallelized. I'll see how I get cached files creation multithread-safe.

MaozGelbart:
I think this is ready, up to lines where coverage decreased (cache dir creation and retrieval of test dataset via url request).

.travis.yml
```diff
@@ -32,17 +32,18 @@ before_install:
...
   - pip install .
 
 
 before_script:
-  - if [ $MPLBACKEND == "qtagg" ]; then
-    export DISPLAY=:99.0;
-    sh -e /etc/init.d/xvfb start;
-    sleep 3;
-    fi
+  - python ci/cache_test_datasets.py
```
mwaskom, line 46: I may be missing something but how does this line replace the xvfb setup?

MaozGelbart: It is declared earlier as a service:
https://github.com/mwaskom/seaborn/blob/eb690eca5ced20de231c0bdc8ed5901b7e0e5f83/.travis.yml#L5-L6

mwaskom: Oh cool.

PR #2151, opened 2020-07-01
DOC: enable Sphinx rules for linking to GitHub issues, PRs and users

### LINKED ISSUE #2062: Add sphinx rule for linking to relevant github issues in release notes
Look at what scikit learn does here.

This will be helpful for providing both context and credit.

MaozGelbart:
This PR adds the ability to easily link to GitHub issues, PRs and users (e.g. for release notes), to easily provide broader context and credit. This is achieved by adding `sphinx-issues` extension (as done in scikit-learn, see [here](https://github.com/sloria/sphinx-issues)). Rules examples:
- ``:issue:`2062` `` links to #2062 
- ``:issue:`2052,2081` `` will link to both issues
- ``:pr:`2151` `` links to PR #2151
- ``:user:`MaozGelbart` `` links to user @MaozGelbart 

It can also link to other repos: ``:issue:`user/repo#12345` `` 

This PR also removes what seems like an unnecessary matplotlib import from sphinx conf file ('agg' is already the default backend in modern matplotlibs)

Closes #2062

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -13,8 +13,6 @@
 
 import sys, os
 import sphinx_bootstrap_theme
-import matplotlib as mpl
-mpl.use("Agg")
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -39,7 +39,11 @@ extensions = [
     'matplotlib.sphinxext.plot_directive',
     'gallery_generator',
     'numpydoc',
+    'sphinx_issues',
 ]
+
+# Sphinx-issues configuration
+issues_github_path = 'mwaskom/seaborn'
 
 # Generate the API documentation when building
 autosummary_generate = True
--- a/doc/releases/v0.11.0.txt
+++ b/doc/releases/v0.11.0.txt
@@ -2,14 +2,14 @@
 v0.11.0 (Unreleased)
 --------------------
 
-- TODO stub for explaining improvements to variable specificiation. Make this good!  GH2017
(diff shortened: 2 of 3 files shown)


mwaskom:
> ('agg' is already the default backend in modern matplotlibs)

That might be true on linux systems, but in distributions that target macos, it defaults to the MacOSX backend, which has been flaky in the past.

mwaskom:
Otherwise looks good. Would you be able to replace the existing placeholders in the v011.0 release notes? And then I think there's an open issue about enabling this that could be linked and closed.

MaozGelbart:
> Would you be able to replace the existing placeholders in the v011.0 release notes?

Done!



> That might be true on linux systems, but in distributions that target macos, it defaults to the MacOSX backend, which has been flaky in the past.

If I understand correctly, the backend that renders png files by default is `'agg'` even if the active backend is MacOSX. See [here](https://matplotlib.org/3.3.0/tutorials/introductory/usage.html#the-builtin-backends). Anyway this change is not required for this PR so the problematic commit could be ignored or reverted. Do you prefer that I revert or restore the `mpl.use` lines?

mwaskom:
I think you're right that *saving* a png uses the Agg backend by default. The question is about what happens when the figure needs to be drawn during plotting (e.g. for `tight_layout`). I think, on a mac, that's going to happen with the macosx backend code now.

That said, most of the docs are not created using the inline notebook backend (and this will become even more true as API docs are ported over to that system). In the meantime if it turns out to be an issue, a better solution would be to use the `MPLBACKEND` environment variable anyway.

Otherwise, getting the github issue linking up and running is great, thank you!

PR #2155, opened 2020-07-03
adding option for parametric (t-based) confidence intervals

Spaak:
In some common cases (e.g. for comparison with other statistics and plots) it is desirable to display parametric confidence intervals, typically based on a T-distribution. Usually, these will be very similar to the bootstrapped CI, but I think it's good to have the option for exact parametric intervals as well. This PR adds this functionality.

Currently there is no option for `ci_estimator` or so, and I thought implementing such a thing would be too big an API overhaul without discussing. I think the following is a lightweight enough option that it might be useful to many:

```python

sns.pointplot(data, ci=("parametric", 95))
```

for a parametric CI. This is how the current PR exposes the functionality.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1483,8 +1483,9 @@ class _CategoricalStatPlotter(_CategoricalPlotter):
 
                     elif type(ci) == tuple and ci[0] == "parametric":
 
-                        confint.append(stats.t.interval(ci[1]/100, len(stat_data)-1,
-                            loc=estimator(stat_data), scale=stats.sem(stat_data)))
+                        confint.append(stats.t.interval(ci[1] / 100, len(stat_data) - 1,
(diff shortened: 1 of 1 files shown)


mwaskom:
Hi, implementing something like this is on the roadmap, but 1) it needs to be implemented everywhere you can plot confidence intervals in seaborn and 2) this is not the planned API.

Spaak:
> Hi, implementing something like this is on the roadmap, but 1) it needs to be implemented everywhere you can plot confidence intervals in seaborn and 2) this is not the planned API.

Thanks for the quick feedback. I agree, a different API makes sense; for now I was trying to stick to the API that was already present (e.g. the ad-hoc option for "sd").

Let me know if there's anything I can contribute! (Not necessarily limited to new functionality, also happy to work on refactoring API etc.)

PR #2160, opened 2020-07-09
Improve seaborn discoverability

MaozGelbart:
This PR updates the pypi trove classifiers list to improve package discoverability. This includes adding the new matplotlib framework classifier and OS-independent classifier (since no OS-specific code or extension is used directly by seaborn).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,10 @@ before_install:
 install:
   - conda create -n testenv pip python=$PYTHON
   - source activate testenv
-  # Conda's matplotlib package pulls qt but this is not required for
-  # non-interactive tests; use the compact matplotlib-base instead
+  # Add pyqt for interactive tests since it doesn't always listed as a dependency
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks! I had been meaning to add this.

Looks like there is an idiosyncratic failure in your dataset cache script, any idea what's going on there?

MaozGelbart:
> Looks like there is an idiosyncratic failure in your dataset cache script, any idea what's going on there?

Seems like the fresh matplotlib=3.2.2 builds for linux on conda-forge do not contain qt as a dependency on linux systems (as opposed to previous builds). See [here](https://github.com/conda-forge/matplotlib-feedstock/issues/253). Not sure if this is intentional or not, waiting for a response from that repo owners.

MaozGelbart:
Travis is green now. It required listing pyqt as an explicit dependency for the interactive tests. This also exercised matplotlib==3.3.0 and no failures were observed.

PR #2193, opened 2020-08-12
Handle matplotlib==3.3 deprecations

MaozGelbart:
* Adapts to a recent change in matplotlib where `basex/basey` parameters of `axis.set_scale` are now just `base`.
* Adapts to another change affecting `heatmap` where passing `norm` to `ax.pcolormesh` accompanied with `vmin/vmax` is deprecated. `norm` is not a `heatmap` parameter, so this PR avoids setting `vmin/vmax` if `norm` is given as a kwarg.

### DIFF
--- a/seaborn/_core.py
+++ b/seaborn/_core.py
@@ -5,6 +5,7 @@ from functools import partial
 from collections.abc import Iterable, Sequence, Mapping
 from numbers import Number
 from datetime import datetime
+from distutils.version import LooseVersion
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


mwaskom:
It's curious that codecov thinks the mpl < 3.3 branch is not covered, since it should be tested by the "pinned" build.

MaozGelbart:
codecov on the "pinned" tests instance did not complete for some reason:
```
Error: No coverage report found
```
so the "pinned" tests coverage report was not uploaded.

MaozGelbart:
Well, I see that this phenomena is shared across many PRs: the first "pinned" job codecov command errors out and doesn't complete.

mwaskom:
Ah! Good sleuthing.

PR #2196, opened 2020-08-16
FIX: allow un-sharing categories in catplot

### LINKED ISSUE #1702: Can't use catplot with melt for different categories
I would have expected that I can visualize different categorical variables in a count plot using ``catplot`` and ``melt``.
However, it looks like the data is converted to categorical before grouping for the columns, and so the categories are shared among all the count plots. That doesn't really make sense if the different columns correspond to different categories.
Am I overlooking something or is there a different way of doing this?

```python
import pandas as pd
import seaborn as sns
ames = pd.read_excel("http://www.amstat.org/publications/jse/v19n3/decock/AmesHousing.xls")

cat_cols = ['MS Zoning', 'Street', 'Alley', 'Lot Shape', 'Land Contour',
       'Utilities', 'Lot Config', 'Land Slope', 'Neighborhood', 'Condition 1',
       'Condition 2', 'Bldg Type', 'House Style', 'Roof Style', 'Roof Matl',
       'Exterior 1st', 'Exterior 2nd', 'Mas Vnr Type', 'Exter Qual',
       'Exter Cond', 'Foundation', 'Bsmt Qual', 'Bsmt Cond', 'Bsmt Exposure',
       'BsmtFin Type 1', 'BsmtFin Type 2', 'Heating', 'Heating QC',
       'Central Air', 'Electrical', 'Kitchen Qual', 'Functional',
       'Fireplace Qu', 'Garage Type', 'Garage Finish', 'Garage Qual',
       'Garage Cond', 'Paved Drive', 'Pool QC', 'Fence', 'Misc Feature',
       'Sale Type', 'Sale Condition']
ames_cat = ames[cat_cols]
sns.catplot(x='value', col='variable', data=ames_cat.melt(), sharex=False, sharey=False, col_wrap=5, kind='count')
```

MaozGelbart:
Fixes #1702. With this PR, the same categorical axis (and order) for all facets is only enforced by `catplot` if `sharex` (or `sharey`) are not `False` (on master, in any case the categorical axis is shared between facets). Note that `sharex` and `sharey` can also accept `"row"` or `"col"` but this use case is not handled within this PR.


seaborn/categorical.py
```diff
@@ -3889,14 +3891,15 @@ def catplot(
...
     {col_wrap}
     {stat_api_params}
-    {order_vars}
+    {order_vars} ``order`` is ignored if the ``share`` parameter for the
+        categorical axis is ``False``.
     row_order, col_order : lists of strings, optional
         Order to organize the rows and/or columns of the grid in, otherwise the
         orders are inferred from the data objects.
-    kind : string, optional
+    kind : {{"strip", "swarm", "box", "violin", "boxen", "point", "bar", \
```
mwaskom, line 3899: Given that the enumeration of possible values isn't used, I think I'd rather have them in the description if there are too many to fit on one line.

Two related comments:
- I'm gradually removing "optional" from the docstrings, as basically every parameter of the plotting functions is optional and it just adds noise.
- The pyflakes config now allows lines up to 88 characters, per the Black recommendations, so there might be space

MaozGelbart: Thanks! Moved the possible values back to the description.
Thanks for the comments, will use them.

mwaskom:
Of note, I've been looking into how to hook into the matplotlib categorical units machinery (as our oldest-supported matplotlib version now has them), and just encountered a potential related limitation on that end: https://github.com/matplotlib/matplotlib/issues/18273

MaozGelbart:
Cool, thanks for sharing!

seaborn/categorical.py
```diff
@@ -3889,14 +3891,15 @@ def catplot(
         Categorical variables that will determine the faceting of the grid.
     {col_wrap}
     {stat_api_params}
-    {order_vars}
+    {order_vars} ``order`` is ignored if the ``share`` parameter for the
```
mwaskom, line 3894: This doesn't work, unfortunately. It renders as:

```
order, hue_order : lists of strings, optional
    Order to plot the categorical levels in, otherwise the levels are
    inferred from the data objects.         ``order`` is ignored if the ``share`` parameter for the
    categorical axis is ``False``.
```

mwaskom:
I think `catplot` should always honor `order` if passed. The change to handle `share{x,y}=False` should apply only when `order=None`. Granted, it doesn't make much sense to pass both, but `sharex` is really a matplotlib kwarg, and seaborn should still supersede it by ordering the categorical axes if you ask it to.

Also, that would be consistent with what happens when you pass a variable with explicit category ordering, which still shows empty categories with this change.

mwaskom:
I also don't love how you end up with different variables sharing a color:

```python
sns.catplot(data=penguins, x="species", col="island", kind="count", sharex=False)
```
image

I have work on the categorical plots planned for 0.12. Among those plans are a change to how colors are chosen in categorical plots that will require an explicit `hue` mapping to see multiple colors. I think I'm inclined to punt on this change until then.

seaborn/categorical.py
MaozGelbart: Although obsolete code now, this construct rendered ok in the result HTML output (there wasn't a gap there).

MaozGelbart:
> I think `catplot` should always honor `order` if passed. The change to handle `share{x,y}=False` should apply only when `order=None`. Granted, it doesn't make much sense to pass both, but `sharex` is really a matplotlib kwarg, and seaborn should still supersede it by ordering the categorical axes if you ask it to.

Done. This indeed makes more sense.



> I also don't love how you end up with different values sharing a color

Yes the result plot is misleading. Let's wait for 0.12 then (although we can emit a warning for such cases until the 0.12 change is made, if the un-share functionality is of interest sooner).

mwaskom:
I think in most cases where someone uses "unshared" facets, they really want to be using `PairGrid`. (Indeed, in the original example, the data need to go through a convoluted melt just to get to the point where they need this feature).

The main reason you wouldn't use `PairGrid` is if you have a lot of variables and want to be using `col_wrap`.  I think `PairGrid` should support something like `col_wrap`, although it will take a little bit of thinking about the API.

MaozGelbart:
> I think in most cases where someone uses "unshared" facets, they really want to be using `PairGrid`. (Indeed, in the original example, the data need to go through a convoluted melt just to get to the point where they need this feature).

My personal experience- this happened to me when the categories (which were just levels of some variable) were mutually exclusive between facets in the first place, intentionally (this was the first time I dived into seaborn's codebase).

mwaskom:
OK, nested variables is a good point. I think issuing a warning (when ``color=None``) along the lines of (needs adaptation based on plot orientation)

> "Setting `sharex=False` with `color=None` will cause different levels of the ``x`` variable to share colors. This will change in a future version."

is a good suggestion, and then this can go in for 0.11.

mwaskom:
Please add an entry in the release notes, also.

MaozGelbart:
Done, thanks! (although, this warning will raise also if `x` is categorical and the axis share categories anyway. do you know a simple way to test for `x` being categorical using a `_CategoricalPlotter` method or property?)

mwaskom:
Do you mean if the variable is a pandas `Categorical`?

mwaskom:
If so, I'm not sure, given that `catplot` (for now) only accepts `DataFrame` input, you have unambiguous access to that input variable and so can quickly check `hasattr(data[x], "cat")`, I think. But I'm inclined to say that the intersection of `Categorical` data and `sharex=False` is likely small enough that it's not worth adding a lot of special code to avoid a false-alarm warning.

mwaskom:
It also occurred to me that the warning is a false-alarm when `hue` is assigned as well. But again, it just says "may cause", so — while warnings are often annoying — it's not technically wrong.

MaozGelbart:
The PR is ready, and I've fixed the failing jointplot example as well. 

> But I'm inclined to say that the intersection of `Categorical` data and `sharex=False` is likely small enough that it's not worth adding a lot of special code to avoid a false-alarm warning.

Agreed.

> It also occurred to me that the warning is a false-alarm when `hue` is assigned as well.

Good point, I've changed to code to warn only if there's no hue.

Thanks for all the help!

mwaskom:
I feel like I don't understand how that `jointplot` example didn't fail recent PRs.

MaozGelbart:
This is possibly related to [https://github.com/mwaskom/seaborn/pull/2069/commits](https://github.com/mwaskom/seaborn/pull/2069/commits) ? The code there has been created before the distribution plots overhaul was available on master.

mwaskom:
Oh yes ... I missed that you added `rug=True` in one of the `jointplot` examples but I didn't rebase before merging. OK, that's a relief: it's not a problem with the test infra itself.

(I feel like travis used to run a build after a PR was merged so that it could notify if master breaks when that happens, but maybe I am misremembering).

Mentioned that this deprecation needs smoother handling here: #2198

MaozGelbart:
> (I feel like travis used to run a build after a PR was merged so that it could notify if master breaks when that happens, but maybe I am misremembering).

It did, see [here](https://travis-ci.org/github/mwaskom/seaborn/builds/719451894).

mwaskom:
Oh I guess it sent you the notification, not me?

MaozGelbart:
No, I don't receive travis notifications at all.

mwaskom:
`¯\_(ツ)_/¯`

PR #2219, opened 2020-08-25
Update err_style parameter documentation.

fhaase2:


mwaskom:
Hm, while accurate, I don't think this is a good change, because the preferred way to disable the confidence intervals is setting `ci=None`. If you just set `err_style` to `None`, it won't show them, but it will still compute them. I'm not sure there's a reason anyone would want that?

I don't recall off the top of my head why `None` is allowed, but probably best to not advertise that (and perhaps deprecated it).

fhaase2:
Ah i see. Thanks for the explanation. I guess its best to deprecate the `None` case.

PR #2236, opened 2020-08-31
MAINT: avoid deprecation warnings with future numpy releases

MaozGelbart:
The next minor numpy version (1.20) [deprecated](https://github.com/numpy/numpy/pull/14882) the usage of numpy types that are just aliases to python builtin types. This includes `np.bool`, `np.int`, `np.float` and others. This PR replaces these occurrences with the native types to avoid deprecation warnings on future numpy versions. I changed `np.object` to `np.object_` and not the builtin type since it resulted with slight differences in the relevant test suit (`test_matrix.py`).

### DIFF
--- a/examples/many_pairwise_correlations.py
+++ b/examples/many_pairwise_correlations.py
@@ -21,7 +21,7 @@ d = pd.DataFrame(data=rs.normal(size=(100, 26)),
 corr = d.corr()
 
 # Generate a mask for the upper triangle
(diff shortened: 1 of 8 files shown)


MaozGelbart:
Obtaining wheels for nightly builds of `numpy==1.20.0dev0` may be done with
```
dev_anaconda_url=https://pypi.anaconda.org/scipy-wheels-nightly/simple
pip install --pre --upgrade --timeout=60 --extra-index $dev_anaconda_url numpy
```

PR #2242, opened 2020-09-03
DOC: minor fixes

MaozGelbart:
This PR fixes an internal broken link in the tutorial and a typo. It's great to see 0.11 RC!

### DIFF
--- a/doc/docstrings/scatterplot.ipynb
+++ b/doc/docstrings/scatterplot.ipynb
@@ -257,7 +257,7 @@
    "cell_type": "raw",
    "metadata": {},
    "source": [
-    "Use :func:`relplot` to combine :func:`scatterplot` and :class:`FacetGrid`. This allows grouping within additional categorical variables. and plotting them across multiple subplots.\n",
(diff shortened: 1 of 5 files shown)


mwaskom:
Thanks for catching! I think these should be 

```
:doc:`$PROSE TERM <tutorial/function_overview>
```

Using :doc: rather than :ref: will avoid the annoying bug where the page is not vertically aligned properly after you click on the link (which i think I have tracked down to an issue in the specific theme). But I believe (not certain) you can only link to pages that way, and not to sections.

MaozGelbart:
Thanks, I've updated accordingly.

MaozGelbart:
Now also addresses the first action item in #2244 but not [https://github.com/mwaskom/seaborn/issues/2244#issuecomment-686845902](https://github.com/mwaskom/seaborn/issues/2244#issuecomment-686845902), since it seems to require css knowledge which I'm not familiar with.

mwaskom:
Thanks!

PR #2302, opened 2020-09-27
throw a meaningful error message in lmplot in case  kwarg is missing

neutrinoceros:
`lmplot` de facto requires a `data` keyword argument. Currently, not specifying it throws the following error.
```
TypeError: NoneType is not suscriptatble
```

This patch improves the error message as
```
TypeError: Missing required keyword argument `data`.
```

### DIFF
--- a/seaborn/regression.py
+++ b/seaborn/regression.py
@@ -579,6 +579,9 @@ def lmplot(
                "please update your code.")
         warnings.warn(msg, UserWarning)
 
+    if data is None:
+        raise TypeError("Missing required keyword argument `data`.")
+
     # Reduce the dataframe to only needed columns
     need_cols = [x, y, hue, col, row, units, x_partial, y_partial]
     cols = np.unique([a for a in need_cols if a is not None]).tolist()
--- a/seaborn/tests/test_regression.py
+++ b/seaborn/tests/test_regression.py
@@ -530,6 +530,12 @@ class TestRegressionPlots(object):
         nt.assert_equal(len(ax.lines), 6)
         nt.assert_equal(len(ax.collections), 2)
 
+    def test_lmplot_no_data(self):
+
+        with pytest.raises(TypeError):
+            # keyword argument `data` is required
+            lm.lmplot(x="x", y="y")
+
     def test_lmplot_basic(self):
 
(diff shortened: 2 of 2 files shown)


mwaskom:
Thanks for the PR (with a test even, very nice). I think this will likely be mooted by planned changed for 0.12 that will allow `lmplot` to accept vector `x`/`y`/`hue` inputs (as in other axes-level functions), but this will still be good to have in 0.11.1.

neutrinoceros:
> I think this will likely be mooted by planned changed for 0.12 that will allow lmplot to accept vector x/y/hue inputs (as in other axes-level functions)

Cool ! I also note several inconsistencies in `lmplot` as compared with other plotting functions

- doesn't accept `ax` keyword argument
- doesn't accept `yerr` keyword argument (part of mpl api)
- doesn't work with a non numerical (dates ?) x data

Are you planning to address those for 0.12.0 as well ?

mwaskom:
- doesn't accept ax keyword argument

No, that won't change; see here for details: http://seaborn.pydata.org/tutorial/function_overview.html#figure-level-vs-axes-level-functions. (Other figure-level functions pull `ax` out of kwargs and give a more informative warning about it, so that might get added, but the basic distinction will remain).

- doesn't accept yerr keyword argument (part of mpl api)

I don't think either of the specific underlying matplotlib functions accept `yerr`, but in general seaborn doesn't apply any of its transformations to the kwargs it passes through to matplotlib, so ... if plotting a simple vector it could work, if doing some kind of grouping or semantic mapping it likely won't)

-  doesn't work with a non numerical (dates ?) x data

Yes, there's now (experimental but apparently useful) support for doing generic computations on dates in the new core code, so once the regression module gets refactored to use it, it should be possible to do regressions with dates.

neutrinoceros:
> I don't think either of the specific underlying matplotlib functions accept yerr, but in general seaborn doesn't apply any of its transformations to the kwargs it passes through to matplotlib, so ... if plotting a simple vector it could work, if doing some kind of grouping or semantic mapping it likely won't)

You're right. I think that what I really want here requires a change upstream in mpl.
Thank you so much for this detailed answer, I appreciate it !

PR #2323, opened 2020-10-16
Fix: do not annotate clustermap w/ annot=False

### LINKED ISSUE #2319: Annotations are printed although annot=False in clustermap
#### Description
When plotting a clustermap and specifying `annot=False`, the annotations are printed.
When using heatmap, this doesn't happen.

Currently, this workaround is removing the annotations, when `annot=False` was passed.
```python
if not annot:
   annot = None
```

#### Expected Behaviour
Annotations shouldn't be printed when specifying `annot=False`.

#### Code Example
```python
uniform_data = np.random.rand(10, 12)
sb.clustermap(uniform_data, annot=False)
```
Produces:
image

#### Versions
seaborn 0.11.0; matplotlib 3.2.1

CRiddler:
Fixes #2319 the `clustermap` function was not catching False as it should have been.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -1166,7 +1166,7 @@ class ClusterGrid(Grid):
 
         # Reorganize the annotations to match the heatmap
         annot = kws.pop("annot", None)
-        if annot is None:
+        if annot is None or annot is False:
             pass
         else:
             if isinstance(annot, bool):


CRiddler:
@mwaskom is the failing check due to a CI issue related to python 3.9? Or did I miss something?

mwaskom:
Thanks! (Python 3.9 issues are unrelated).

PR #2328, opened 2020-10-20
MAINT: Drop nose and test on python==3.9

MaozGelbart:
This PR drops the obsolete `nose` package, and rewrites all nose-style tests as modern `pytest`-style tests. Most changes are simple translations of nose asserts to their equivalent pytest `assert` expressions. In a few places pytest parameterization was utilized as well. I did my best to keep the changes to the minimum, but I also changed a few variable names occasionally and removed subclassing from object when I encountered such cases (leftovers from py2-style code).

This PR also adds python==3.9 to the CI tests matrix.

### DIFF
--- a/seaborn/conftest.py
+++ b/seaborn/conftest.py
@@ -7,6 +7,29 @@ import matplotlib.pyplot as plt
 import pytest
 
 
+def has_verdana():
+    """Helper to verify if Verdana font is present"""
+    # This import is relatively lengthy, so to prevent its import for
+    # testing other tests in this module not requiring this knowledge,
+    # import font_manager here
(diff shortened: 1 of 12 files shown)


seaborn/conftest.py
```diff
@@ -7,6 +7,29 @@
 import pytest
 
 
+def has_verdana():
```
MaozGelbart, line 10: I hope this is an acceptable position for these lines. Perhaps we could even generalize the helper to `has_font(fontname)`.

mwaskom (approved):

mwaskom:
This is great — thank you!

PR #2332, opened 2020-10-24
Added option for ci to be a callable.

msf235:
Added the option for ci to be a callable in relational.py. If ci is a callable, it should take a list of values representing multiple observations of the ``y`` variable at the same ``x`` level and return a list [lower_bound, upper_bound] where lower_bound is the lower value for the shaded region and upper_bound is the upper value for the shaded region. ci can also return a different length-two array-like object. A ValueError exception is raised if ci is callable but doesn't satisfy these conditions. I also modified the documentation to reflect this change. Also fixed a comment typo.

Why I think this is a good idea: (1) It mirrors the functionality of estimator, (2) it doesn't change the way the code works when ci isn't passed as a callable, and (3) it allows the user to make error bars based off of any criteria they choose.

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -143,9 +143,14 @@ estimator : name of pandas method or callable or None
     be drawn.
     """,
     ci="""
-ci : int or "sd" or None
+ci : int or "sd" or callable or None
     Size of the confidence interval to draw when aggregating with an
     estimator. "sd" means to draw the standard deviation of the data.
+    If callable, ci should take a list of values representing multiple
+    observations of the ``y`` variable at the same ``x`` level and return
+    a list [lower_bound, upper_bound] where lower_bound is the lower value
+    for the shaded region and upper_bound is the upper value for the shaded
+    region.
     Setting to ``None`` will skip bootstrapping.
     """,
     n_boot="""
@@ -357,7 +362,7 @@ class _LinePlotter(_RelationalPlotter):
         sort=True, err_style=None, err_kws=None, legend=None
     ):
 
-        # TODO this is messy, we want the mapping to be agnoistic about
+        # TODO this is messy, we want the mapping to be agnostic about
         # the kind of plot to draw, but for the time being we need to set
         # this information so the SizeMapping can use it
         self._default_size_range = (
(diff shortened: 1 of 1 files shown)


msf235:
Note that I have not used coverage in the past and so didn't implement my own tests here. I am happy to do that if it is requested.

mwaskom:
Could you comment on #2403, which lays out plans for expanding the flexibility of the error bar plotting? Would you still find a fully generic callable argument necessary if that scheme is implemented? It would significant additional complexity (both implementational and explanatory).

msf235:
I can join the discussion there. My hope was that adding the option for ci to be a callable wouldn't add undue complexity. It seems to be analogous to passing a callable for estimator (which is currently implemented), the only difference being that instead of returning one value the callable should return two. As such, I was hopeful that the approach needed to, say, document this behavior was at least partly worked out already. In my changes to the documentation I followed the template established by the documentation for estimator.

My personal use case is that I am creating error bars by fitting a gamma distribution. I'm doing this because I have data that I know is bounded and highly skewed, and want to take that knowledge into consideration when creating error bars. These are not standard gamma distributions either -- they are shifted and reflected. I believe this is the "proper" approach from a Bayesian perspective -- to choose a prior that captures the basic elements of the data. Regarding #2403, I don't see any bigram approach really fitting my use case.

My perspective is that, if a general callable is implemented, it takes the pressure off of developing a progressively more complicated interface for error bars, which may happen if users increasingly ask for new ways to use error bars (such as if people follow the Bayesian perspective of fitting generic distributions to create these error bars). Then the interface could be kept simple in a way that fits "most" use cases, and the callable is always available to satisfy the remaining cases. That said, a bigram is probably not too complicated and may fit most people's use cases.

For implementing multiple error bars on the same plot, one could potentially pass a list of things for ci. This could, for instance, be a list of callables, each callable corresponding to one error bar. This method could produce an arbitrary number of arbitrary error bars.

So that's my plug. I'll follow along with #2403 and comment where I think I can contribute.

mwaskom:
Out of curiosity, how different are the bootstrap intervals from your bespoke intervals?

msf235:
Here is an example of the difference I see. The first plot shows all the individual realizations, the second plot uses ci=95, and the third uses a (reflected and shifted) gamma distribution fit to make the error bars.
seaborn_demo
I'm not sure why the error bars are so small when ci=95. It could be I'm doing something wrong. Here is a csv file with the data: [data file here](http://s000.tinyupload.com/?file_id=00648673576752029569)

mwaskom:
> I'm not sure why the error bars are so small when ci=95. It could be I'm doing something wrong. 

It looks like your error bars are representing the spread of the distribution of observations, not a proper confidence interval (i.e. the spread of the bootstrap replicates of the estimator), so that discrepancy makes some sense. In the new scheme, the `pi` method will be a nonparametric approach for accomplishing that, but in this example (where you have small N) it would work poorly relative to your Bayesian method.

mwaskom:
Closed by #2407

PR #2341, opened 2020-11-01
add sem error band option to lineplot 

fMeow:
Currently only standard deviation (SD) and confidence interval (CI) are allowed when specifying error band in lineplot. This PR add an option to use standard error of mean (SEM) as error band.


fMeow:
This PR is useless, since setting ci to 68 result in a error band/bar of SEM.

PR #2347, opened 2020-11-12
better checks for `sharey` and `sharex` in `axesgrid.py`

### LINKED ISSUE #2346: FacetGrid sharey='col' prevents non-left yticklabels
I want to use `FacetGrid` and only `sharey` across `col`umns (not rows). (I am not sure whether this is against the philosophy of faceting.) When doing so, I would expect to see yticklabels in the all rows, but dont find them. Is this a bug or intended?
I dont want to use `sharey=False`, because I want to compare the columns, e.g. having them on the same scale.

```python
import seaborn as sns
sns.__version__
'0.12.0.dev0'
df = sns.load_dataset("titanic")
```
What I get:
```python
sns.catplot(
    data=df, kind="point", sharex=True, sharey='col',
    x="class", y="fare", row='alive', col='sex')
```
image

What I'd expect:
```python
expected = sns.catplot(
    data=df, kind="point", sharex=True, sharey='col',
    x="class", y="fare", row='alive', col='sex')
for ax in expected.axes.flatten():
    ax.tick_params(labelleft=True)
```
image

What `plt` does:
```python
plt.subplots(ncols=2,nrows=2,sharex=True, sharey='col')
```
image

aaronspring:
better checks for `sharey` and `sharex` in `axesgrid.py`

closes #2346

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -461,13 +461,13 @@ class FacetGrid(Grid):
         if despine:
             self.despine()
 
-        if sharex:
+        if sharex != 'row':
             for ax in self._not_bottom_axes:
                 for label in ax.get_xticklabels():
                     label.set_visible(False)
                 ax.xaxis.offsetText.set_visible(False)
 
-        if sharey:
+        if sharey != 'col':
             for ax in self._not_left_axes:
                 for label in ax.get_yticklabels():
                     label.set_visible(False)
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -461,13 +461,13 @@ class FacetGrid(Grid):
         if despine:
             self.despine()
 
-        if sharex != 'row':
+        if sharex in [True, 'col']:
             for ax in self._not_bottom_axes:
                 for label in ax.get_xticklabels():
                     label.set_visible(False)
                 ax.xaxis.offsetText.set_visible(False)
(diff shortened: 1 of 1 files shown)


seaborn/axisgrid.py
```diff
@@ -461,13 +461,13 @@ def __init__(
         if despine:
             self.despine()
 
-        if sharex:
+        if sharex != 'row':
```
mwaskom, line 464: Won’t this hide the labels when `sharex=False`?

mwaskom (changes requested):

seaborn/axisgrid.py
aaronspring: true. checked by eye again and fixed. I didnt find any tests for `sharex` and `sharey` in `axesgrid.py`. hope without tests this is ok.

```python
import seaborn as sns
df = sns.load_dataset("titanic")
for sharex in [True, False, 'row','col']:
    for sharey in [True, False, 'row','col']:
        g=sns.catplot(
            data=df, kind="point", sharex=sharex, sharey=sharey,
            x="class", y="fare", row='alive', col='sex')
        plt.suptitle(f'sharex={sharex} & sharey={sharey}')
        plt.show()
```

mwaskom (approved):

mwaskom:
Thanks!

PR #2353, opened 2020-11-15
Allow independent variable to be plotted on y-axes in lineplot

matzegoebel:
In lineplots, seaborn assumes the independent variable to be drawn on the x-axis and thus sorts the x-dimension first. If we want the independent variable to be drawn on the y-axis and therefore swap the axes by using the x and y arguments of `lineplot`, we do not get a mirror image of the former plot. Instead, the values are still ordered along the x dimension. Drawing error bars or bands, then will not work at all.
To allow producing such a mirror image, the `vals` and `grouper` variables for the `aggregate` function need to be swapped. I introduced the argument `sort_dim` to allow the user to decide which dimension should be sorted first.
To draw the errors correctly in the case of `sort_dim="y"`, `fill_betweenx` is used for band style and `xerr` is used instead of `yerr` for bar style.
I find this feature important when plotting e.g. meteorological variables as a function of height. Apart from [this](https://stackoverflow.com/questions/55362200/switch-seaborn-x-and-y-axes-but-compute-standard-deviation-on-original-orientati) stackoverflow question, I couldn't find any information on this issue.

example code:

```
import seaborn as sns
import numpy as np
import pandas as pd

height= np.arange(0, 6.3, 0.1)
height_ind = np.repeat(height, 200)
rand = np.random.normal(scale=0.1, size=200*len(height))
vals = np.sin(height_ind) + rand

df = pd.DataFrame()
df["height"] = height_ind
df["vals"] = vals

#independent variable on x-axis
plt.figure()
g = sns.lineplot(x="height", y="vals", ci="sd", data=df)
plt.show()

#independent variable on y-axis with data still sorted along x
plt.figure()
g = sns.lineplot(y="height", x="vals", ci="sd", data=df)
plt.show()

#independent variable on y-axis with data sorted along y
plt.figure()
g = sns.lineplot(y="height", x="vals", ci="sd", sort_dim="y", data=df)
plt.show()

```

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -354,7 +354,7 @@ class _LinePlotter(_RelationalPlotter):
         self, *,
         data=None, variables={},
         estimator=None, ci=None, n_boot=None, seed=None,
-        sort=True, err_style=None, err_kws=None, legend=None
+        sort=True, sort_dim="x", err_style=None, err_kws=None, legend=None
     ):
 
         # TODO this is messy, we want the mapping to be agnoistic about
@@ -371,6 +371,7 @@ class _LinePlotter(_RelationalPlotter):
         self.n_boot = n_boot
         self.seed = seed
         self.sort = sort
+        self.sort_dim = sort_dim
         self.err_style = err_style
         self.err_kws = {} if err_kws is None else err_kws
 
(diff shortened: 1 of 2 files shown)


mwaskom:
Better late than never ....

I pulled this into #2854, mostly to deal with some merge conflicts (caused by abstractions that actually made the implementation a little easier). I also changed the parameter name to `orient`, as the purpose is a bit more general than controlling the sort (i.e. it also controls the aggregation) and to align with the general functionality in the new objects interface.

Thanks for getting the ball rolling here!

matzegoebel:
nice, thank you @mwaskom! I didn't think this would ever be merged. Yes, `orient` is definitely a better name for the argument.

PR #2359, opened 2020-11-18
Fix jointplot reference syntax in distributions tutorial.

bdice:
Hi! I caught a typo in the Sphinx syntax of this page: https://seaborn.pydata.org/tutorial/distributions.html
This PR fixes the missing colon, which will make the reference render.

I also fixed one other rST syntax error, there was a missing space that prevented the Sphinx double-backtick code block from rendering correctly.

Thanks for the wonderful package. 😄


mwaskom:
Thanks!

PR #2363, opened 2020-11-25
TST: FIX: Consider array length in artist arrays comparison

### LINKED ISSUE #2329: `_testing.assert_artists_equal()` does not compare artists properly
This has a similar root cause as in #2281 : the test uses zip to compare two artist arrays but in the case one array is empty and the other isn't, no assertion will fail although artists arrays differ. Identified this while working on #2328 but I didn't want to fix it there since it's out of scope for that PR. A quick fix involves asserting arrays length equality first. Trying to add this comparison results with a few failures:
```
FAILED seaborn/tests/test_axisgrid.py::TestJointPlot::test_kde - AssertionError
FAILED seaborn/tests/test_axisgrid.py::TestJointPlot::test_scatter_hue - AssertionError
FAILED seaborn/tests/test_axisgrid.py::TestJointPlot::test_palette - AssertionError
FAILED seaborn/tests/test_distributions.py::TestDisPlot::test_with_rug[kwargs0] - AssertionError
FAILED seaborn/tests/test_distributions.py::TestDisPlot::test_with_rug[kwargs2] - AssertionError
```

MaozGelbart:
Fixes #2329 .

### DIFF
--- a/seaborn/_testing.py
+++ b/seaborn/_testing.py
@@ -40,6 +40,7 @@ BAR_PROPS = [
 
 def assert_artists_equal(list1, list2, properties):
 
+    assert len(list1) == len(list2)
     for a1, a2 in zip(list1, list2):
         prop1 = a1.properties()
         prop2 = a2.properties()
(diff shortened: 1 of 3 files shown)


mwaskom:
Thanks! That default `zip` behavior is a real gotcha...

MaozGelbart:
> Thanks! That default `zip` behavior is a real gotcha...

So much! At least python got that sorted for python 3.10+ with a new `strict=True` [zip kwarg](https://docs.python.org/3.10/library/functions.html#zip).

PR #2375, opened 2020-12-07
fix for an OverflowError

Dimasik007:
Libraries like PyFolio still use this function and in some cases it might lead to an OverflowError: cannot convert float infinity to integer. Here is a quick fix for these cases returning just a default value of bins.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -2396,7 +2396,10 @@ def _freedman_diaconis_bins(a):
     if h == 0:
         return int(np.sqrt(a.size))
     else:
-        return int(np.ceil((a.max() - a.min()) / h))
+        try:
+            return int(np.ceil((a.max() - a.min()) / h))
+        except OverflowError:
+            return 30
 
 
 def distplot(a=None, bins=None, hist=True, kde=True, rug=False, fit=None,


mwaskom:
Can you please give a little bit more context?

Dimasik007:
> Can you please give a little bit more context?

the line that I've modified `int(np.ceil((a.max() - a.min()) / h))` should always return an int but this will fail in some cases if the value inside int() will be an infinity, e.g. `int(float("inf"))` and will result in an OverFlowError. I ran into this issue when I was creating tear sheets with PyFolio and my values had lots of decimal places. 
The fix with some default value was helpful to me, so I decided to propose it here.

mwaskom:
A code example that produces an error/invalid output would be helpful. I’m not sure what you mean by “my values had lots of decimal places”.

Dimasik007:
> A code example that produces an error/invalid output would be helpful. I’m not sure what you mean by “my values had lots of decimal places”.

here is a code example that produces OverflowError: cannot convert float infinity to integer
```
import pandas as pd

some_data = [10, 50, 2, 1000, 30, 98, 0, 12, 50]
df = pd.DataFrame(some_data, columns=['a'])
df['new'] = df.pct_change()
result = _freedman_diaconis_bins(df['new'][1:])
```

mwaskom:
In this example, your input data contains `inf`.

Does your third-party library call this function directly? It's a private function and may be removed at any time without warning...

mwaskom:
Moreover, if you did

```python
sns.distplot(df["new"][1:], bins=30)
```

it would continue to fail with

```python-traceback
<ipython-input-460-1ee25462e92b> in <module>
----> 1 sns.distplot(df["new"][1:], bins=30)

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/seaborn/distributions.py in distplot(a, bins, hist, kde, rug, fit, hist_kws, kde_kws, rug_kws, fit_kws, color, vertical, norm_hist, axlabel, label, ax, x)
   2610         orientation = "horizontal" if vertical else "vertical"
   2611         hist_color = hist_kws.pop("color", color)
-> 2612         ax.hist(a, bins, orientation=orientation,
   2613                 color=hist_color, **hist_kws)
   2614         if hist_color != color:

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/matplotlib/__init__.py in inner(ax, data, *args, **kwargs)
   1436     def inner(ax, *args, data=None, **kwargs):
   1437         if data is None:
-> 1438             return func(ax, *map(sanitize_sequence, args), **kwargs)
   1439 
   1440         bound = new_sig.bind(ax, *args, **kwargs)

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/matplotlib/axes/_axes.py in hist(self, x, bins, range, density, weights, cumulative, bottom, histtype, align, orientation, rwidth, log, color, label, stacked, **kwargs)
   6654             # this will automatically overwrite bins,
   6655             # so that each histogram uses the same bins
-> 6656             m, bins = np.histogram(x[i], bins, weights=w[i], **hist_kwargs)
   6657             tops.append(m)
   6658         tops = np.array(tops, float)  # causes problems later if it's an int

<__array_function__ internals> in histogram(*args, **kwargs)

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/numpy/lib/histograms.py in histogram(a, bins, range, normed, weights, density)
    790     a, weights = _ravel_and_check_weights(a, weights)
    791 
--> 792     bin_edges, uniform_bins = _get_bin_edges(a, bins, range, weights)
    793 
    794     # Histogram is an integer or a float array depending on the weights.

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/numpy/lib/histograms.py in _get_bin_edges(a, bins, range, weights)
    424             raise ValueError('`bins` must be positive, when an integer')
    425 
--> 426         first_edge, last_edge = _get_outer_edges(a, range)
    427 
    428     elif np.ndim(bins) == 1:

~/miniconda3/envs/pulses-gambling/lib/python3.8/site-packages/numpy/lib/histograms.py in _get_outer_edges(a, range)
    313                 'max must be larger than min in range parameter.')
    314         if not (np.isfinite(first_edge) and np.isfinite(last_edge)):
--> 315             raise ValueError(
    316                 "supplied range of [{}, {}] is not finite".format(first_edge, last_edge))
    317     elif a.size == 0:

ValueError: supplied range of [-1.0, inf] is not finite
```

PR #2379, opened 2020-12-09
FIX: Infer categorical dtypes before boolean resolution

### LINKED ISSUE #2317: Categorical variable with numerical classes incorrectly plotted only on y axis
Dear All,

I think I may have found a bug in Seaborn categorical plotting functions.
If a DataFrame column contains numbers, and it is explicitely converted to CategoricalDType, it will be correctly treated as categorical by functions such as boxplot only if it is passed to `x` argument.
If it is instead passed to `y` argument of the plotting function, it will be treated as numerical, and the other variable will be treated as categorical (even though the other variable is obviously numerical).

I am using Python 3.7, Matplotlib 3.3.1, Seaborn 0.11, Numpy 1.19.1, and Pandas 1.1.1 .
I am using Matplotlib backend: 'module://ipykernel.pylab.backend_inline'

Reproducible code:
```python

#Create DataFrame with a numerical column, and a categorical column with number as classes.
n_points = 50

df = pd.DataFrame({"value": np.random.rand(n_points),
                   "class": np.random.randint(0, 2, n_points),
                   })

#Explicitely set `class` column to categorical data type.
df["class"] = df["class"].astype("category")

#If `class` column is passed to `x` argument, the plot is correct: column is treated as categorical.
sns.boxplot(data=df,
            x="class",
            y="value",
            );

#BUG: If `class` column is passed to `y` argument, the plot is incorrect: `class` is treated as numerical!
sns.boxplot(data=df,
            x="value",
            y="class",
            );

```

Please take a look at
[bug_categorical_numbers.pdf](https://github.com/mwaskom/seaborn/files/5371402/bug_categorical_numbers.pdf)
 the attached PDF file with the correct and wrong plots.

MaozGelbart:
Fixes #2317, with small changes to the docstring of `variable_type`.

### DIFF
--- a/seaborn/_core.py
+++ b/seaborn/_core.py
@@ -1199,6 +1199,10 @@ def variable_type(vector, boolean_type="numeric"):
     var_type : 'numeric', 'categorical', or 'datetime'
         Name identifying the type of data in the vector.
     """
+    # If a categorical dtype is set, infer categorical
+    if pd.api.types.is_categorical_dtype(vector):
+        return "categorical"
+
     # Special-case all-na data, which is always "numeric"
     if pd.isna(vector).all():
         return "numeric"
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks!

PR #2385, opened 2020-12-15
Update README.md

aflah02:
Added Links for Python 3.6, Python 2 and PyPI so that someone who is not familiar with either of them can check them out easily.


mwaskom:
Thanks, but I don't think these changes are necessary.

aflah02:
Ohkk got it thanks 👍  @mwaskom

PR #2386, opened 2020-12-16
Add/restore functionality to long-form data processing

### LINKED ISSUE #2263: Referencing hierarchical columns gives error in seaborn 0.11
I updated to seaborn 0.11 and get an error with code that did not even give a warning with seaborn 0.10.

I have a pandas DataFrame with hierarchical columns. That means that I reference a column like so:

```df[('0th_order_col_name', '1st_order_col_name')]```

When I try to specify such a hierarchical column label in, for instance, a scatterplot, seaborn now gives me an error because it tries to interpret this hierarchical column label as data.

```
sns.scatterplot(x=('0th_order_col_name', '1st_order_col_name_FOO'), y=('0th_order_col_name', '1st_order_col_name_BAR'), data=df)
```

The error message is

```
ValueError: Length of tuple vectors must match length of `data` when both are used, but `data` has length 5 and the vector passed to `x` has length 2.
```

The change was introduced in #2205

### LINKED ISSUE #2279: scatterplot throws an error when referencing int column by name
## Summary
`scatterplot` throws an error when trying to reference column by name if the column name is an int

In seaborn 0.10 and lower, I was able to pass column names to `scatterplot` that were integers (e.g. `scatterplot(data=data, x=0, y=1)`). After updating to 0.11, I'm getting the error below. 

Apologies if already posted but I did not see anything about this in the issue tracker. Thanks for the awesome package!

## Environment
seaborn version: 0.11.0 (with 0.10.1, the code below will run fine)
matplotlib version: 3.1.3
matplotlib backend: TkAgg

## Code to reproduce
```
import seaborn as sns
import matplotlib as mpl

print(f"seaborn version: {sns.__version__}")
print(f"matplotlib version: {mpl.__version__}")
print(f"matplotlib backend: {mpl.get_backend()}")

iris_data = sns.load_dataset("iris")

# this one works
sns.scatterplot(data=iris_data, x="sepal_length", y="sepal_width")

# make a new column called int(0)
iris_data[0] = iris_data["sepal_length"].copy()

# this one breaks (used to work in seaborn <0.11)
sns.scatterplot(data=iris_data, x=0, y="sepal_width")

```

## Output
```
seaborn version: 0.11.0
matplotlib version: 3.1.3
matplotlib backend: TkAgg
  File "/Users/bpedigo/JHU_code/graspy_misc/graspy-misc/localization/sns_bug.py", line 16, in <module>
    sns.scatterplot(data=iris_data, x=0, y="sepal_width")
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/_decorators.py", line 46, in inner_f
    return f(**kwargs)
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/relational.py", line 798, in scatterplot
    alpha=alpha, x_jitter=x_jitter, y_jitter=y_jitter, legend=legend,
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/relational.py", line 580, in __init__
    super().__init__(data=data, variables=variables)
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/_core.py", line 604, in __init__
    self.assign_variables(data, variables)
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/_core.py", line 668, in assign_variables
    data, **variables,
  File "/Users/bpedigo/miniconda3/envs/graspy_misc/lib/python3.7/site-packages/seaborn/_core.py", line 895, in _assign_variables_longform
    if val is not None and len(data) != len(val):
TypeError: object of type 'int' has no len()
```

mwaskom:
The variable processing refactor made long-form argument processing more strict, allowing only string(y) types as keys to the `data` object. While this is basically what was documented, many functions previously worked properly with numeric arguments or even tuples (i.e. as keys for multi-index columns).

This PR makes the processing of named variables with long-form data more flexible.

It introduces a couple minor ambiguities, e.g. when `data` is present and a variable is specified with a number or a tuple, the variable could be interpreted as a key for `data` or as a data value itself (preferring the former). But I think in most cases, the expected behavior it to interpret it as a key.

~Needs tests, but I wanted to get CI going.~

Fixes #2279
Fixes #2263

Wide-form data objects with hierarchical column index still don't work properly, and making that work is nontrivial, so those remain unsupported (for now at least).

### DIFF
--- a/doc/releases/v0.11.1.txt
+++ b/doc/releases/v0.11.1.txt
@@ -2,13 +2,15 @@
 v0.11.1 (Unreleased)
 --------------------
 
+- |Fix| Restored support for using tuples or numeric keys to reference fields in a long-form `data` object (:pr:`2386`).
+
 - |Fix| Fixed a bug in :func:`lineplot` where NAs were propagating into the confidence interval, sometimes erasing it from the plot (:pr:`2273`).
 
 - |Fix| Fixed a bug in :class:`PairGrid`/:func:`pairplot` where diagonal axes would be empty when the grid was not square and the diagonal axes did not contain the marginal plots (:pr:`2270`).
 
 - |Fix| Fixed a bug in :class:`PairGrid`/:func:`pairplot` where off-diagonal plots would not appear when column names in `data` had non-string type (:pr:`2368`).
 
-- |Fix| Fixed a bug where categorical dtype information was ignored when data consisted of boolean values (:pr:`2379`).
+- |Fix| Fixed a bug where categorical dtype information was ignored when data consisted of boolean or boolean-like values (:pr:`2379`).
 
 - |Fix| Fixed a bug in :class:`FacetGrid` where interior tick labels would be hidden when only the orthogonal axis was shared (:pr:`2347`).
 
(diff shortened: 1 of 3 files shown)


mwaskom:
One possibility would be to only enable the new functionality for proper columns of the dataframe, requiring string keys to reference the index. This probably makes sense? (Otherwise there's actually an ambiguity in many cases, although it's sensible to expect that the dataframe columns are prioritized over the index levels). But it does add complexity to what was otherwise a simple change.

mwaskom:
Decided to go with the approach proposed in the previous comment ... it is the easiest for the user to work around because if you really want to access index columns that have no name, you can just do `df.reset_index()`, whereas if a value of `0` always pulled out the index (unless it had a name) I expect it would be more surprising and annoying to work around.

Issue #2283, opened 2020-09-16
SEM and custom errorbars on pointplot


JoeZiminski:
Thank you very much for this excellent resource. For my own use I have added the functionality to also plot standard error of the mean, and custom errorbars for pointplot:

starting line 1477 of categorical.py in the _PointPlotter.estimate_statistic function (shown after sd):
```
                    if ci == "sd":  

                        estimate = estimator(stat_data)
                        sd = np.std(stat_data)
                        confint.append((estimate - sd, estimate + sd))

                    elif ci == "sem":                                                       

                        estimate = estimator(stat_data)
                        sem = np.std(stat_data) / np.sqrt(len(stat_data))
                        confint.append((estimate - sem, estimate + sem))

                    elif ci[0] == 'custom':

                        estimate = estimator(stat_data)
                        ci_ = ci[1][i]
                        confint.append((estimate - ci_, estimate + ci_))
```

At the moment it is a bit ugly, and requires ci to be bassed as a two inputs (one 'custom', the other an array of errorbars). so might lead to issues in its current form. However am happy to work on it if its something youd consider having as part of the package.

Best,
Joe

mwaskom:
I'm -1 on special casing this, but more general support is on the roadmap.

If you want error bars that correspond to the standard error of the mean, you can also show the 68% bootstrap confidence interval.

But 95% CI is always more informative.

Issue #1501, opened 2018-07-17
Support plotting quantiles of the data distribution with lineplot


shoyer:
`seaborn.lineplot` makes it easy to plot either a confidence intervals around the estimator, or the standard deviation of the data. It would be nice to directly support plotting quantiles of the data distribution instead, rather than only the standard deviation.

Maybe it would make sense to have a separate key-word argument for controlling whether `ci` refers to the estimator or data distribution? e.g., `ci_source='estimator'` vs `ci_source='data'`? Then `ci` would only indicate how large of a confidence interval to plot, not the source.

shoyer:
Looking more carefully at Seaborn's API, I see that the `ci` argument is also used with the same meaning in `pointplot` and `barplot`. So certainly caution is warranted in updating this API, given that these are *not* new functions.

My normal inclination would be to support this sort of thing by adding configuration objects (e.g., `ci=EstimateCI(alpha=0.05)` or `ci=DataCI(standard_deviations=2)`), but this would be an unusual pattern for an external API in Seaborn.

Maybe an `dict` would make more sense for Seaborn, e.g., `ci=dict(source='estimate', standard_deviations=2)`? If we're OK with fully explicit names, we could start with fields:
- `source='data'` or `source='estimate`
- `standard_deviations` or `confidence_level` (mutually exclusive) for indicating the number of standard deviations or the confidence level

mwaskom:
> Looking more carefully at Seaborn's API, I see that the ci argument is also used with the same meaning in pointplot and barplot.

Also in `regplot`. And I got lazy when implementing `ci="sd"` and just duplicated the logic everywhere instead of writing so more abstract way of doing things.

Another reason I've been resistant to overloading the `ci` parameter when this came up in the past is terminological. It doesn't *really* make sense to say "the confidence interval of the data" (I think technically this is a "predictive interval?") or "the confidence interval is two standard deviations". A better forward-looking name for this would have been derived from "error bar" that could take a simple tuple spec (e.g. `ebar=("ci", 95)`).

> My normal inclination would be to support this sort of thing by adding configuration objects (e.g., ci=EstimateCI(alpha=0.05) or ci=DataCI(standard_deviations=2)), but this would be an unusual pattern for an external API in seaborn.

Yeah I've also thought about something like this, especially when we were discussing adding more advanced bootstrapping algorithms that require a few more parameters, but as you say it would be unlike the way anything else works in seaborn.

cypressf:
[...]

Issue #1492, opened 2018-07-12
[Feature request] Support `lineplot` with error bands of list of scaled standard deviations


zuoxingdong:
In the context of research, e.g. we collect data with 10 random seeds, it is very nice to use `lineplot` to draw uncertainty bands. 

However, we often need to have a few more shaded uncertainty bands to plot, for example
1. darker shaded area with +- 0.5* standard deviation
2. medium translucent shaded area with +- 1* standard deviation
3. lighter shaded area with +- 2* standard deviation

What has been supported for now is we can either set up `ci` values as percentage of confidence interval or using `ci="sd"` to draw +- 1 standard deviation. 

It will be much easier to support things like `ci=[(0.5, "sd"), (1.0, "sd"), (2.0, "sd")]` and transparency automatically adjusted for each bands.

mwaskom:
#2407 added the ability to draw errorbars with a parameterized (scaled) standard deviation errorbar  e.g. 

```python
fmri = sns.load_dataset("fmri")
sns.lineplot(data=fmri, x="timepoint", y="signal", errorbar=("sd", 2))
```

Adding multiple errorbars within a single function call ended up feeling like it added too much complexity at this stage. However, it is of course possible to achieve by calling `lineplot` multiple times e.g.

```python
fmri = sns.load_dataset("fmri")
sns.lineplot(data=fmri, x="timepoint", y="signal", errorbar=("sd", 1), color="C0")
sns.lineplot(data=fmri, x="timepoint", y="signal", errorbar=("sd", 2), color="C0", ls="")
```

matplotlib's alpha blending is not perfect, but I think this works well enough for most cases. It's possible that multiple errorbars will get revisited in the future if the architecture changes enough to make it more straightforward to add, but I am going to close this issue for now.

PR #2407, opened 2020-12-29
Increase errorbar flexibility

### LINKED ISSUE #2403: Notes on errorbar enhancements
Here are some notes on planned changes to error bar specification.

Currently, what the error bars show is controlled through the `ci` parameter. This can be either a number, setting the width of a bootstrap confidence interval, or the string `"sd"`, indicating that the error var covers +/- the standard deviation of the data around the estimate value.

Some problems with this have been routinely noted:
- There's no option for parametric confidence intervals/standard error
- There's no option for showing a measure of data spread other than +/- 1 sigma
- `ci="sd"` does not really make conceptual sense as a parametrizing (it is the result of a short-sighted API decision)

In effect, you can think of the options as having a 2D taxonomy defined by whether the error bars show a measure of estimate certainty or data spread and whether the computation is parametric or nonparametric. Currently, we occupy two cells in this matrix:

| | Estimate certainty | Data spread |
|-|-|-|
|Parametric| | `ci="sd"`|
|Nonparametric| `ci=95`, `ci=68`, etc. | |

I would like to fill out the matrix. But we have a few challenges:

- As mentioned, the current API is not great, and overloading the meaning of `ci` further is a nonstarter
- There is no centralized location in the code where this parameter is interpreted and used

Plans for the new API will involve a new parameter, probably called `errorbar` but possibly `error`, `errbar` or some other shorthand, that accepts a tuple of the form `(kind, level)`. The first element determines what the error bars show, and the second parametrizes them. One proposal is to fill out the space like this:

| | Estimate certainty | Data spread |
|-|-|-|
| Parametric | `("se", scale)` | `("sd", scale)` |
| Nonparametric | `("ci", size)` | `("pi", size)` |

IMO, there is a lot of sense to this. You have four options for `kind`, each named using a bigram initialism. There are two kinds of level parameters:
- `scale`: multiplicatively scales a parametric error metric (e.g. `("sd", 3)` gives you a 3-sigma error bar, `("se", 1.96)` gives you a ~95% parametric confidence interval
- `size` sets the size of a nonparametric interval with percentiles (of the boostrap distribution for `ci` and the input data for `pi`) of `(1 - size) / 2, 1 - (1 - size) / 2`

There are also some potential drawbacks
- `"pi"` (i.e., "percentile interval") doesn't seem to be a commonly used term for a nonparametric measure of data spread. Actually I'm not sure there really is a term in the stats literature for such an interval, even though it's a very reasonable thing to plot (e.g. #1501)
- If you really want parametric 95% confidence intervals, this parametrization leaves you limited to a Z interval (and requires you to understand how to construct one from a standard error)

API decisions aside, the right implementation is going to take some thinking. Currently each module does its own errorbar computations. Most errorbars appear in the context of an aggregation-with-estimator operation. This can likely be abstracted. The other place they show up is in the regression module, where error bars are shown around the regression line. This needs to be handled differently, but statsmodels now has the `get_prediction` method which will do a lot of the work for us. We'll need a general enough implementation such that we can handle special cases (like logistic regression, where the SE/SD scaling should happen in logit space).

Here are some assorted open questions

- Should we accept simple strings (e.g. `errorbar="sd"`) with a default level value used internally?
- This simple 4 option system is still fairly limiting; it may disappoint those who would like to be able to use a generic function to get error bars (e.g. #2332). What might that API look like?
- Is it a sensible API option for `sd` to correspond to the prediction interval in a regression model?
- Should standard error correspond to the estimator and raise if the one used doesn't have a defined standard error? In other words, what would we do with `estimator="median", errorbar="se"`? And if the estimator is a callable, should we use its name to associate with the correct standard error function?
- It would be nice to have seaborn support multiple error bars from a sequence of `level` parameters, e.g. 1-2-3 sigma bands or 68-95-99 CIs (e.g. #1492). I like this kind of plot, but each plotting function will have to define its own logic for showing multiple error bars (e.g. layered alpha for error bands in `lineplot`, lines of diminishing width in `pointplot`). But still, if it's going to happen, we should at least plan for it here.
- What about additional arguments for bootstrapping (i.e. `n_boot`, `seed`?) It would be nice to reduce the number of parameters in the main function signatures, but I would like to keep the argument for `errorbar` a simple tuple and not a more complex object that could take optional parameters. I think...
- What about `loess`? (#552). Bootstrapping is still very slow, but statsmodels seems to still not have analytic confidence bands.
- What's the right order of operations for working on this? It should probably not be (fully) implemented until the categorical/regression modules can be refactored to use the core objects (where this should be handled).

mwaskom:
This PR adds a new internal statistics object for doing aggregation with an estimator function and expands the flexibility of the error bars that operation can produce.

This implements and closes #2403. More logic is given there. Briefly, the new errorbars can be:

- "ci", for a bootstrap confidence interval of a given width (default 95)
- "pi", for a percentile interval of the data with a given width (default 95)
- "se", for +/- the standard error, scaled by a given factor (default 1)
- "sd", for +/- the standard deviation, scaled by a given factor (default 1)
- an arbitrary callable that maps a vector to a (min, max) interval

To Dos:
- [x] Use this function within the relational module (other modules will need to come after refactor)
- [x] Address TODOs in bootstrap function, making it a bit easier to work with
- [x] Deprecate old `ci=` parameter
- [x] Decision about raising on `errorbar="se"` when `estimator` is not `mean`
- [x] Handle log-scaled axes properly
- [x] Write out a stub for a user guide page on this (completion will need integration in other modules)
- [x] Write out a stub for the release notes

This (partly) addresses several longstanding issues. It won't fully address them until the logic is integrated into the other modules; doing so will happen after they are refactored to use the new core objects. But I will close the open issues now, and then link back here when the downstream work is implemented. Relevant issues include #331, #1501, #1427, #1492, #1892, #2283, #2332,

### DIFF
--- a/doc/docstrings/lineplot.ipynb
+++ b/doc/docstrings/lineplot.ipynb
@@ -222,7 +222,7 @@
    "cell_type": "raw",
    "metadata": {},
    "source": [
-    "Show error bars instead of error bands and plot the 68% confidence interval (standard error):"
+    "Show error bars instead of error bands and extend them to two standard error widths:"
    ]
   },
   {
@@ -232,7 +232,7 @@
    "outputs": [],
    "source": [
     "sns.lineplot(\n",
-    "    data=fmri, x=\"timepoint\", y=\"signal\", hue=\"event\", err_style=\"bars\", ci=68\n",
+    "    data=fmri, x=\"timepoint\", y=\"signal\", hue=\"event\", err_style=\"bars\", errorbar=(\"se\", 2),\n",
     ")"
    ]
   },
--- a/doc/releases/v0.12.0.txt
+++ b/doc/releases/v0.12.0.txt
@@ -2,6 +2,8 @@
 v0.12.0 (Unreleased)
 --------------------
 
+- |API| |Feature| |Enhancement| TODO (Flesh this out further). Increased flexibility of what can be shown by the internally-calculated errorbars (:pr:2407).
+
 - Made `scipy` an optional dependency and added `pip install seaborn[all]` as a method for ensuring the availability of compatible `scipy` and `statsmodels` libraries. This has a few minor implications for existing code, which are explained in the Github pull request (:pr:`2398`).
 
 - Following `NEP29 <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_, dropped support for Python 3.6 and bumped the minimally-supported versions of the library dependencies.
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -52,6 +52,25 @@ User guide and tutorial
             </div>
           </div>
         </div>
+        <div class="row">
+          <div class="col-md-6">
+            <div class="col-md-4">
+              <a href=./tutorial/error_bars.html>
+                <img src="_images/error_bars_4_0.png" class="img-responsive center-block">
+              </a>
+            </div>
+            <div class="col-md-8">
+
+.. toctree::
+   :maxdepth: 2
+
+   tutorial/error_bars
+
+.. raw:: html
+
+            </div>
+          </div>
+        </div>
       </div>
     </div>
     <div class="row">
--- a/doc/tutorial/error_bars.ipynb
+++ b/doc/tutorial/error_bars.ipynb
@@ -1,0 +1,358 @@
+{
+ "cells": [
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    ".. _errorbar_tutorial:\n",
+    "\n",
+    ".. currentmodule:: seaborn"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "hide"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import seaborn as sns\n",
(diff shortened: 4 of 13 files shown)


mwaskom:
It's possible this behavior will change a bit if difficulties are encountered while incorporating it into the categorical and regression modules, but I am going to merge this for now.

sdaza:
This doesn't seem to work with `barplot`: 

```python 
print(sns.__version__)
0.12.0.dev0

sns.barplot(y='g4', x='p_genetic_variance', hue='p_vertical_trans_variance', data=s, errorbar = ('pi', 95))
AttributeError: 'Rectangle' object has no property 'errorbar'
```

mwaskom:
Is there anything that claims it currently does? The [draft release notes](https://github.com/mwaskom/seaborn/blob/master/doc/releases/v0.12.0.txt#L48) say

> Increased the flexibility of what can be shown by the internally-calculated errorbars for :func:`lineplot`. With the new `errorbar` parameter, it is now possible to select bootstrap confidence intervals, percentile / predictive intervals, or intervals formed by scaled standard deviations or standard errors. As a consequence of this change, the `ci` parameter has been deprecated. Similar changes will be made to other functions that aggregate data points in future releases. (:pr:`2407`).

sdaza:
My bad. I thought it was a general change that would affect most plotting functions. Thanks!

Issue #2231, opened 2020-08-28
Notes on ongoing legend challenges


mwaskom:
Here are some notes about legends, which are an area of future work.

Challenges with legends include:

## Multi-variable legends

When multiple semantic mappings exist (currently only possible in the relational plots), the legends use a hack that adds a legend entry with an invisible artist handle to serve as a "subtitle". This is not ideal. Matplotlib legends do not have any concept of a "subtitle". It's possible to add multiple legends to a plot. But it's not possible to add multiple legends that both use `loc="best"`; they will be placed in the same location. Because `loc="best"` is determined at draw-time, it's also not possible how to but one legend in the "best" location and then stick the other one next to or under it. It would also be better if it were possible to assign the sub-legends to e.g. different columns). See #1519 for some related work.

## Numeric semantic mappings

When using a numeric semantic mapping, a set of handles is added to the legend with evenly-spaced labels. This is intended to act like a "colorbar", but it doesn't look that much like a colorbar, and it's an ongoing source of confusion. Ideally, there would be something that looks like a colorbar inside the legend itself.

## No consistent approach to legend creation

Most plotting functions add proxy artists with label attributes to the axes and then call `legend()` to pick them up. This has the advantage that `legend()` can be called again with new kwargs (e.g. a different location) and it will be recreated. But it adds a lot of extraneous artists to the legend, which is confusing when trying to do deep customization. The new distribution plots create proxy artists but just pass them directly to `legend()`. The disadvantage is that calling `legend()` again will erase these artists. And in general, this should be predictable, and users should have access to the legends and their handles should they need to modify them.

## Legends are not very customizable

For numeric mappings, there's no way to specify exactly which labels to use. And in general, plotting functions don't take a `legend_kws` dict, but probably they should.

## Matplotlib legend API has limitations

Some of these problems exist because the matplotlib legend API has some limitations, both for users and for seaborn. Some of these problems would be less painful if it were easier to modify a legend after plotting, e.g. to move it or to add new entries to it. Moving a legend seems basically impossible through the public API, adding entries doesn't seem possible either. Some aspects of the legends can be modified in-place, but not through obvious methods. And the legend code uses a lot of private attributes and functions, so extending it (i.e. through subclassing) is not really possible

Unclear at this point whether the best path forward is to try and get an enhanced legend API from matplotlib or to fork the legend code and adapt to get what we need in seaborn itself.

alualex:
[...]

PR #2413, opened 2021-01-03
Modernize categorical plotting and refactor stripplot

mwaskom:
This PR begins the process of refactoring the categorical plots; see https://github.com/mwaskom/seaborn/issues/2429 for context, including a list of enhancements and modified default features.

It implements much of the shared code supporting categorical plots (maintaining substantial, but not complete backwards compatibility) within the new core system, including some extensions within the core itself.

It also includes a re-implementation of `stripplot` and `catplot` with `kind="strip"`.


mwaskom:
Here are some notes on the initial refactoring effort (focusing on `stripplot`):

- It was possible to make all tests pass without touching them. That's heartening, and the quantitative coverage is good, but `stripplot` was mostly tested by directly interacting with its internals, rather than by calling it. The new internals are very well tested, but informal testing did reveal that some of the categorical-specific behavior was not initially correct but also did not fail the tests. So test coverage should be improved.

- The current approach to forcing categorical plots is to cast the categorical variable to a string and then to set a `category` dtype with the category order reflecting the rules for ordering seaborn categorical plots. Both elements are necessary for properly hooking into the matplotlib categorical units system — "categorical" units in matplotlib are really just "string data", but there's no explicit ordering mechanism. Initially I had thought the plan was going to be to write a custom converter object that used categorical locator/formatting rules even for numeric or datetime inputs, but that ran into some issues (https://github.com/matplotlib/matplotlib/issues/19229), and this approach seems to work well even though it feels a little odd.

- Currently categorical plots do not follow the active property cycle. The new code doesn't either, but we'll want to enable that. There are some notes in comments about subtle issues that will be encountered; also I continue to want a less ad-hoc approach to doing that than we have in current functions.

- There's an ad-hoc creation of new categorical variable in the case where only x/y is defined. This seems fine? But maybe should think about having that be an optional default in core.

- A big decision with the categorical plot refactor is whether to treat the order and structure of the matplotlib artists that get added as part of the public API. This is a weird part of `stripplot` — the number of artists depends on both `hue` and `dodge`, when it would be possible to draw just a single scatterplot. (This is kind of true for `swarmplot` too, but the artist separation there makes more sense. Supporting the old behavior did require a change in `_core` to allow `iter_data` to return an empty subset.

- We now have numeric hue mapping, but the legend represents every unique hue value. Also a problem in the distribution plots. Desperately needs a good solution (#2231).

- Previously, the categorical plots used different colors for different categorical levels in the absence of `hue`, depending on the presence of `palette` and `color`. I'm committed to breaking this and requiring explicit `hue`, but haven't done so here. There are some hacks in the new `stripplot` to support the old behavior that can eventually be removed. Question is whether this should get a deprecation cycle with an (always on) warning.

- Similarly, the old functions had a trick where defining `hue` and `color` (but not palette) would assign a dark or light palette seeded with `color`. I also want to break this, but have added a hack to enable  it here. Same question as above.

- The new function naturally supports `hue` with only `x` or `y` specified, which was a big annoyance with the old code. Great! But needs to be explicitly tested.

Other new functionality we want to add but haven't done here yet (this should have its own tracking issue):

- Control over the legend
- An option *not* to force categorical treatment of numeric or datetime data. I have a prototype branch where this seems to work pretty well with the current system. But it needs a good parameter name.

mwaskom:
Going to leave a few things unfinished, as this is getting unwieldy and their solution will need to be shared well across multiple categorical plots:

- docstring (other than API examples)
- legend
- color cycle integration

Issue #2414, opened 2021-01-05
The Freedman-Diaconis rule implementation should be NaN - agnostic


fabianegli:
The [Freedman-Diaconis rule](https://en.wikipedia.org/wiki/Freedman%E2%80%93Diaconis_rule) implementation as implemented in seaborn [seaborn/distributions.py](https://github.com/mwaskom/seaborn/blob/aad96f8d2e36ceceb82a42b69aa3a8f47ef7210d/seaborn/distributions.py#L2392) is not na-agnostic and fails if `a` contains `np.nan`.

I argue that the function should be made more robust and NaNs should be removed from any array submitted to the function since NaNs do not contribute to plots for which the function is used.

mwaskom:
Do you have an example of an actual issue that you have encoountered? That function is private, and is used internally by the (deprecated) `seaborn.distplot` function, which handles the na removal.

fabianegli:
I work with a project where I am bound to an old seaborn version (0.9.0) and there it popped up. I do not have the time now to reproduce the error and worked around the issue by removing nas before submitting the data to distplot.

Upon code inspection it is clear to me that while the function is private it still breaks for inputs containing NaNs. I am not familiar with the development plans, but the remove_na would not hurt in any way. Also, for me, adding it in a previous version fixed some issues.

mwaskom:
Your issue was addressed with https://github.com/mwaskom/seaborn/pull/1913.

PR #2416, opened 2021-01-05
make Freedman-Diaconis-rule NaN-agnostic

fabianegli:
This makes the function `_freedman_diaconis_bins` more robust.

As discussed in #2414 the arrays that are passed to the function in the seaborn code undergo are submitted to the `remove_na` function beforehand and the na removal is thereby mute.

@mwaskom Feel free to close the PR, if you disagree that this change adds value to the codebase.


fabianegli:
Closed by #1913

PR #2427, opened 2021-01-10
boxplot whis keyword docstring clarification

joooeey:
The docstring for the `whis` keyword in `boxplots` has an inaccurate description. That is, whiskers don't extend to 1.5*IQR, rather they extend to the furthest datum within 1.5*IQR. I propose a clearer wording.

Evidence for confusion caused by this inaccuracy can be found [here on Stackoverflow](https://stackoverflow.com/questions/49139299/whisker-is-defined-as-1-5-iqr-how-could-two-whikers-in-plot-from-python-seabor).

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2276,9 +2276,9 @@ boxplot.__doc__ = dedent("""\
         Size of the markers used to indicate outlier observations.
     {linewidth}
     whis : float, optional
-        Proportion of the IQR past the low and high quartiles to extend the
-        plot whiskers. Points outside this range will be identified as
-        outliers.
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks, that's a good edit. I tweaked the language a bit (I appreciate the proper latin, but "datapoint" is probably a bit more accessible) and will merge when CI is green.

joooeey:
> Thanks, that's a good edit. I tweaked the language a bit (I appreciate the proper latin, but "datapoint" is probably a bit more accessible) and will merge when CI is green.

I agree that datapoint is better. I didn't think of that.

Issue #2434, opened 2021-01-15
Support pandas Float64Dtype in boxenplot


afrendeiro:
Pandas 1.2.0 added a [Float64DType](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.2.0.html?highlight=float64dtype#experimental-nullable-data-types-for-float-data) type extension. Seaborn boxenplot fails with that input:
```python
data = pd.DataFrame(
   {"cont": np.random.random(20), "cat": np.random.choice(["a", "b"], 20)}
).convert_dtypes()

sns.boxplot(data=data, x='cat', y='cont')  #  okay
sns.boxenplot(data=data, x='cat', y='cont')  # error
```
At the same time Barplot and Boxplot don't fail. This is because boxenplot is at some point converting the input into a numpy array which makes the array dtype 'object'. That dtype fails when is input to `np.isfinite` [here](https://github.com/mwaskom/seaborn/blob/d334ea6b50dd8e5101c7b4c71b500cf01350499b/seaborn/categorical.py#L1843).

mwaskom:
This is something that should magically start to work once the work in https://github.com/mwaskom/seaborn/issues/2429 is complete.

afrendeiro:
Cool, sounds good! Feel free to close the PR or just recycle the test case.

mwaskom:
Closed by #2435

PR #2435, opened 2021-01-15
Support Float64Dtype in boxenplot, related to #2434

afrendeiro:
This PR solves the above issue.
Test added for when pandas >= 1.2.


mwaskom:
Thanks @afrendeiro as mentioned this will be mooted by upcoming work but it's always nice to get a PR with a test!

PR #2462, opened 2021-02-01
Fix multiple resolution when hue variable has no name

### LINKED ISSUE #2452: histplot ignoring multiple= when hue argument is a numpy array or list
e.g.

```
sns.histplot(x=tips["total_bill"], hue=tips["time"].to_numpy(), multiple="dodge")
```

It works fine when passing a pandas Series directly or referencing a column in an array.

mwaskom:
Fixes #2452

Also simplifies the logic that disabled the `multiple` parameter when no hue was assigned (which caused the issue originally), instead allowing the default visual attributes to vary with multiple even if it has no effect on the position of the distributions.

### DIFF
--- a/doc/releases/v0.12.0.txt
+++ b/doc/releases/v0.12.0.txt
@@ -26,7 +26,7 @@ v0.12.0 (Unreleased)
 
 - |Fix| In :func:`histplot` and :func:`kdeplot`, fixed a bug where the `alpha` parameter was ignored when `fill=False` (:pr:`2460`).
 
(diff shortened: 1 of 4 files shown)


doc/releases/v0.12.0.txt
```diff
@@ -18,13 +18,17 @@ v0.12.0 (Unreleased)
 
 - |Enhancement| In :func:`histplot`, added `stat="percent"` as an option for normalization such that bar heights sum to 100 (:pr:`2461`).
 
+- |Enhancement| |Fix| Improved integration with the matplotlib color cycle in most axes-level functions (:pr:`2449`).
```
mwaskom, line 21: This note is not relevant to this PR; I think I was tying off loose ends from a previous PR.

PR #2468, opened 2021-02-03
Fix faceted bivariate histogram normalization at displot

### LINKED ISSUE #2465: bivariate displot with kind="hist" ignores common_norm=False if hue is not assigned
Somehow similar to #2377, but for `kind='hist'` and with `common_norm=False` being ignored.

If I understand it well, passing `common_norm=False` to `sns.distplot` should produce plots with independent "z" scale. While this is the behaviour for kde plots, it seems to be ignored for bivariate histograms unless hue is also specified.

Test case:
```python
x, y = np.random.randn(2, 200)
z = [0] * 150 + [1] * 50
sns.displot(x=x, y=y, col=z, kind="hist", cbar=True, common_norm=False)
sns.displot(x=x, y=y, col=z, hue=z, kind="hist", cbar=True, common_norm=False)
```

Running seaborn version 0.11.1
(I tried to look for other (open or closed) related issues, hope not duplicating them)

mgab:
Fixes #2465

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -737,7 +737,7 @@ class _DistributionPlotter(VectorPlotter):
             )
             full_heights.append(sub_heights)
 
-        common_color_norm = "hue" not in self.variables or common_norm
+        common_color_norm = not set(self.variables) - {"x", "y"} or common_norm
 
         if pthresh is not None and common_color_norm:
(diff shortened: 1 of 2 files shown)


mwaskom:
Perfect! Thanks!

PR #2485, opened 2021-02-19
Replace version numbering interface

### LINKED ISSUE #2466: Move away from distutils for version checks
seaborn uses `distutils.version` to check dependency versions when relevant for optional features, but distutils is [being deprecated](https://www.python.org/dev/peps/pep-0632/).

I believe [packaging](https://pypi.org/project/packaging/) is the recommended path for adaptation.

vitorserpa:
As the `distutils.version` is deprecating, it was replaced
with the `packaging.version`. In order to give support for all
version numbering, according to PEP 440, parse was used.

Fixes #2466

### DIFF
--- a/seaborn/_core.py
+++ b/seaborn/_core.py
@@ -6,7 +6,7 @@ from collections import UserString
 from collections.abc import Iterable, Sequence, Mapping
 from numbers import Number
 from datetime import datetime
-from distutils.version import LooseVersion
+from packaging import version
 
 import numpy as np
 import pandas as pd
@@ -1251,7 +1251,7 @@ class VectorPlotter:
                     if scale is True:
                         set_scale("log")
                     else:
-                        if LooseVersion(mpl.__version__) >= "3.3":
+                        if version.parse(mpl.__version__) >= version.parse("3.3"):
                             set_scale("log", base=scale)
                         else:
                             set_scale("log", **{f"base{axis}": scale})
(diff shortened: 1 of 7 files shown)


mwaskom:
Thanks!

I guess `packaging` should go in the `install_requires` field in `setup.py`. It's a dependency of `pip` and it would be weird if someone installed seaborn without having pip installed, but probably not impossible? I haven't found any guidance on this.

setup.py
```diff
@@ -33,6 +33,7 @@
     'numpy>=1.16',
     'pandas>=0.24',
     'matplotlib>=3.0',
+    'packaging>=29.0',
```
mwaskom, line 36: Why this version floor?

mwaskom: (I am mostly asking if you have a link/reference to support the choice ... I don't know the right answer here).

vitorserpa: I was testing with the most recent `packaging` version. I am going to make some tests to evaluate if it can be used with previous versions.

mwaskom: I would expect the parse behavior to be highly stable across versions of packaging; I'm not sure it will be necessary to specify a minimum version.

mwaskom:
Note that pandas decided to vendor the `Version` class rather than include packaging as a dependency, which I would be inclined towards as well: https://github.com/pandas-dev/pandas/pull/41207

mwaskom:
Superseded and closed by #2659

PR #2489, opened 2021-02-25
Working "hello world" for non-notebook user

jameshfisher:
I was pretty confused by the intro docs. I wanted a "hello world" ASAP, but `sns.pairplot` was doing nothing. It just returns an object.

It turns out that behind the scenes, seaborn updates the pyplot context, and to see the plot, I needed to use `.show()`.

The current docs seem to assume the user is in a notebook environment. I do not believe this applies to everyone.

### DIFF
--- a/doc/introduction.ipynb
+++ b/doc/introduction.ipynb
@@ -52,6 +52,23 @@
    "cell_type": "raw",
    "metadata": {},
    "source": [
+    "To see your plot outside of a notebook environment, run:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
     "A few things have happened here. Let's go through them one by one:"
    ]
   },


mwaskom:
Hi @jameshfisher if you read just slightly farther you will see

"Behind the scenes, seaborn uses matplotlib to draw its plots. For interactive work, it’s recommended to use a Jupyter/IPython interface in matplotlib mode, or else you’ll have to call matplotlib.pyplot.show() when you want to see the plot."

Also the other place you get a "hello world" example you see this in code: http://seaborn.pydata.org/installing.html#quickstart

It's not strictly true that `plt.show` is always required outside of a notebook, because IPython's %matplotlib mode will also open up figures from the terminal interface as soon as they are created. (And other cases exist too, like if you just want to save the plot, not see it).

jameshfisher:
So, my mistake! When I first visited the seaborn site, the first thing I clicked on was "Tutorial" at the top. I couldn't find any mention of how to show a plot. Then when I went to open a PR, apparently I edited the introduction page, which is not the page I had visited. I feel a bit stupid for that.

(However, it does suggest another problem: perhaps the "introduction" should be more prominent in the website nav?)

Thanks for the `%matplotlib` pointer. I didn't know about that.

mwaskom:
The top level organization of the docs is fairly "flat" and doesn't all fit in the header bar. You can click the "Site" drop-down menu to get the pages/sections in order, or use the "contents" box on the homepage. I am likely going to transition the site to use the new pydata sphinx theme which assumes a bit more of a "nested" docs structure, so that may change.

The "tutorial" is really more of a user-guide ... it's didactic but not exactly oriented towards "starting from 0" as you (reasonably) suspected. The language is gradually changing, although I haven't technically renamed it yet because it will involve setting up a bunch of redirects, etc.

Issue #2409, opened 2020-12-30
Need better approach for handing log scales internally


mwaskom:
The `comp_data` property will produce log-scale x/y data when the corresponding axis is log scaled. That's useful because statistics will be computed in log space. But then the variables need to be converted back to the original domain before plotting.

There needs to be a general solution for this. e.g. the approach now in `lineplot` is something like:

```python
            for var in ["x", "y"]:
                if self._log_scaled(var):
                    for col in sub_data.filter(regex=f"^{var}"):
                        sub_data[col] = np.power(10, sub_data[col])
```

Complicating matters:

- The computation can add variables (e.g. `_min`, `_max` when error bars are produced)
- The distribution computations don't (currently) produce an output data with `x`, `y` columns (but they probably should

mwaskom:
Also we really should handle this sort of thing using the matplotlib `Transform` objects (which should usually or always -- need to check -- have inverses)

mwaskom:
Closing as there is WIP on this, it will only gradually role out to functions in the `seaborn` namespace, but it is not an external feature so it does not need to be tracked.

jlumpe:
> Also we really should handle this sort of thing using the matplotlib `Transform` objects (which should usually or always -- need to check -- have inverses)

Has there been any progress extending this to use arbitrary `Transform` objects?

PR #2504, opened 2021-03-12
Fix log scaling in distribution plots

### LINKED ISSUE #2502: displot(kind='ecdf',..., log_scale=True) not working
The following line of code gives an error:

```
sns.displot(kind='ecdf', data=df, x='col_1', log_scale=True)

UserWarning: Data has no positive values, and therefore cannot be log-scaled.
```

My data is all positive and kind='hist' or 'kde' works just fine.

mwaskom:
Fixes #2502 

This is a huge development footgun; see #2409 for thoughts on how this can be made automatic to reduce the risk of such bugs


mwaskom:
This needs a line in the release notes!

PR #2531, opened 2021-03-30
Add Read the Docs configuration

astrojuanlu:


mwaskom:
Hi, some context here would be helpful....

seaborn's doc build process is more complicated than a standard sphinx site and in the past it has proven not to work with readthedocs

astrojuanlu:
😱 I'm so sorry @mwaskom , I was doing some tests and prematurely opened a pull request to the wrong place.

> seaborn's doc build process is more complicated than a standard sphinx site and in the past it has proven not to work with readthedocs

Yes, that is what I was figuring out :) I don't want to add noise here, will close the pull request and keep working on my fork.

mwaskom:
Cool, no worries. Would be nice to have if you can get it working. If you haven't seen it, the best guide to the build workflow is the README in the doc/ directory.

Issue #2532, opened 2021-03-30
Generating full documentation requires creating the `example_thumbs` first


astrojuanlu:
Hi! I am [figuring out](https://github.com/mwaskom/seaborn/pull/2531) how to build seaborn docs on Read the Docs, [which indeed requires some care](https://github.com/mwaskom/seaborn/issues/958#issuecomment-253279718). I [almost have it](https://github.com/astrojuanlu/seaborn/pull/1), but I noticed a weird thing, perhaps because I'm doing something wrong.

The `doc/README.md` states that `NB_KERNEL="python3 make notebooks html` should work. However, this command produces this warning:

```
Running Sphinx v3.3.1                                                                                                                                                                                                               [108/1702]making output directory... done                                                                                        
/home/juanlu/Projects/RTD/migrate-rtd/seaborn/doc/conf.py:290: RemovedInSphinx40Warning: The app.add_javascript() is deprecated. Please use app.add_js_file() instead.
  app.add_javascript('copybutton.js')                                                                                  
/home/juanlu/Projects/RTD/migrate-rtd/seaborn/doc/conf.py:291: RemovedInSphinx40Warning: The app.add_stylesheet() is deprecated. Please use app.add_css_file() instead.
  app.add_stylesheet('style.css')                                                                                      
WARNING: html_static_path entry 'example_thumbs' does not exist
```

and then, even though the `example_thumbs` directory is next to the sources, the thumbnails are not present in `_build/html` (presumably because they were not there at the beginning of the build).

Then, running `make html` a second time fixes this, and all the thumbnails appear in `_build/html/static`.

_Edit: it's enough with creating the `example_thumbs` directory first, see discussion below_

Am I missing something? I assume this is either (a) something I'm doing wrong, (b) something you decided on purpose, despite being undocumented or (c) a necessary minor annoyance given the current workflow that did not make it to the README.

mwaskom:
This is probably just something I never noticed because `make clean` deletes everything *in* the directory but doesn't remove the directory itself, and it doesn't fail the doc build since sphinx only notifies at the warning level. Not sure what the best way to fix it would be.

astrojuanlu:
You're right (now I understood after [your clarification](https://github.com/mwaskom/seaborn/pull/2533#issuecomment-810592286)), it's as simple as creating the `example_thumbs` directory first. Editing the issue title.

PR #2533, opened 2021-03-30
Build documentation on Read the Docs

astrojuanlu:
Hi! This is a proposal to build the seaborn documentation on Read the Docs. Doing so has a bunch of interesting niceties, the best one for me being having the complete history of past seaborn versions (I've had to use Wayback Machine in the past to easily look for old documentation).

_Full disclaimer: I work as Developer Advocate at Read the Docs, with a mission to help scientific projects improve their documentation, and I am more than happy to hear feedback or complaints!_

This was proposed [a while back](https://github.com/mwaskom/seaborn/issues/958#issuecomment-253268750), but the build process requires some tweaks to make it work properly on Read the Docs. To that end, this pull request contains three things:

- A proper `.readthedocs.yaml` configuration
- A hack to build the notebooks if an environment variable is detected
- A little hack to create the thumbnails directory before the actual build starts (see #2532)

The end result can be seen here: https://astrojuanlu-seaborn.readthedocs.io/en/rtd/

Although nowadays parts of the seaborn documentation workflow, and in particular `tools/nb_to_doc.py`, could be rewritten using something like [nbsphinx](https://nbsphinx.readthedocs.io/) (which I personally use in other projects to create example galleries based on Jupyter notebooks), I wanted to keep things as simple as possible.

@mwaskom What do you think?

cc @ericholscher

_Edit: Simplified "hacks" after clarification, see discussion below_

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,9 @@
+version: 2
+python:
+  version: 3.8
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - all
+    - requirements: doc/requirements.txt
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -303,3 +303,13 @@ intersphinx_mapping = {
     'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
     'statsmodels': ('https://www.statsmodels.org/stable/', None)
 }
+
+# Build notebooks if on Read the Docs
+import subprocess
+if os.getenv("READTHEDOCS") == "True":
+    print("Processing notebooks, please wait...")
+    nb_output = subprocess.check_output(
+        ["make", "notebooks"],
+        env={"NB_KERNEL": "python3", "PATH": os.environ["PATH"]},
+    )
+    print(nb_output.decode(errors="ignore"))


mwaskom:
Glad you got it to work! However I'm not sure I find the argument for switching particularly compelling ... 

If the main argument is archival versions, that is now [basically solved](http://seaborn.pydata.org/archive.html). Are there other features I'm missing?

FWIW, last I looked into it, `nbsphinx` wasn't *exactly* a replacement for the way seaborn uses notebooks to write docs, though perhaps it's possible to fiddle with it to get what I want. But at this point, I'm the main and basically only person who needs to build the seaborn docs, and so I'm not thaaaat enthusiastic about replacing my workflow (which is arguably a sequences of hacks, but hacks that I know) with a new approach that I'd have to get accustomed to. It just seems like the time spent learning something new would be better spent on any of the other demands that already overwhelm the time I have to spent on seaborn.

But perhaps I'm missing something?

astrojuanlu:
You're right, it was as simple as having the directory before the build started! Just edited that part of the code.

> But at this point, I'm the main and basically only person who needs to build the seaborn docs...

I totally agree with you, it makes little sense to change the build code for little gain.

> If the main argument is archival versions, that is now basically solved. Are there other features I'm missing?

Nice, I hadn't seen the archive! There are other things you might find interesting:

- Since January this year, RTD allows you to [easily preview the documentation from pull requests](https://docs.readthedocs.io/en/stable/pull-requests.html). This means that you can check in the PR itself how the docstrings look like, instead of waiting until release time to find those. It integrates nicely with the rest of the GitHub Actions checks, you have more info in the link.
- RTD taking care of the versioning means that you can have a development/`master` version of the documentation. While you probably want to still redirect https://seaborn.pydata.org/ to the latest _released_ version, it can be useful to check out the current status of the development branch.
- All of this happens from the `mkwaskom/seaborn` repository, so you wouldn't need to push new versions to https://github.com/seaborn/seaborn.github.io/. In fact, [every time you tag a new version](https://docs.readthedocs.io/en/stable/automation-rules.html#activate-all-new-tags-and-branches-that-start-with-v-or-v), you would have its documentation online instantly. 
- And speaking about domains and redirections, RTD also gives you [user-defined redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html) (for example, if you change the name of a tutorial but want old links to work), custom domains (similar to how GitHub pages is already configured), [custom 404 pages](https://docs.readthedocs.io/en/stable/hosting.html#custom-not-found-404-pages) (instead of the default GitHub pages one)... And much more 😁

astrojuanlu:
Rebased and force-pushed on top of #2572.

EwoutH:
What might be interesting, is that there now is a much used [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/). NumPy, Pandas, SciPy, NetworkX and [many other projects](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/gallery.html) use it.

So that would add a possibility to switch to Read the Docs while maintaining the same styling.

mwaskom:
The concern was never about styling, just about workflow. I appreciate that RTD has some nice features that make collaborating on doc updates easier, but those just aren't particularly compelling for seaborn at this time. This PR is a useful POC and perhaps can be revisited in the future, but I think for now I've no appetite for moving away from an approach that currently satisfies all of my needs.

PR #2545, opened 2021-04-06
interquartile range errorbars

joAschauer:
When making categorical plots with median estimator, it would sometimes be useful to show the interquartile range as errorbars similar to the boxes in boxplots. 

What do you think of an option for the confidence interval parameter comparable to the "sd" option such as ci="iqr"?

```
import seaborn as sns
import numpy as np

ax = sns.pointplot(x="time", 
                   y="total_bill",
                   data=tips,
                   hue="smoker",
                   estimator=np.median, 
                   ci="iqr")
```

I'm not sure if this really is a valuable feature and it adds another option to the ci parameter which I think is not optimal. Feel free to close if you think it is unnecessary or can be achieved in another simple way without changes to the codebase (which I did not find out at least).

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1440,6 +1440,10 @@ class _CategoricalStatPlotter(_CategoricalPlotter):
                         sd = np.std(stat_data)
                         confint.append((estimate - sd, estimate + sd))
 
+                    if ci == "iqr":
+                        q25, q75 = np.percentile(stat_data, [25 ,75])
+                        confint.append((q25, q75))
+
                     else:
 
                         boots = bootstrap(stat_data, func=estimator,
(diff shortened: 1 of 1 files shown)


mwaskom:
Hi, see here: https://github.com/mwaskom/seaborn/pull/2407. This will become possible in the next release, but with a different API.

joAschauer:
Cool, I didn't see  #2407 and #2403. The upcoming API looks more reasonable than my hacky approach ;-) Will close accordingly.

mwaskom:
Yeah I had tended to reject suggestions like this because `ci="iqr"` felt very hacky. Transitioning to the new parameter will also be a little annoying, but the end result will be saner and also much more flexible.

joAschauer:
I think the new errorbar API looks super good. Thanks for your work!

PR #2548, opened 2021-04-12
DOC: Clarify how to derive dodge parameter value for aligning pointplot in stripplot

sappelhoff:
Coming from Twitter: https://twitter.com/michaelwaskom/status/1376595092014833667

This PR clarifies an example. Previously there was the magic number `.532`, which is now replaced by `.8 - .8/3` and an accompanying comment-explanation.

- 0.8 is the default width allotted to each category
- `0.8 - 0.8 /3` equals `.532` but will allow users to adjust the example to their need if they have different numbers of hues


mwaskom:
Thanks!

PR #2558, opened 2021-04-17
DOC: misc fixes.

Carreau:
Typos, non-existing parameters, and the fact that space before colon in
parameters is necessary for numpydoc to correctly parse the name and
type


mwaskom:
Thanks!

Issue #2539, opened 2021-04-02
Bug in `size_norm` extrapolation beyond data range


mwaskom:
This popped up on StackOverflow and looks like a bug:

```python
import numpy as np
import pandas as pd
import seaborn as sns
from matplotlib import pyplot as plt

np.random.seed(12345)
df = pd.DataFrame({'x': np.random.rand(60),
                   'y': np.random.rand(60),
                   'sz': np.append(np.random.randint(0, 2500, 30), np.random.randint(2500, 6000, 30))})

fig, axs = plt.subplots(ncols=2, figsize=(10, 4), sharex=True, sharey=True)
sns.scatterplot(data=df[:30], x='x', y='y', size='sz', sizes=(0.01, 200), size_norm=(0, 6000), ax=axs[0])
sns.scatterplot(data=df, x='x', y='y', size='sz', sizes=(0.01, 200), size_norm=(0, 6000), ax=axs[1])
```
![image](https://user-images.githubusercontent.com/315810/113413219-1f021e80-9388-11eb-9401-9e5b603fc1e7.png)

However I am having trouble reducing it down to a simpler example, e.g. this seems to work properly:

```python
x = np.arange(1, 10)
f, axs = plt.subplots(1, 2, sharex=True, sharey=True)
sns.scatterplot(x=x, y=x, size=x, sizes=(10, 200), size_norm=(1, 9), legend="full", ax=axs[0])
sns.scatterplot(x=x[2:7], y=x[2:7], size=x[2:7], sizes=(10, 200), size_norm=(1, 9), ax=axs[1])
assert np.allclose(axs[0].collections[0].get_sizes()[2:7], axs[1].collections[0].get_sizes())
```
![image](https://user-images.githubusercontent.com/315810/113413212-190c3d80-9388-11eb-803d-c651a8432daa.png)

If I had to guess at this point, the relevant code might be here: https://github.com/mwaskom/seaborn/blob/master/seaborn/_core.py#L338

lilisako:
@mwaskom 
I was able to reproduce this issue with this code which is similar to yours. 

```
x = np.arange(0, 3500, 80)
f, axs = plt.subplots(1, 2, sharex=True, sharey=True)
sns.scatterplot(x=x, y=x, size=x, sizes=(0.1, 200), size_norm=(1, 3500), ax=axs[0])
[...]

PR #2579, opened 2021-05-10
fix bug in size_norm extrapolation beyond data range 

lilisako:
## Original Issue
- https://github.com/mwaskom/seaborn/issues/2539

## Behavior 
```
import numpy as np
import pandas as pd
import seaborn as sns

np.random.seed(12345)
df = pd.DataFrame({'x': np.random.rand(60),
                   'y': np.random.rand(60),
                   'sz': np.append(np.random.randint(0, 2500, 30), np.random.randint(2500, 6000, 30))})

fig, axs = plt.subplots(ncols=2, figsize=(10, 4), sharex=True, sharey=True)
sns.scatterplot(data=df[:30], x='x', y='y', size='sz', sizes=(0.01, 200), size_norm=(0, 6000), ax=axs[0])
sns.scatterplot(data=df, x='x', y='y', size='sz', sizes=(0.01, 200), size_norm=(0, 6000), ax=axs[1])
```
- Before
スクリーンショット 2021-05-10 18 33 06

- After 
スクリーンショット 2021-05-10 18 31 40

### DIFF
--- a/seaborn/_core.py
+++ b/seaborn/_core.py
@@ -334,6 +334,8 @@ class SizeMapping(SemanticMapping):
             normed = self.norm(key)
             if np.ma.is_masked(normed):
                 normed = np.nan
+            if self.sizes is None:
+                self.sizes = self.plotter._default_size_range
             value = self.sizes[0] + normed * np.ptp(self.sizes)
         return value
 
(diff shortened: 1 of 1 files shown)


lilisako:
Thanks for the review & merging my commits! I'll make sure to create a new branch next time.

PR #2580, opened 2021-05-10
Fix size norm extrapolation

mwaskom:
Builds on #2579 with

- a test for the original bug
- the addition of a `size_range` attribute on the `SizeMapping` object, rather than accessing the `_default_size_range` attribute, which is regrettable and will ideally get refactored away somehow.


mwaskom:
This needs a release note!

PR #2598, opened 2021-06-16
Update README.md

tanujdhiman:
Add License to Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,3 +76,9 @@ Seaborn development takes place on Github: https://github.com/mwaskom/seaborn
 
 Please submit bugs that you encounter to the [issue tracker](https://github.com/mwaskom/seaborn/issues) with a reproducible example demonstrating the problem. Questions about usage are more at home on StackOverflow, where there is a [seaborn tag](https://stackoverflow.com/questions/tagged/seaborn).
 
+
+License
+-------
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks @tanujdhiman but as the license is packaged in the source distribution I don't think it needs to be duplicated in the README.

tanujdhiman:
Yes, @mwaskom.
Actually, Readme.md is for explaining things for a particular repository. So, if we have provide the License in this Readme then user just come to know this library issued by which License. 
Hope you understand !!
Thanks

mwaskom:
> So, if we have provide the License in this Readme then user just come to know this library issued by which License otherwise we have to find out the file in a large bunch.

That information is already present in both the Github UI and as a badge in the README, so anyone looking at the file in a rich text interface will see it. Those who are stubbornly committed to plain text will not have to look through a large bunch of files because the license file has the canonical name (much like README).

> Examples like - Tensorflow etc.

Libraries can choose to duplicate or not duplicate information. Tensorflow doesn't have the license badge. I also suspect that, as a project backed by a large corporation they feel motivated to boost the signal on their liberal open source license.

tanujdhiman:
Okay. Thanks

PR #2617, opened 2021-07-15
Fix various typos

luzpaz:
Found via `codespell -q 3 -L ans,hist,whis`


mwaskom (approved):

mwaskom:
Thanks! Don't see any reason why this would break the tests, but gonna let them run anyway before merging.

PR #2620, opened 2021-07-18
JointGrid reference lines

### LINKED ISSUE #2249: JointGrid could have axhline/axvline analogues that would plot on joint and marginal axes
Would need to thing about the right name, but probably a good idea.

stefmolin:
Closes #2249 &ndash; Adds new `JointGrid.refline()` method: 
```python
JointGrid.refline(self, *, x=None, y=None, joint=True, marginal=True, **line_kws)
```

Example usage:
```python
import seaborn as sns

penguins = sns.load_dataset('penguins')
g = sns.jointplot(data=penguins, x='bill_length_mm', y='bill_depth_mm')

# by default plot on both the marginal and joint axes
g.refline(y=16, linestyle='dashed', color='gray')

# only on joint
g.refline(x=45, marginal=False, linestyle='dotted', color='red')

# only on marginal
g.refline(x=55, joint=False, linestyle='dashed', color='blue')
```

Needs:

- [X] Test
- [X] Documentation in the JointGrid and FacetGrid API examples
- [X] Mention in release notes

### DIFF
--- a/examples/many_facets.py
+++ b/examples/many_facets.py
@@ -26,7 +26,7 @@ grid = sns.FacetGrid(df, col="walk", hue="walk", palette="tab20c",
                      col_wrap=4, height=1.5)
 
 # Draw a horizontal line to show the starting point
-grid.refline(y=0, linestyle=":", color=".5")
+grid.refline(y=0, linestyle=":")
 
 # Draw a line plot to show the trajectory of each random walk
 grid.map(plt.plot, "step", "position", marker="o")
--- a/examples/many_facets.py
+++ b/examples/many_facets.py
@@ -26,7 +26,7 @@ grid = sns.FacetGrid(df, col="walk", hue="walk", palette="tab20c",
                      col_wrap=4, height=1.5)
 
 # Draw a horizontal line to show the starting point
-grid.map(plt.axhline, y=0, ls=":", c=".5")
+grid.refline(y=0, linestyle=":", color=".5")
 
 # Draw a line plot to show the trajectory of each random walk
 grid.map(plt.plot, "step", "position", marker="o")
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1819,6 +1819,8 @@ class JointGrid(object):
         """
         if x is None and y is None:
             raise ValueError('Must provide ``x`` or ``y``')
+        elif x is not None and y is not None:
+            raise ValueError('Provide either ``x`` or ``y``, not both')
         elif x is not None:
             func = 'axvline'
             ax_marg = self.ax_marg_x
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1807,7 +1807,7 @@ class JointGrid(object):
         orient : "v" | "h"
             Orientation for reference line.
         joint, marginal : bools
-            Whether to add the reference line to the joint and/or marginal axes.
+            Whether to add the reference line to the joint/marginal axes.
         kwargs : key, value mappings
             Other keyword arguments are passed to :meth:`matplotlib.axes.Axes.axhline`
             or :meth:`matplotlib.axes.Axes.axvline` depending on ``orient``.
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1809,7 +1809,7 @@ class JointGrid(object):
             Value(s) to draw the line(s) at.
         joint, marginal : bools
             Whether to add the reference line(s) to the joint/marginal axes.
-        color : str or RGB tuple
+        color : :mod:`matplotlib color <matplotlib.colors>`
             Specifies the color of the reference line(s).
         linestyle : str
             Specifies the style of the reference line(s). Defaults to dashed.
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1827,6 +1827,8 @@ class JointGrid(object):
         else:
             raise ValueError(f'"{orient}" is not a supported value for ``orient``')
 
+        if not any((joint, marginal)):
+            raise ValueError('At least one of ``joint`` and ``marginal`` must be True')
         if joint:
             getattr(self.ax_joint, func)(value, **kwargs)
         if marginal:
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -957,6 +957,37 @@ class FacetGrid(Grid):
                 self.axes.flat[i].set_title(title, **kwargs)
         return self
 
+    def refline(self, *, x=None, y=None, color='.5', linestyle='--', **line_kws):
+        """Add a reference line(s) to each facet.
+
+        Parameters
+        ----------
+        x, y : numeric
+            Value(s) to draw the line(s) at.
+        color : :mod:`matplotlib color <matplotlib.colors>`
+            Specifies the color of the reference line(s).
+        linestyle : str
+            Specifies the style of the reference line(s). Defaults to dashed.
+        line_kws : key, value mappings
+            Other keyword arguments are passed to :meth:`matplotlib.axes.Axes.axvline`
+            when ``x`` is not None and :meth:`matplotlib.axes.Axes.axhline` when ``y``
+            is not None.
+
+        Returns
+        -------
+        :class:`FacetGrid` instance
+            Returns ``self`` for easy method chaining.
+
+        """
+        line_kws['color'] = color
+        line_kws['linestyle'] = linestyle
+
+        if x is not None:
+            self.map(plt.axvline, x=x, **line_kws)
+
+        if y is not None:
+            self.map(plt.axhline, y=y, **line_kws)
+
     # ------ Properties that are part of the public API and documented by Sphinx
 
     @property
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1797,6 +1797,42 @@ class JointGrid(object):
 
         return self
 
+    def plot_refline(self, *, value, orient, joint=True, marginal=True, **kwargs):
+        """Add a reference line to joint and/or marginal axes.
+
+        Parameters
+        ----------
+        value : numeric
+            Value to draw the line at.
+        orient : "v" | "h"
+            Orientation for reference line.
+        joint, marginal : bools
+            Whether to add the reference line to the joint and/or marginal axes.
+        kwargs : key, value mappings
+            Other keyword arguments are passed to :meth:`matplotlib.axes.Axes.axhline`
+            or :meth:`matplotlib.axes.Axes.axvline` depending on ``orient``.
+
+        Returns
+        -------
+        :class:`JointGrid` instance
+            Returns ``self`` for easy method chaining.
+
+        """
+        if orient == 'h':
+            func = 'axhline'
+            ax_marg = self.ax_marg_y
+        elif orient == 'v':
+            func = 'axvline'
+            ax_marg = self.ax_marg_x
+        else:
+            raise ValueError(f'"{orient}" is not a supported value for ``orient``')
+
+        if joint:
+            getattr(self.ax_joint, func)(value, **kwargs)
+        if marginal:
+            getattr(ax_marg, func)(value, **kwargs)
+        return self
+
     def set_axis_labels(self, xlabel="", ylabel="", **kwargs):
         """Set axis labels on the bivariate axes.
 
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1797,7 +1797,10 @@ class JointGrid(object):
 
         return self
 
-    def refline(self, *, x=None, y=None, joint=True, marginal=True, **line_kws):
+    def refline(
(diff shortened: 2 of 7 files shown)


seaborn/axisgrid.py
```diff
@@ -1797,6 +1797,47 @@ def plot_marginals(self, func, **kwargs):
...
+        Returns
+        -------
+        :class:`JointGrid` instance
+            Returns ``self`` for easy method chaining.
+
+        """
+        if x is None and y is None:
+            raise ValueError('Must provide ``x`` or ``y``')
+        elif x is not None and y is not None:
+            raise ValueError('Provide either ``x`` or ``y``, not both')
```
mwaskom, line 1823: Why have this limitation?

seaborn/axisgrid.py
```diff
@@ -1797,6 +1797,47 @@ def plot_marginals(self, func, **kwargs):
...
+            is not None.
+
+        Returns
+        -------
+        :class:`JointGrid` instance
+            Returns ``self`` for easy method chaining.
+
+        """
+        if x is None and y is None:
+            raise ValueError('Must provide ``x`` or ``y``')
```
mwaskom, line 1821: I'm not sure about this check (or the one below that requires either `joint` or `marginal` to be True).

These checks will make life difficult for any sort of code that wraps `JointGrid` and exposes the refline parameters. They'll need to modify the logic of their code based on the parameter values, instead of just passing them through to seaborn.

It can be helpful to provide feedback to users when surprising behavior would otherwise happen. But I think that the API here is simple enough that calling it with no value for x or y will be trivial to debug, so it's not worth the complexity it would add to downstream code. Do you disagree?

stefmolin: > These checks will make life difficult for any sort of code that wraps JointGrid and exposes the refline parameters. They'll need to modify the logic of their code based on the parameter values, instead of just passing them through to seaborn.

I don't fully understand what you mean by this. Are you trying to avoid raising any exceptions in the seaborn code for this method? What logic would they need to modify?

Regarding the check for either `joint` or `marginal` to be `True`, if both are `False` the effect of the method call would be nothing. This feels like it is failing silently.

Without the first check, the resulting error would actually be quite confusing due to the implementation (because we don't yet know if we should use `axhline()` or `axvline()`, same for which marginal axes):
```
   1833         if joint:
-> 1834             getattr(self.ax_joint, func)(value, **line_kws)
   1835         if marginal:
   1836             getattr(ax_marg, func)(value, **line_kws)

UnboundLocalError: local variable 'func' referenced before assignment
```

Of course, the implementation can be altered to first assign `func` to say `axhline()` and then allow matplotlib to raise an error when it receives `None`:
```
<ipython-input-8-0de983514f78> in <module>
      5 # by default plot on both the marginal and joint axes
      6 # g.refline(linestyle='dashed', color='gray')
----> 7 g.ax_joint.axvline(None)

~/seaborn_dev/lib/python3.7/site-packages/matplotlib/axes/_axes.py in axvline(self, x, ymin, ymax, **kwargs)
    785         # Strip away the units for comparison with non-unitized bounds.
    786         xx, = self._process_unit_info([("x", x)], kwargs)
--> 787         scalex = (xx < xmin) or (xx > xmax)
    788 
    789         trans = self.get_xaxis_transform(which='grid')

TypeError: '>' not supported between instances of 'float' and 'NoneType'
```
I'm fine with changing the implementation and removing that check, but I don't think the check should be removed without changing the implementation. Do you prefer any exceptions to come from matplotlib?

seaborn/axisgrid.py
stefmolin: I was aligning this with the `axhline()` / `axvline()` behavior i.e. 1 call for each line you want. How did you initially envision this?

mwaskom: I don't see any reason not to allow `g.refline(x=..., y=...)`.

seaborn/axisgrid.py
mwaskom: My point is, imagine you are writing a function that uses `JointGrid`. You want to expose `refx`, `refy` parameters, defaulting to `None`. With your exceptions, that function would need to add in some logic

```python
def wrapper(..., refx=None, refy=None, ...):
    g = JointGrid(...)
    ...
    if refx is not None or refy is not None:
        g.refline(refx, refy)
    ...
```
Not the most complicated code in the world, but a little annoying.

> Do you prefer any exceptions to come from matplotlib?

No, I think that calling `ax{h,v}line` should only happen when `x` or `y` are not `None`, I just think that logic should be internal to the method.

I'm not really concerned about silently doing nothing if both `x` and `y` are `None` (or with `marginal=False, joint=False`). I don't think that's an error; you're doing what the user asked for.

seaborn/axisgrid.py
stefmolin: I'll make an update then.

seaborn/axisgrid.py
stefmolin: Ok, that's helpful.  I was developing this for more direct use, but since we often use wrapper functions in seaborn it makes sense that we may want to bubble up this functionality in the future &ndash; in which case, removing the checks makes it easier. I'll remove the them to align with the seaborn philosophy 🙂

stefmolin:
Now both `joint = marginal = False` and `x = y = None` don't raise any errors and don't do anything:
```python
# x and y missing -- nothing happens
g.refline(linestyle='dashed', color='gray')

# marginal and joint are False -- nothing happens
g.refline(x=45, y=16, marginal=False, joint=False, linestyle='dashed', color='gray')
```

Passing both `x` and `y` will now plot both:
```python
g.refline(x=45, y=16, linestyle='dashed', color='gray')
```

mwaskom:
Great! So, three thoughts now:

1. Default visual properties. `axhline` and `axvline` default to the first color in the color cycle, but I change the color to some grayscale value 100% of the time that I use them. I also usually make them dashed. This is because they typically represent annotations, not data, and I want to visually distinguish them from the other marks on the plot. (This is something the [matplotlib devs agree with](https://github.com/matplotlib/matplotlib/issues/14331)). I think it would be good to encode defaults like that in the `refline` signature, e.g. `color=".5", linestyle="--"`. Does that sound good to you?

2. It occurs to me that this functionality would be pretty useful in `FacetGrid` too. The implementation will need to be a little different since there's no concept of joint/marginal there, so they can't share the same method, but in a sense it's even simpler. Do you agree and would you be up for adding that as part of this PR?

3. I edited the PR summary to include a short TODO list of additional tasks that will been to be accomplished before this can be merged. Please let me know if you need any guidance on implementing them.

stefmolin:
1. I agree with the default style for reference lines &ndash; it's typically what I do as well. To confirm, you want to add `color` and `linestyle` with the defaults above to the signature? Since these are also technically `line_kws`, we could also leave that default logic internal to the `refline()` method without changing the signature. Which do you prefer?
2. Sure, I'll work on adding this to `FacetGrid`. I would like to get your thoughts on that first though. When `sharex = sharey = True`, it's pretty straightforward to add the reference lines, but when they aren't equal, the line may not show up on all subplots. Do we want to allow for multiple x's/y's (one per subplot)?

3. I'll take a look and let you know if I have any questions.

mwaskom:
> I agree with the default style for reference lines – it's typically what I do as well. To confirm, you want to add color and linestyle with the defaults above to the signature? Since these are also technically line_kws, we could also leave that default logic internal to the refline() method without changing the signature. Which do you prefer?

Given that the defaults will differ a bit from the analagous matplotlib functions, I think it would be good to have them explicit in the signature. (But the code will be cleaner if you inject them into the `line_kws` dict before plotting.

> Sure, I'll work on adding this to FacetGrid. I would like to get your thoughts on that first though. When sharex = sharey = True, it's pretty straightforward to add the reference lines, but when they aren't equal, the line may not show up on all subplots. Do we want to allow for multiple x's/y's (one per subplot)?

I don't think allowing multiple x/y's is worth the code complexity / thinking about what that API would look like. Let's have `FacetGrid.refline` with the same x/y on every facet for now.

stefmolin:
I've added the `color` and `linestyle` parameters to the method signature. Before I move on to the `FacetGrid` though, could you give me some feedback on the tests and example for `JointGrid.refline()`?

seaborn/axisgrid.py
```diff
@@ -1797,6 +1797,50 @@ def plot_marginals(self, func, **kwargs):
...
+    ):
+        """Add a reference line(s) to joint and/or marginal axes.
+
+        Parameters
+        ----------
+        x, y : numeric
+            Value(s) to draw the line(s) at.
+        joint, marginal : bools
+            Whether to add the reference line(s) to the joint/marginal axes.
+        color : str or RGB tuple
```
mwaskom, line 1812:
```suggestion
color : :mod:`matplotlib color <matplotlib.colors>`
```

mwaskom (commented):
One tiny nit in the docstring, otherwise looks perfect!

mwaskom:
Everything looks great so far!

Once implemented in `FacetGrid`, it would be good to update these two examples to use `g.refline` instead of mapping `plt.axhline`: 

- http://seaborn.pydata.org/examples/many_facets.html
- http://seaborn.pydata.org/examples/kde_ridgeplot.html

stefmolin:
Can you confirm that the kde_ridgeplot example is running correctly for you? When I run it (without switching to `refline()`), the y-axis labels ("Density") show up for each subplot, only by changing `g.set(yticks=[])` to `g.set(yticks=[], ylabel='')` do I get the result in the docs.

mwaskom:
Ah, interesting! That does seem to be broken on master (not on v0.11.1).

mwaskom:
Probably the result of this: https://github.com/mwaskom/seaborn/pull/2583. "Improvement", indeed...

mwaskom:
OK thinking about this a bit more, I think that is the desired result of #2583. Using a more standard case:

```python
sns.FacetGrid(tips, col="time").map(sns.kdeplot, "total_bill")
```

v0.11.1:

image

master:

image

I think master is right (or better) here, `FacetGrid` shouldn't remove useful axis labels (from the "exterior" axes) unless you ask it too.

It produces an undesirable result in the ridgeplot example because lots of additional tweaking has been done to the plot, but that's ok, all that needs to be done is to add `ylabel=""` where `g.set()` is called.

stefmolin:
For the release notes, how should this be tagged? I took a look at the release notes for v0.11.0, which also has some new entries for new methods. Those are tagged "Feature", but since this PR is tagged "Enhancement", I figured I would double check.

mwaskom:
Hm good question. Actually there's no feature label on GitHub 🙃.

I think the feature/enhancement boundary can be a little blurry (is this adding a wholly new feature, or enhancing the existing JointGrid/FacetGrid functionality?) but I think Feature probably fits best.

stefmolin:
Agreed. I'll push the rest of the changes up shortly.

seaborn/axisgrid.py
```diff
@@ -957,6 +957,37 @@ def set_titles(self, template=None, row_template=None, col_template=None,
...
+        """Add a reference line(s) to each facet.
+
+        Parameters
+        ----------
+        x, y : numeric
+            Value(s) to draw the line(s) at.
+        color : :mod:`matplotlib color <matplotlib.colors>`
+            Specifies the color of the reference line(s).
+        linestyle : str
+            Specifies the style of the reference line(s). Defaults to dashed.
```
mwaskom, line 970: Small thing, but I think it's best not to duplicate information about defaults in the parameter description. It's too easy for things to fall out of sync if the defaults change for some reason down the road.

examples/many_facets.py
```diff
@@ -26,7 +26,7 @@
                      col_wrap=4, height=1.5)
 
 # Draw a horizontal line to show the starting point
-grid.map(plt.axhline, y=0, ls=":", c=".5")
+grid.refline(y=0, linestyle=":", color=".5")
```
mwaskom, line 29: Nit: can probably omit `color=".5"` since that's the default.

seaborn/axisgrid.py
```diff
@@ -957,6 +957,37 @@ def set_titles(self, template=None, row_template=None, col_template=None,
...
 
+    def refline(self, *, x=None, y=None, color='.5', linestyle='--', **line_kws):
+        """Add a reference line(s) to each facet.
+
+        Parameters
+        ----------
+        x, y : numeric
+            Value(s) to draw the line(s) at.
+        color : :mod:`matplotlib color <matplotlib.colors>`
+            Specifies the color of the reference line(s).
```
mwaskom, line 968: Should we document that passing `color=None` will use hue mapping?

mwaskom (commented):
A couple of tiny comments, otherwise I think this looks great!

seaborn/axisgrid.py
mwaskom: (Perhaps both here and with a comment in the ridgeplot example — it took me a minute to figure out what was going on there (but cool that it works!))

mwaskom (approved):

mwaskom:
Thanks @stefmolin!

PR #2651, opened 2021-08-31
allowing hue_norm to be subclass of Normalize object

nitaifingerhut:
Explain with an example:
user wants to use matplotlib.colors.LogNorm and not the standard Normalize.

currently hue_norm only supports instance of Normalize object,
but it can work just fine with any subclass of it which implements the __call__ method (such as LogNorm)

### DIFF
--- a/seaborn/_core.py
+++ b/seaborn/_core.py
@@ -244,7 +244,7 @@ class HueMapping(SemanticMapping):
                 norm = mpl.colors.Normalize()
             elif isinstance(norm, tuple):
                 norm = mpl.colors.Normalize(*norm)
-            elif not issubclass(norm, mpl.colors.Normalize):
+            elif not issubclass(type(norm), mpl.colors.Normalize):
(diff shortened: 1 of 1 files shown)


nitaifingerhut:
usage example:
```
from matplotlib.colors import LogNorm
sns.lineplot(data=df, x="x", y="y", hue="z", hue_norm=LogNorm())
```

mwaskom:
Hi, I'm not sure I understand the proposed change here?

```python
isinstance(mpl.colors.LogNorm(), mpl.colors.Normalize)
```
```
True
```

Instances of subclasses of a type are instances of that type.

In fact the [API docs for lineplot](https://seaborn.pydata.org/generated/seaborn.lineplot.html) have an example using `LogNorm` (I can't link to individual examples, but ctrl-f for "LogNorm" and you'll see it.

nitaifingerhut:
yah yah 
you are right
it's an old code mistakingly got here...
closing

PR #2659, opened 2021-09-09
Vendor the Version package, use it instead of distutils

dannysepler:
Closes https://github.com/mwaskom/seaborn/issues/2466. Unlike https://github.com/mwaskom/seaborn/pull/2485, this chooses to make a vendored Version class much like Pandas did here https://github.com/pandas-dev/pandas/pull/41207.

I took out a bit of what was in the two `packaging` files since it seemed like seaborn only needs to use the `Version` class, but happy to put it back in!

### DIFF
--- a/seaborn/external/version.py
+++ b/seaborn/external/version.py
@@ -1,0 +1,461 @@
+"""Extract reference documentation from the pypa/packaging source tree.
+
+In the process of copying, some unused methods / classes were removed.
+These include:
+
+- parse()
+- anything involving LegacyVersion
+
+This software is made available under the terms of *either* of the licenses
+found in LICENSE.APACHE or LICENSE.BSD. Contributions to this software is made
+under the terms of *both* these licenses.
+
+Vendored from:
+- https://github.com/pypa/packaging/
+- commit ba07d8287b4554754ac7178d177033ea3f75d489 (09/09/2021)
+"""
+
+
+# This file is dual licensed under the terms of the Apache License, Version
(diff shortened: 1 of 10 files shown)


mwaskom:
Awesome! Yeah, I think this is the better path forward here.

mwaskom:
> it seemed like seaborn only needs to use the Version class

yep I think including the minimal amount possible to make this work is the right idea.

seaborn/external/version.py
```diff
@@ -0,0 +1,461 @@
...
+
+In the process of copying, some unused methods / classes were removed.
+These include:
+
+- parse()
+- anything involving LegacyVersion
+
+This software is made available under the terms of *either* of the licenses
+found in LICENSE.APACHE or LICENSE.BSD. Contributions to this software is made
+under the terms of *both* these licenses.
```
mwaskom, line 11: Can you put a copy of the [BSD version](https://github.com/pypa/packaging/blob/main/LICENSE.BSD) of the license in `licenses/PACKAGING_LICENSE.md`?

mwaskom (changes requested):
Thanks! We're actually inside the window to bump the minimum version and strip out a lot of these conditionals, but let's not do that here. Only thing is to add the full text of the license.

dannysepler:
ready for review!

mwaskom (approved):

mwaskom:
Thanks @dannysepler!

Issue #1844, opened 2019-09-24
Make optimal leaf ordering for linkage an option for clustermap


votti:
Enhancement suggestion: Adding optimal leaf ordering as an option for `clustermap`

Background:
Many users are unaware that the leaf ordering in hierarchical clustering without any explicit leaf ordering is fairly arbitrary, as there are 2^(n-1) options to order a dendrogram. This often causes pattern across subtrees to be arbitrarily broken up, which can be a sub optimal data representation.
The solution is to optimize the leaf ordering, such that the distance between successive leafs is as small as possible.

This has been recognized in the literature (https://academic.oup.com/bioinformatics/article/17/suppl_1/S22/261423) and also addressed in scipy (from v1.0, https://github.com/scipy/scipy/issues/5238, https://github.com/scipy/scipy/pull/7501) based on the polo implementation (https://github.com/adrianveres/polo). As far as I have experienced or could find out, this is not considered in the currently default used `fastcluster` package.

Apart from the additional computational cost, there seems to be no real downside of enabling optimal ordering. It just improves a previously random order, making clustered heatmaps smoother and the visualization more meaningful. I experienced several times in real world datasets that this can reveal significantly more structure in heatmaps.

Example: https://github.com/adrianveres/polo/blob/master/data/demo.png

The additional computational cost is <1s for less than 256 leafs (https://github.com/adrianveres/polo/blob/master/data/bench.png), arguably not an issue for small visualizations.

While it is possible to just pre-calculate the optimal ordering linkage via `scipy.cluster.hierarchy.linkage` and the `optimal_ordering` argument separately, a lot of users are likely not even aware that optimal leaf ordering is something to consider. Thus, I would really argue that the benefit of having this more optimal clustering more easily available, e.g. via an 'order_dendrogram_optimal' option, would be beneficial for many and would justify the additional complexity of having an additional optional argument.

Thanks for your hard work and for considering this suggestion.

mwaskom:
> Apart from the additional computational cost

How much is this, in practice?

votti:
@Benchmark:
As the scipy implementation is a direct copy of the `polo` C-code, I think their benchmark should still apply:
![image](https://user-images.githubusercontent.com/3743016/65546294-829bc800-df17-11e9-9e36-c500906026d6.png)
 https://github.com/adrianveres/polo/blob/master/data/bench.png
-> <=1s for dendrograms with ~2**9 = 512 leafs

[...]

PR #2678, opened 2021-10-14
add dendrogram_kws for clustermap

lukebfunk:
When using clustermap, I have sometimes desired the ability for fine-grained control of the dendrogram properties, e.g., leaf ordering. This is achieved here by introducing a dendrogram_kws argument which is passed to scipy.cluster.hierarchy.dendrogram. Now arguments such as distance_sort and count_sort are accessible.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -561,7 +561,8 @@ def heatmap(
 class _DendrogramPlotter(object):
     """Object for drawing tree of similarities between data rows/columns"""
 
-    def __init__(self, data, linkage, metric, method, axis, label, rotate):
+    def __init__(self, data, linkage, metric, method, axis, label, rotate,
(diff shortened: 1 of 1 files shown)


mwaskom:
Ping @lukebfunk do you think you'll be able to get this over the line? Let me know if my suggestion about a test is unclear.

mwaskom:
Closing as stalled. I did look into this a bit to see if it would be easy to land with a small test, but I actually think the API decision here is more complicated Most of the dendrogram kwargs seem irrelevant in the case of a clustermap (or likely to open other worm cans) so I wonder if enabling what is desired here would be easier with a single kwarg, e.g. `tree_order`, which could also be the basis for a solution to #1844 (though they seem to be implemented in different parts of the scipy stack?)

PR #2679, opened 2021-10-15
Option for kdeplot to interpret `levels` as densities.

pSpitzner:
The main motivation for this PR is that I wanted to improve the clarity of a histplot by adding lines from a kdeplot.
Those lines should be at know value that aproximately represent the histogram height (or, rather, the probability density).

Currently, the `levels` argument in `kdeplot` is interpreted as iso-proportions.
This PR proposes a `levels_as_densities` kwarg. This allows to specify `levels`
as iso-densities instead of iso-proportions.

This has  two advantages:
- lines (e.g. of matching styles) can be compared across different kdeplots.
- when enabeling a colorbar for kdeplot, the values can be compared to histplot. In principle, the two cbars can be combined.

I illustrated this in the layered_bivariate example, see below.

sample

### DIFF
--- a/examples/layered_bivariate_plot.py
+++ b/examples/layered_bivariate_plot.py
@@ -19,5 +19,8 @@ x, y = rng.multivariate_normal(mean, cov, n).T
 # Draw a combo histogram and scatterplot with density contours
 f, ax = plt.subplots(figsize=(6, 6))
 sns.scatterplot(x=x, y=y, s=5, color=".15")
-sns.histplot(x=x, y=y, bins=50, pthresh=.1, cmap="mako")
-sns.kdeplot(x=x, y=y, levels=5, color="w", linewidths=1)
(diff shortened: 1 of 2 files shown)


mwaskom:
I'm probably -1 on this change because `kdeplot` already has too many parameters and I find it difficult to understand why one would want to specify or report raw density values. You've demonstrated a plot that could be made with this change, but what additional understanding of the data does the feature unlock?

In any case, a few comments on the proposed API:

- The parameter name is too long
- Why limit this just to specified `levels`, rather than letting the default evenly spaced levels be chosen in density space?
- I've learned that it's often a mistake to expand flexibility by adding a boolean parameter, unless there's a good case to be made that there are no other possible arguments.

mwaskom:
Thanks @pSpitzner but I'm going to close this for now.

Issue #2663, opened 2021-09-17
matplotlib 3.5.0b1 breaks test suite


mwaskom:
I'm pretty sure these failures are all related to things that happen as part of tests but not normal operations and so this isn't an urgent problem, but they should get addressed before the final release drops and breaks CI.

<details>

```python-traceback
$ pytest seaborn -q
............ssssss..................................................................................... [  8%]
...............................................................................F.FF.................... [ 17%]
.............................................................................s......................... [ 25%]
............s................................F......................................................... [ 34%]
....................................................................................................... [ 43%]
.........................sss........................................................................s.. [ 51%]
...........F..FFFFFFF..F............................................................................... [ 60%]
........................................................................................F.............. [ 68%]
.....................................ss................................................................ [ 77%]
........................................................ss.ss.........ssss...................s......... [ 86%]
....................................................................................................... [ 94%]
.......s........................................................                                        [100%]
================================================== FAILURES ===================================================
________________________________________ TestBoxPlotter.test_axes_data ________________________________________

self = <seaborn.tests.test_categorical.TestBoxPlotter object at 0x7f9dce7f2d60>

    def test_axes_data(self):

        ax = cat.boxplot(x="g", y="y", data=self.df)
>       assert len(ax.artists) == 3
E       AssertionError: assert 0 == 3
E        +  where 0 = len(<Axes.ArtistList of 0 artists>)
E        +    where <Axes.ArtistList of 0 artists> = <AxesSubplot:xlabel='g', ylabel='y'>.artists

seaborn/tests/test_categorical.py:858: AssertionError
___________________________________ TestBoxPlotter.test_draw_missing_boxes ____________________________________

self = <seaborn.tests.test_categorical.TestBoxPlotter object at 0x7f9dc93d8eb0>

    def test_draw_missing_boxes(self):

        ax = cat.boxplot(x="g", y="y", data=self.df,
                         order=["a", "b", "c", "d"])
>       assert len(ax.artists) == 3
E       AssertionError: assert 0 == 3
E        +  where 0 = len(<Axes.ArtistList of 0 artists>)
E        +    where <Axes.ArtistList of 0 artists> = <AxesSubplot:xlabel='g', ylabel='y'>.artists

seaborn/tests/test_categorical.py:887: AssertionError
______________________________________ TestBoxPlotter.test_missing_data _______________________________________

self = <seaborn.tests.test_categorical.TestBoxPlotter object at 0x7f9dceec4b50>

[...]

PR #2693, opened 2021-11-04
Fix box(en)plot colour tests

QuLogic:
As was noted in https://github.com/mwaskom/seaborn/issues/2663#issuecomment-937359794, the `test_box_colors` tests are broken because `zip` only iterates to the shortest input.

For `boxplot`, there are in fact 3 copies of the palette not 2 for the second test.

For `boxenplot`, it doesn't add `Patch` at all but `PatchCollection`, as can be seen in the related tests.

### DIFF
--- a/seaborn/tests/test_categorical.py
+++ b/seaborn/tests/test_categorical.py
@@ -5,7 +5,7 @@ import numpy as np
 import pandas as pd
 import matplotlib as mpl
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks @QuLogic, I think I determined that the short-circuit zip was actually intentional, or fine at least, since it was excluding the artists that get added to support the legend. But this makes it more clean.

PR #2699, opened 2021-11-15
setup.py: Update runtime error to correct minimal Python version

EwoutH:
Small nitpick after #2396: Also update the runtime error to list Python 3.7 as minimal version


mwaskom:
Thanks!

mwaskom:
Merging because broken docs build is unrelated.

Issue #2093, opened 2020-05-23
Consider adopting new PyData sphinx theme


mwaskom:
https://github.com/pandas-dev/pydata-sphinx-theme

mwaskom:
Blocking issue upstream: https://github.com/pandas-dev/pydata-sphinx-theme/issues/146

Anselmoo:
Dear @mwaskom ,

if you consider updating the **doc theme**, can you please take look on the **search engine**.

For not listed items, the docs are searching forever.

For example: Searching for **title** on [seaborn](https://seaborn.pydata.org/search.html?q=title&check_keywords=yes&area=default)

1. will not stop searching after 10min or more
2. will not show any results

Maybe it's caused to an older version of `mkdocs`??

mwaskom:
Theme is unrelated to search engine. I don't know what's happening there. I also don't know what `mkdocs` is. seaborn uses sphinx.

PR #2700, opened 2021-11-15
Loosen doc/requirements.txt

EwoutH:
Loosen the package requirements in `doc/requirements.txt`.
 - `docutils<0.18` allows docutils 0.17.1 and maybe a future 0.17.2
 - sphinx and sphinx_bootstrap_theme now only have minimal versions

This enables the docs to be build with Python 3.10.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,6 +1,6 @@
-docutils<=0.17  # https://sourceforge.net/p/docutils/bugs/431/
-sphinx==3.3.1
-sphinx_bootstrap_theme==0.7.1
+docutils<0.18  # https://sourceforge.net/p/docutils/bugs/431/
+sphinx>=3.3.1
+sphinx_bootstrap_theme>=0.7.1
 numpydoc
 nbconvert
 ipykernel


mwaskom:
Hi, is there a problem you had that motivated this change? The Sphinx and sphinx-bootstrap versions were pinned for a reason, I believe.

EwoutH:
Thanks for the very quick reply! I'm trying to build the docs with Python 3.10, to revert #2685. To do that, at least sphinx [v4.3.0](https://github.com/sphinx-doc/sphinx/releases/tag/v4.3.0) is needed.

On my branch, the docs build with the loosend requirements passes with both Python [3.9](https://github.com/EwoutH/seaborn/runs/4211202434) and [3.10](https://github.com/EwoutH/seaborn/runs/4211289900).

mwaskom:
Yes but, if I remember correctly, they look wrong with more recent sphinxes.

EwoutH:
I did a [test run](https://github.com/EwoutH/seaborn/actions/runs/1462658008) in which I generated all the docs and uploaded them as artifact. When viewing them in a .rst viewer they look fine.

In #2393 the `doc/requirements.txt` [was updated](https://github.com/mwaskom/seaborn/commit/a586ce8ca8cc30c14c9169b087a106fe14b70225#diff-246970f36ae7ff82cc458ecc03e34c76cd67f7b50a0e4899b12c31c283b26fcf), pinning `sphinx` and `sphinx_bootstrap_theme` at seemingly arbitrary version, and removing the comment that is messed up the css. Might that issue be fixed in later versions?

mwaskom:
I think you're right that whatever specific issue I had in mind is no longer a problem. But still, that doesn't answer whether is there a reason for this update? Pinning the versions prevents builds from suddenly breaking and causing problems, or from having something upstream cause a change in outputs and missing it at the version when it happened, making it harder to debug.

Also, for some reason, your artifact build is missing plots on the introduction page and for the first three sections of the tutorial.

mwaskom:
Thanks @EwoutH but I don't think it's the right time to update these versions since there apparently is some issue preventing a full build of the docs. I'd like to move over to the pydata theme anyway (#2093) and don't think it's worth the effort to spend a lot of time making the current version of the docs work on newer versions of sphinx. I did pull your tweak to the docutils version requirement into #2704 (odd that docutils doesn't use `.0` for micro releases; I didn't even realize they did them).

Issue #2652, opened 2021-08-31
Performance regression with duplicate index 0.11.1->0.11.2


zerothi:
I am using `histplot` for plotting simple histogram plots for medium sized data-sizes.

I was using `histplot` for histogram determinations in my dataset using 0.11.1 and all worked very nicely and fast.

Then upon upgrading to 0.11.2 it suddenly never finishes and I am stuck in plotting stuff...

I have a plotting script looking like this:
```python
plt.figure()
sns.histplot(data=df, x="AllocCores", binwidth=1, discrete=True, log_scale=(False, True))
```

The data-frame has 789412 entries and the span of the data `df["AllocCores"].min / max` is `1, 1200`. So 1200 bins which I agree is very large but required in our analysis plots ;)

Using **0.11.1** I get this:
```bash
$> time python script.py
seaborn version = 0.11.1
Starting to plot...
Choosing binwidth = 120
Choosing binwidth = 1070
Choosing binwidth = 16
Choosing binwidth = 16
python script.py  40.34s user 0.83s system 155% cpu 26.479 total
```
note that this is plotting 16 files in less than a minute.

Using **0.11.2** I get this:
```bash
$> time python script.py
seaborn version = 0.11.2
Starting to plot...
```
it basically never ends... I have waited more than 10 minutes, but it isn't even done with the 1st plot.

When killing it `Ctrl+^C` I get this stack-trace:
```bash
  File "script.py", line 96, in <module>
    plot_alloccores(df)
  File ".../script_lib.py", line 98, in plot_alloccores
    sns.histplot(data=df, x="AllocCores", binwidth=1, discrete=True, log_scale=(False, True))
  File "/opt/gnu/9.3.0/python/3.8.5/lib/python3.8/site-packages/seaborn/distributions.py", line 1462, in histplot
    p.plot_univariate_histogram(
  File "/opt/gnu/9.3.0/python/3.8.5/lib/python3.8/site-packages/seaborn/distributions.py", line 428, in plot_univariate_histogram
    for sub_vars, sub_data in self.iter_data("hue", from_comp_data=True):
  File "/opt/gnu/9.3.0/python/3.8.5/lib/python3.8/site-packages/seaborn/_core.py", line 983, in iter_data
    data = self.comp_data
  File "/opt/gnu/9.3.0/python/3.8.5/lib/python3.8/site-packages/seaborn/_core.py", line 1057, in comp_data
[...]

PR #2714, opened 2021-12-17
fix error when using lineplot and duplicate index

### LINKED ISSUE #2515: Using dataframe with duplicate index raises ValueError: cannot reindex from a duplicate axis
Tested with seaborn's latest dev version, pandas 1.2.3.

Dataframes can have indices with repeated entries. For example, `df.explode` which converts lists in cells into separate rows, keeps the original indices.  Some seaborn functions generate an error `ValueError: cannot reindex from a duplicate axis` with such dataframes.  `lineplot`, `stripplot`, `swarmplot` give such error.  `pointplot`, `barplot` and `boxplot` work without problem.

Calling `.reset_index()` makes the problem go away.
```
import` pandas as pd
import seaborn as sns

df = pd.DataFrame({'x': [1, 2],
                   'val': [[3, 4, 5], [6, 7]] })
sns.stripplot(data=df.explode('val'), x='x', y='val')
```
In seaborn 0.11.1 this same problem doesn't seem to exist. 

PS: `sns.violinplot(data=df.explode('val').reset_index(), x='x', y='val')` gives a complete different error, although something similar to `boxplot` would be expected (`TypeError: No loop matching the specified signature and casting was found for ufunc add`, during a call to `gaussian_kde`, numpy version `1.20.1`, scipy `1.5.2`).

MarcoGorelli:
closes #2515

### DIFF
--- a/doc/releases/v0.12.0.txt
+++ b/doc/releases/v0.12.0.txt
@@ -27,3 +27,5 @@ A paper describing seaborn was published in the `Journal of Open Source Software
 - Removed several previously-deprecated utility functions (`iqr`, `percentiles`, `pmf_hist`, and `sort_df`).
 
 - FacetGrid subplot titles will no longer be reset when calling `FacetGrid.map()` or `FacetGrid.map_dataframe()` after `FacetGrid.set_titles()` (:pr:`2705`).
+
+- :func:`lineplot` no longer throws an error when `data` contains an index with duplicates (:pr:`2714`)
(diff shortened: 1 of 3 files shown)


mwaskom:
Hi, see https://github.com/mwaskom/seaborn/issues/2652 for some more thoughts here.

- It's not just an issue in `lineplot`
- I'm not sure this is the best solution

Gonna approve CI as I'm curious if it passes, though

MarcoGorelli:
> Hi, see #2652 for some more thoughts here.

Thanks, that's useful!

For what it's worth, I like the solution

> a) raise with an informative error

- in any case, I'll close this PR

PR #2757, opened 2022-03-12
Fix #1530, change default cache directory

### LINKED ISSUE #1530: Change load_dataset cache dir default
The current default for `seaborn.load_dataset`s `data_home` directory, which is used for caching is `$HOME/seaborn-data`. This leads to unwanted pollution of the user's home directory. The calls used in many seaborn examples use only the dataset's name, just from this, it is unclear that by default a cache directory will be created in such a prominent location.

I believe a good alternative location be `$XDG_CACHE_DIR`, which according to the freedesktop specification is supposed to be used for _user-specific non-essential (cached) data_ (see [specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)). If the environment variable is unset, a default of `$HOME/.cache` is recommended.

Freed-Wu:
Change default cache directory to respect the standard of every platform.

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -12,6 +12,7 @@ import matplotlib as mpl
 from matplotlib.colors import to_rgb
 import matplotlib.pyplot as plt
 from matplotlib.cbook import normalize_kwargs
+from platformdirs import user_cache_dir
 
 from .external.version import Version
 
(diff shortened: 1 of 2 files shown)


mwaskom:
Hi, thank you for the submission, but I'm not going to add a dependency for this.

Freed-Wu:
Hi. I support to reduce dependencies. However, standardize
the cache directory is not conflict with reducing a dependency. You can refer
<https://github.com/pytorch/pytorch/blob/master/torch/_appdirs.py#L314-L317>.
A similar discussion can be found on
<https://github.com/scikit-learn/scikit-learn/pull/22776>. Maybe it can give
you other thought.

At last, thanks for your reply.

mwaskom:
As stated in the issue you are claiming to fix, I am open to vendoring appdirs (like torch does). In fact, seaborn really only needs the one specific `user_data_dir` function, since it doesn't write logs or bother with cross-user data. Also, it's really only important to cache the example datasets for CI, since they are lightweight and not essential for normal use. So I think it would be fine to do something very simple that handles linux and mac os and maybe modern windows systems (we don't have any windows CI) but otherwise ignores complicated edge cases and gracefully avoids caching. That said,

> I support to reduce dependencies. However, standardize the cache directory is not conflict with reducing a dependency

I agree, but your PR _did_ add a dependency, which is why it got closed here (and by scikit-learn).

Freed-Wu:
> I am open to vendoring appdirs (like torch does)

Great. To realize a lightweight submodule to do this job listens like a good idea.

PR #2799, opened 2022-05-13
CI: Update used actions, cache pip deps, 3.10 tests

EwoutH:
A bit of maintenance on the CI:
- Allow to manually trigger a the CI from the GitHub UI ([workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch))
- Update the used actions to actions/checkout@v3, actions/setup-python@v3 and codecov/codecov-action@v2
- Cache pip dependencies ([docs](https://github.com/actions/setup-python#caching-packages-dependencies))
- Run more tests with Python 3.10
- Bit of cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,32 +1,31 @@
-name : CI
+name: CI
 
 on:
   push:
     branches: master
   pull_request:
     branches: master
+  workflow_dispatch:
 
 env:
   NB_KERNEL: python
   MPLBACKEND: Agg
 
 jobs:
-
   build-docs:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
 
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Setup Python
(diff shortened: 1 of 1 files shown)


mwaskom:
Hi, thanks for this! CI maintenance is always super appreciated. Although I am a bit uncertain about what the right policy would be for the Python version that most tests run on ... if you believe [pypistats](https://pypistats.org/packages/seaborn), 3.10 gets vanishingly little usage relative to other versions. That said, I don't know that Python versioning per se has ever been an issue (aside from 2/3) so I suppose it makes sense to use the newest one, especially as they're investing in speed these days.

EwoutH:
The latest Python versions always have the latest features and most depreciation warnings, so it's good to have all test run on the latest. The older versions you mainly test for backwards compatibility, so only testing the oldest and newest should be enough. Sometimes you encounter a niche bug, for that the versions in between could be nice to test. But the oldest supported and newest are by far the most important.

mwaskom:
Makes sense. Thanks!

PR #2800, opened 2022-05-13
Use f-strings for string formatting

EwoutH:
Reformats all the text from the old "%-formatted" and .format(...) format to the newer f-string format, as defined in [PEP 498](https://www.python.org/dev/peps/pep-0498/). This requires Python 3.6+.

[Flynt](https://github.com/ikamensh/flynt) 0.76 was used to reformat the strings. 45 f-strings were created in 13 files.

F-strings are in general more readable, concise and performant. See also: [PEP 498#Rationale](https://www.python.org/dev/peps/pep-0498/#rationale)

```
Running flynt v.0.76

Flynt run has finished. Stats:

Execution time:                            2.066s
Files checked:                             102
Files modified:                            13
Character count reduction:                 411 (0.03%)

Per expression type:
Old style (`%`) expressions attempted:     28/30 (93.3%)
`.format(...)` calls attempted:            23/33 (69.7%)
No concatenations attempted.
F-string expressions created:              45
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -68,7 +68,7 @@ master_doc = 'index'
 # General information about the project.
 project = u'seaborn'
 import time
-copyright = u'2012-{}'.format(time.strftime("%Y"))
+copyright = f"2012-{time.strftime('%Y')}"
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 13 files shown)


mwaskom:
LGTM.

mwaskom:
LGTM.

PR #2808, opened 2022-05-16
Pyupgrade 3.7: Update syntax with Python 3.7+ features

EwoutH:
This PR updates the Python code with Python 3.7+, removing old, depreciated and redundant code. The biggest advantage is increased readability of the code. [pyupgrade](https://github.com/asottile/pyupgrade) 2.32.1 was used with `--py37-plus`.

This includes:
- f-string formatting of strings ([PEP 498](https://www.python.org/dev/peps/pep-0498/), Python 3.6+)
- utf-8 encoding is now the default ([PEP 3120](https://www.python.org/dev/peps/pep-3120/), Python 3.0+)
- Replace list comprehensions by Generator Expressions ([PEP 289](https://www.python.org/dev/peps/pep-0289/), Python 2.4+)
- Replace `yield` loop by `yield from` ([PEP 380](https://www.python.org/dev/peps/pep-0380/), Python 3.3+)
- Replace the IOError alias by OSError ([PEP 3151](https://www.python.org/dev/peps/pep-3151/), Python 3.3+)
- Remove the default subclass `(object)` when defining a class
- Define sets with curly braces `{}` instead of `set()`
- Remove `"r"` parameter from open function, which is default

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # seaborn documentation build configuration file, created by
 # sphinx-quickstart on Mon Jul 29 23:25:46 2013.
@@ -66,7 +65,7 @@ source_suffix = '.rst'
 master_doc = 'index'
 
 # General information about the project.
-project = u'seaborn'
+project = 'seaborn'
 import time
 copyright = f"2012-{time.strftime('%Y')}"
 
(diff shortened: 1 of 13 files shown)


mwaskom:
LGTM, thanks!

mwaskom:
Oops, weird, I wonder why this didn't fail the lint check until it landed on master.

PR #2819, opened 2022-05-25
Logarithmic hist plot with multi='dodge': unequal bar width

### LINKED ISSUE #2820: Logarithmic hist plot with multi='dodge': unequal bar width
Using a hist plot with logarithmic x-axis results in unequal bar width.

Fix PR is provided in #2819

infosec-it-init:
Fixes #2820


mwaskom:
Hi, thanks for this. The code looks right to me but could you please add a simple test to cover it?  The test should fail with the previous code but pass with this fix.

PR #2877, opened 2022-06-25
Add Scale.label interface for formatting ticks

### LINKED ISSUE #2840: Seaborn objects log/symlog scales - default ticks
I experimented a little bit with the objects interface. My use case involves plotting large numbers using log/symlog scaled y-axis. I noticed that the default selection of tick locators/formatters in the objects interface differs from the corresponding (log/symlog) scale's matplotlib default. This is most prominent when dealing with large numbers. Consider the following code:
```python
import seaborn as sns; import seaborn.objects as so

outfile_prefix="so"
diamonds["price"] = diamonds["price"] * 10**3

for scale in ["log", "symlog"]:
    fig, (ax1, ax2) = plt.subplots(1, 2)
    fig.suptitle(f"Scale={scale}")
    (
        so.Plot(
            diamonds, "carat", "price",
        )
        .on(ax2)
        .add(so.Scatter())    
        .scale(y=scale)
        .plot()
    )
    ax1.set_yscale(scale)
    sns.scatterplot(data=diamonds, x="carat", y="price", ax=ax1)
    ax1.set_title("Seaborn classic")
    ax2.set_title("Seaborn objects")
    savepath=f"{outfile_prefix}_diamonds_{scale}.png"
    plt.savefig(savepath, dpi=300)
```
The log-scaled plot uses a similar locator, but a different formatter:
so_diamonds_log

While the symlog-scaled plot differ uses a different locator and a different formatter:
so_diamonds_symlog

Adding the following lines right after the plotting command restores the matplotlib default formatting:
```python
    from matplotlib.ticker import LogFormatterSciNotation, LogLocator, SymmetricalLogLocator
    ax2.yaxis.set_major_locator(SymmetricalLogLocator(base=10, linthresh=10) if scale=="symlog" else LogLocator(base=10))
    ax2.yaxis.set_major_formatter(LogFormatterSciNotation(base=10))
```
I could not find a straightforward way to change the formatting using the objects API (while I appreciate that `Scale` objects accept ticks input). To me, the default matplotlib selection of locators/formatters is the desired behavior. I suggest this should be the default behavior for the objects interface (when using log/symlog scales).

mwaskom:
e.g.

```python
(
    so.Plot(tips, "total_bill")
    .add(so.Bar(), so.Hist())
    .scale(x=so.Continuous().label(like="${x:.0f}"))
)
```
image

```python
(
    so.Plot(planets, "distance")
    .add(so.Bar(), so.Hist())
    .scale(x=so.Continuous().label(unit="pc"))
)
```
image

This also makes log transformed scales default to a log formatter (closes #2840)

```python
(
    so.Plot(planets, "distance")
    .add(so.Bar(), so.Hist())
    .scale(x="log")
)
```
image

This also shortens the `transform` `Scale` parameter to `trans`, as testing revealed that typing out `transform` is a little cumbersome.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,0 @@
-[pytest]
-filterwarnings =
-;   Warnings raised from within patsy imports
-    ignore:Using or importing the ABCs:DeprecationWarning
-junit_family=xunit1
--- a/seaborn/_core/plot.py
+++ b/seaborn/_core/plot.py
@@ -23,7 +23,7 @@ from seaborn._marks.base import Mark
 from seaborn._stats.base import Stat
 from seaborn._core.data import PlotData
 from seaborn._core.moves import Move
-from seaborn._core.scales import Scale, ScaleTransform
+from seaborn._core.scales import Scale
 from seaborn._core.subplots import Subplots
 from seaborn._core.groupby import GroupBy
 from seaborn._core.properties import PROPERTIES, Property, Coordinate
@@ -677,7 +677,7 @@ class Plotter:
         self._legend_contents: list[
             tuple[str, str | int], list[Artist], list[str],
         ] = []
-        self._scales: dict[str, ScaleTransform] = {}
+        self._scales: dict[str, Scale] = {}
 
     def save(self, loc, **kwargs) -> Plotter:  # TODO type args
         kwargs.setdefault("dpi", 96)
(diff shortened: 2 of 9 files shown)


mwaskom:
Docs still need some tweaking but merging to do that on the pydata-theme branch.

PR #2900, opened 2022-07-12
Update docstring in subplots.py 

Foxed1337:
In commit 6b61a26a462effaea1c80518e98185abb12174ed in `subplots.py` have been removed init argument `data: PlotData`, but it's still in docstring


seaborn/_core/subplots.py
```diff
@@ -21,12 +21,10 @@ class Subplots:
     ----------
     subplot_spec : dict
         Keyword args for :meth:`matplotlib.figure.Figure.subplots`.
-    facet_spec : dict
+    facet_spec : FacetSpec
```
mwaskom, line 24: I don't think this is the right way to document the inputs here; `FacetSpec` is something `mypy` would understand but it's not something you would instantiate and pass. (This class is internal anyway and so probably the signature annotations suffice, and these types could be removed).

Foxed1337: Thanks for the explanation! I got it

PR #2904, opened 2022-07-13
Added new test for heatmap

codemurt:
Added new test in `test_matrix.py` for different shapes in heatmap's annotation. And fixed typo in test name.

### DIFF
--- a/tests/test_matrix.py
+++ b/tests/test_matrix.py
@@ -348,7 +348,7 @@ class TestHeatmap:
 
     def test_heatmap_annotation_different_shapes(self):
 
-        annot_data = self.df_norm.append(self.letters)
+        annot_data = self.df_norm.iloc[:-1]
         with pytest.raises(ValueError):
             mat.heatmap(self.df_norm, annot=annot_data)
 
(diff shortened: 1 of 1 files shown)


mwaskom:
What is the motivation here?

tests/test_matrix.py
```diff
@@ -346,6 +346,12 @@ def test_heatmap_annotation_other_data(self):
             assert text.get_text() == f"{val:.1f}"
             assert text.get_fontsize() == 14
 
+    def test_heatmap_annotation_different_shapes(self):
+
+        annot_data = self.df_norm.append(self.letters)
```
mwaskom, line 351: `DataFrame.append` is a deprecaed method: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.append.html

Please use a different way to construct the annotation matrix (I suggest `self.df_norm.iloc[:-1]`)

mwaskom (changes requested):

mwaskom (approved):

mwaskom:
Thanks @codemurt, improving tests is always welcome. It can help when you submit a PR to provide a short motivation for the change you are proposing. Something as simple as "I looked through the coverage report and found a path that wasn't tested" is fine, and would save a maintainer the time of trying to understand what the PR is about.

PR #2905, opened 2022-07-13
Added new test for establish_colors

Matvey-SD:
Added new tests in `test_categorical.py` for different default palette values and dict as palette without hue.


mwaskom:
What’s the motivation here?

mwaskom:
Thanks for trying to improve test coverage, always a good aim, but these aren't covering realistic usage patterns so I don't think the addition would be an improvement.

Issue #2911, opened 2022-07-17
Support multiple marks / stats in a single layer


mwaskom:
Concrete use-case is something like combining `Dot` and `Interval` to show a point estimate and confidence intervals. Currently that would require something like

```python
Plot(...).add(Interval(), Est()).add(Dot(), Agg())
```

But that gets annoying as soon as you start adding moves, layer-specific variables, etc. It would be preferable to make it easy to combine marks. One can also imagine wanting to have "stat pipelines", e.g normalize and then aggregate.

We already support multiple moves by passing `list[Move]`. So that's the most straightforward solution. But unlike with the move, some aspects of the plot setup depend on properties of the mark (e.g., its default orientation rules) and it would be messier to always be looping over the marks. You might also want  a more sophisticated combination of the objects, e.g. to only need to specify properties directly once and have them "inherited".

I think there's probably three options here:

1. Accept `list[Mark]`
2. Allow `Mark + Mark` to generate a new mark, potentially with some inheritance of properties (though trickier to define)
3. Don't accept multiple marks, but internally have a `CompoundMark` concept that makes it easier to define, e.g. `DotInterval`

I think I would kind of lean towards 2 but maybe just because it feels snazzy (which is to say, not for good reason). It would look something like

```python
Plot(...).add(Dot() + Interval(), Est()
```

IDK, that's kind of nice I guess.

Issue #2894, opened 2022-07-11
Rethink approach to grouping in plot phase of Plot


mwaskom:
Currently we can declare that a `Mark` property should form groups at the property level. The reason we need to do this is that different marks behave differently: e.g. each line has the same values for all its properties and is added separately, while a scatterplot can mix multiple property values in a single artist. But I don't think we have or will encounter cases where only a subset of a marks properties should group, so it is cumbersome to have to set `grouping=` for every property.

Instead, I think this can be determined within `Mark._plot` by adding a parameter to the `split_gen` generator, where the mark passes in the properties that should be grouping.

This also touches on a broader issue which is that the current grouping is relatively inefficient (e.g. see #2881). Ideally, we would do scaling over all data points and then group, which can be faster. The challenge has been that we no longer have a dataframe after scaling. The main reason is that working with colors as rgba tuples / n x 4 arrays is difficult in the context of a dataframe ... you can stick the tuples in a series, but then it has an object dtype that propagates through to the numpy array and works poorly with matplotlib. A few options would be:

- Implement our own groupby logic on the dict of arrays / lists that we have after scaling
- Store rgba values as separate columns in the dataframe we build while scaling (we could perhaps use a differnet internal color representation to facilitate things like `luminance` properties)
- Implement some kind of RGBA extension array that lets a Series hold a 2d data structure (is this possible? I am not sure)

PR #2912, opened 2022-07-18
Add Est stat and Interval mark to show error bars

mwaskom:
This PR adds a new mark and a new stat that together will add error bars to aggregations:

```python
(
    so.Plot(diamonds, x="carat", y="clarity")
    .add(so.Interval(), so.Est("mean", "sd"))
    .add(so.Dot(), so.Agg())
)
```
image

The `Est` stay has the same basic API for error bar selection as we'll be rolling out for the functional interface in 0.12 (i.e., `ci`/`se`/`pi`/`sd`, with an optional scale/level parameter.

Note that is currently necessary to add two layers to get a dot + interval; see #2911 for some ideas there

This PR also addresses some wrinkles relating to matplotlib capstyles in both the interval mark and existing line marks.

### DIFF
--- a/seaborn/_core/plot.py
+++ b/seaborn/_core/plot.py
@@ -1354,10 +1354,11 @@ class Plotter:
                         # Matplotlib (usually?) masks nan data, so this should "work".
                         # Downstream code can also drop these rows, at some speed cost.
                         present = axes_df.notna().all(axis=1)
-                        axes_df = axes_df.assign(
-                            x=axes_df["x"].where(present),
-                            y=axes_df["y"].where(present),
-                        )
+                        nulled = {}
+                        for axis in "xy":
+                            if axis in axes_df:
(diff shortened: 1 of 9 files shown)


mwaskom:
Here's something that needs to be addressed:

```python
(
    so.Plot(penguins, x="bill_length_mm", y="species", color="sex")
    .add(so.Interval(), so.Est(), so.Dodge())
    .add(so.Dot(), so.Agg(), so.Dodge())
)
```
image

This is because `Interval` and `Dot` end up with different `grouping_props`; it can be worked around:

```python
(
    so.Plot(penguins, x="bill_length_mm", y="species", color="sex")
    .add(so.Interval(), so.Est(), so.Dodge(by=["color"]))
    .add(so.Dot(), so.Agg(), so.Dodge())
)
```
image

But this needs a cleaner solution (cf #2894)

PR #2933, opened 2022-08-01
Document the legend argument of categorical scatterplots

MaozGelbart:
The `legend` argument was added in #2828 but is not documented (it is mentioned in the release notes). This PR documents this argument, and uses it in the docstrings and one `stripplot` example.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2134,8 +2134,12 @@ _categorical_docs = dict(
         and tick labels.\
     """),
     legend=dedent("""\
-    legend : str or bool, optional
-        If `False` or an empty string and `hue` is used, don't draw a legend.
+legend : "auto", "brief", "full", or False
+    How to draw the legend. If "brief", numeric `hue` and `size`
+    variables will be represented with a sample of evenly spaced values.
+    If "full", every group will get an entry in the legend. If "auto",
+    choose between brief or full representation based on number of levels.
+    If `False`, no legend data is added and no legend is drawn.
     """),
     ax_in=dedent("""\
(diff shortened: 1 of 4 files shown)


seaborn/categorical.py
```diff
@@ -2133,6 +2133,10 @@ def plot(self, ax, boxplot_kws):
         Function for converting categorical data into strings. Affects both grouping
         and tick labels.\
     """),
+    legend=dedent("""\
+    legend : str or bool, optional
+        If `False` or an empty string and `hue` is used, don't draw a legend.
```
mwaskom, line 2138:
```suggestion
legend : "auto", "brief", "full", or False
    How to draw the legend. If "brief", numeric `hue` and `size`
    variables will be represented with a sample of evenly spaced values.
    If "full", every group will get an entry in the legend. If "auto",
    choose between brief or full representation based on number of levels.
    If `False`, no legend data is added and no legend is drawn.
```

seaborn/categorical.py
```diff
@@ -3272,7 +3278,7 @@ def catplot(
     {color}
     {palette}
     {hue_norm}
-    legend : bool, optional
+    legend : str or bool, optional
         If ``True`` and there is a ``hue`` variable, draw a legend on the plot.
```
mwaskom, line 3282:
```suggestion
    legend : str or bool, optional
        Set to `False` to disable the legend. With `strip` or `swarm` plots,
        this also accepts a string, as described in the axes-level docstrings.
```

mwaskom (approved):
Thanks for catching this!

mwaskom:
Nice catch, thanks!

PR #2940, opened 2022-08-06
Replace setup.py with pyproject.toml configured to use flit for builds

mwaskom:
Tests performed after running `flit build`:

- Install from source tarball
- Untar source tarball and take a look-sie
- Install from wheel

Haven't yet tested any uploads but will send the RC to test pypi first.


Carreau:
I'm not an expert, but that looks good to me. 

Suggestion to be more at ease:
- make 2 wheels, before and after this commit and compare them ?

mwaskom:
Thanks @Carreau, I have built a wheel (`flit build`) and seaborn using it `pip install <wheel_path>` and the files in site-packages look correct, but is there any other way to "diff" two wheels?

Carreau:
wheel are just zip, so your preferred way to diff folders. i would probably just write a custom python for loop.

mwaskom:
Welp, what's the worst that could happen.

PR #2953, opened 2022-08-12
Fix: shifting margins caused by multiple rugplots

beyarkay:
This fix prevents margins from increasing when multiple rugplots are
added to the same `ax`, even if `expand_margins` is `False`.

Screenshot 2022-08-12 at 22 11 42

As a minimum reproducible example (which is similar to the added unit test):

```py
import numpy as np

values = np.linspace(start=0, stop=1, num=5)
ax = sns.lineplot(x=values, y=values)
sns.rugplot(x=values, ax=ax)

ylim = ax.get_ylim()

for _ in range(4):
    sns.rugplot(x=values, ax=ax, expand_margins=False)

if not all(a == b for a, b in zip(ylim, ax.get_ylim())):
    print(f'{ylim} != {ax.get_ylim()}')
plt.suptitle("Example showing that multiple rugplots cause issues")
plt.show()
```

Running the above code:

```sh
$ pip install seaborn numpy matplotlib; python3 test.py
(-0.1, 1.1) != (-0.61051, 1.14641)
```

This bug was caused by how seaborn detects the correct colors to use. In
`seaborn/utils.py`, in method `_default_color`, the following line used
to resolve to `ax.plot([], [], **kws)`:

```py
scout, = method([], [], **kws)
```
But matplotlib has the parameters `scalex` and `scaley` of `ax.plot` set
to `True` by default. Matplotlib would see that the rug was already on
the `ax` from the previous call to `sns.rugplot`, and so it would
rescale the x and y axes. This caused the content of the plot to take up
less and less space, with larger and larger margins as more rugplots
were added.

The fix sets `scalex` and `scaley` both to `False`, since this plot
method is a null-plot and is only used to check what colours should be
used:
```py
scout, = method([], [], scalex=False, scaley=False, **kws)
```

The above line is within an if-elif-else branch, but the other branches
do not suffer this bug and so no change is needed for them.

An additional unit test was also added to catch this bug:
`test_multiple_rugs`.

### DIFF
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -305,17 +305,15 @@ class TestRugPlot(SharedAxesLevelTests):
         assert x1 == x2
         assert y1 + height * 2 == pytest.approx(y2)
 
-    def test_multiple_rugs(self, long_df):
+    def test_multiple_rugs(self):
 
         values = np.linspace(start=0, stop=1, num=5)
-        ax = lineplot(x=values, y=values)
-        rugplot(x=values, ax=ax)
+        ax = rugplot(x=values)
         ylim = ax.get_ylim()
 
-        for j in range(4):
-            rugplot(x=values, ax=ax, expand_margins=False)
+        rugplot(x=values, ax=ax, expand_margins=False)
 
-        assert_array_equal(ylim, ax.get_ylim())
+        assert ylim == ax.get_ylim()
 
     def test_matplotlib_kwargs(self, flat_series):
 
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -31,6 +31,9 @@ from seaborn.distributions import (
     kdeplot,
     rugplot,
 )
+from seaborn.relational import (
+    lineplot,
+)
 from seaborn.external.version import Version
 from seaborn.axisgrid import FacetGrid
 from seaborn._testing import (
@@ -301,6 +304,18 @@ class TestRugPlot(SharedAxesLevelTests):
         x2, y2 = ax.margins()
         assert x1 == x2
         assert y1 + height * 2 == pytest.approx(y2)
+
+    def test_multiple_rugs(self, long_df):
+
+        values = np.linspace(start=0, stop=1, num=5)
+        ax = lineplot(x=values, y=values)
+        rugplot(x=values, ax=ax)
+        ylim = ax.get_ylim()
+
+        for j in range(4):
+            rugplot(x=values, ax=ax, expand_margins=False)
+
+        assert_array_equal(ylim, ax.get_ylim())
 
     def test_matplotlib_kwargs(self, flat_series):
 
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -31,9 +31,6 @@ from seaborn.distributions import (
     kdeplot,
     rugplot,
 )
-from seaborn.relational import (
-    lineplot,
-)
 from seaborn.external.version import Version
 from seaborn.axisgrid import FacetGrid
 from seaborn._testing import (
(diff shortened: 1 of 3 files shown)


beyarkay:
I saw that the unit test doesn't quite fit the format of the others, so let me know if I should change that. I'm happy to do so if it'll leave you with a more consistent code base (:

tests/test_distributions.py
```diff
@@ -302,6 +305,18 @@ def test_expand_margins(self, flat_array):
         assert x1 == x2
         assert y1 + height * 2 == pytest.approx(y2)
 
+    def test_multiple_rugs(self, long_df):
+
+        values = np.linspace(start=0, stop=1, num=5)
+        ax = lineplot(x=values, y=values)
```
mwaskom, line 311: Having the `lineplot` helps demonstrate the issue visually but it doesn't seem necessary to trigger the error programmatically, so I don't think we need it in the test.

tests/test_distributions.py
```diff
@@ -302,6 +305,18 @@ def test_expand_margins(self, flat_array):
...
+        rugplot(x=values, ax=ax)
+        ylim = ax.get_ylim()
+
+        for j in range(4):
```
mwaskom, line 315: Similar to above, calling rugplot once should trigger expansion even if it's a little hard to visually notice, so the loop is probably not necessary.

tests/test_distributions.py
```diff
@@ -302,6 +305,18 @@ def test_expand_margins(self, flat_array):
...
+            rugplot(x=values, ax=ax, expand_margins=False)
+
+        assert_array_equal(ylim, ax.get_ylim())
```
mwaskom, line 318:
```suggestion
        assert ylim == ax.get_ylim()
```
Nit: these will both be tuples, so simple equality comparison should be sufficient.

tests/test_distributions.py
```diff
@@ -302,6 +305,18 @@ def test_expand_margins(self, flat_array):
         assert x1 == x2
         assert y1 + height * 2 == pytest.approx(y2)
 
+    def test_multiple_rugs(self, long_df):
```
mwaskom, line 308: Nit: don't need to call the `long_df` fixture since you don't use it here.

mwaskom (changes requested):
Great sleuthing, and thanks for the fix!

Basic structure of the test looks totally fine, just left a couple of comments about keeping it as simple as possible.

Pleas also add an entry to the v0.12 release notes, e.g. "Fixed a bug where :func:`rugplot` was ignoring `expand_margins=False". This should be categorized as a |Fix|, and please link back to the PR so you get credit!

beyarkay:
And thank you for _the entire friggin' library_!😁

It's getting late at my local time so I'll respond properly in the morning, but I don't see any remarks that I disagree with.

tests/test_distributions.py
beyarkay: TIL. I've replaced it with your suggestion 👍

tests/test_distributions.py
beyarkay: Made the change 👍

tests/test_distributions.py
beyarkay: Ah I actually had assumed that `lineplot` was required, but I see you're correct. I've made the change 👍

beyarkay:
Hang on, I forgot to remove the import.

beyarkay:
There we go, let me know if there are any other changes 😄 

As a side note, is there a release date planned for v0.12? I came across this bug while working on a university project so it would be nice to see the update roll out.

mwaskom (approved):

mwaskom:
Thanks again!

>  is there a release date planned for v0.12?

No specific date planned, but I would expect v0.12.0 by the end of the month.

PR #2958, opened 2022-08-15
Fix: typos in nextgen docs

beyarkay:
I wasn't sure about the API change at first, but having gone through the demonstration I'm so excited now. Anyway, here's some typos I found.


mwaskom:
Sorry @beyarkay for not following up on this originally. These pages ended up getting replaced in the docs for the final release so I think this is moot. But don't worry, I'm sure there are plenty of typos in the new content :)

PR #3012, opened 2022-09-11
Use stable version of PyData Sphinx Theme for doc building

EwoutH:
0.10.0rc2 was a pre-release version of the PyData Sphinx Theme, this PR bumps the required version to 0.10 and thereby use a stable version.

### DIFF
--- a/ci/cache_test_datasets.py
+++ b/ci/cache_test_datasets.py
@@ -12,6 +12,7 @@ datasets = (
     "flights",
     "fmri",
     "iris",
+    "penguins",
     "planets",
     "tips",
     "titanic"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ docs = [
     "sphinx-issues",
     "sphinx-design",
     "pyyaml",
-    "pydata_sphinx_theme==0.10.0rc2",
+    "pydata_sphinx_theme>=0.10",
 ]
 
 [project.urls]


mwaskom:
This is actually intentional — they changed something between the rc and final release that caused the docs to render incorrectly. I do need to figure out what it was, but for the time being the docs need to build with that specific tag.

mwaskom:
(In any case, the doc dependencies should pin a specific version of this theme, because it's still highly unstable).

EwoutH:
It's a very strange error, since all CI jobs fail and it looks like on a URL request to the [seaborn-data](https://github.com/mwaskom/seaborn-data) repository after the penguins dataset can't be found.

That looks like seaborn internals to me, how does that change by bumping a doc building theme?

mwaskom:
Sorry, two separate issues: the doc build does need to be pinned on the rc to look correct (it will still succeed). It looks like you hit an unrelated github flake, though.

EwoutH:
Right, I just validated that it was a weird CI flake on all jobs. I think it's still good to add `"penguins"` to the cached datasets right?

mwaskom:
> I think it's still good to add "penguins" to the cached datasets right?

It looks like this would have no effect as currently configured — the datasets are only cached for the unit/doc tests, but penguins isn't used in any of those. It would be relevant for the doc build, but that doesn't seem to be actually running notebooks in parallel.

In either case I didn't actually realize that specific datasets were enumerated in the caching script; it would be better (though harder) to parse the source and figure out what names are needed on the fly.

EwoutH:
Thanks for the extensive explanation, for now I will close this PR.

mwaskom:
Thanks for catching that penguins wasn't getting cached, prompted me to look into the config further and got the docs to build in parallel now: https://github.com/mwaskom/seaborn/pull/3014

PR #3013, opened 2022-09-11
CI: Add biweekly scheduled run to detect changes in environment or dependencies

EwoutH:
This PR adds schedule that triggers the CI twice a week (on Monday and Thursday 06:00 UTC). This way changes in the GitHub Actions environment or upstream dependencies that results into failures, errors or warnings can be detected and traced back seperately from code changes.

So If the CI fails, with this schedule, there is always an up-to-date reference run from the master branch to compare it to.


mwaskom:
Thanks @EwoutH, this is handy.

PR #3019, opened 2022-09-12
Lines 644 and 761 in relational.py needed editing

biswaroopmitra:
Hello. Lines 644 and 761 needed editing.


mwaskom:
Thanks but I don’t think your proposed changes improve the text.

PR #3031, opened 2022-09-16
errorbar doc patched

spisakt:
This change is proposed to make it explicit in the documentation that 'None' can be used to switch off error bar.


mwaskom:
Thanks, good catch!

Issue #3104, opened 2022-10-21
stripplot edgecolor ignored


markmbaum:
The `stripplot` function is ignoring my `edgecolor` choice. I'm using the Jupyter extension in VS Code, which defaults to the dark matplotlib style. Maybe that has something to do with it? An example is below, using seaborn 0.12.1

```
import seaborn as sns

tips = sns.load_dataset('tips')
sns.stripplot(
    data=tips,
    x='day',
    y='total_bill',
    edgecolor='black'
)
```
![example](https://user-images.githubusercontent.com/18745581/197267866-1d413be6-3ef8-484d-80a8-25b5f723c1cd.png)

mwaskom:
Yeah this is a gotcha; the issue is that stripplot has `linewidth=0` by default so you need to change that too if you want visible edges.

markmbaum:
Can I submit a PR putting a note about this into the `stripplot` docstring?

mwaskom:
sure

mwaskom:
Closed by #3110 (fyi the git auto closing language is pretty narrow)

PR #3110, opened 2022-10-23
include note in stripplot docstring about edgecolor depending on linewidth

markmbaum:
this is a minor update to the `stripplot` docstring closing #3104


mwaskom:
Thanks @markmbaum

Issue #2985, opened 2022-08-30
Consider adding optional caps to objects.Range


ItamarOlmert:
The classic stylistic choice, and still the one usually expected for papers, is that error bars have caps on their bottom/top.

Currently, this is possible in the old seaborn interface using the `capsize` keyword, setting it to above 0:

```py
import matplotlib.pyplot as plt
import seaborn as sns
import seaborn.objects as so

tips = sns.load_dataset("tips")

sns.barplot(tips, x='day', y='total_bill', hue='sex', capsize=0.05)
plt.show()
```
![output](https://user-images.githubusercontent.com/29255731/187063609-82d33834-0a73-4114-ae35-1d8e45a7bcc0.png)

However, as of version `0.12.0rc0` when doing something similar with the nextgen interface:
```py
(
    so.Plot(tips, "day", "total_bill", color="sex")
    .add(so.Bar(), so.Agg(), so.Dodge())
    .add(so.Range(), so.Est(), so.Dodge(), legend=False)
    .show()
)
```
![image](https://user-images.githubusercontent.com/29255731/187063843-d055c310-cb43-41f5-bcdb-3334020cfaca.png)
There isn't any (built-in) way to add caps to the `seaborn.objects.Range` object, at least as far as I can see.

Thank you!

PR #3134, opened 2022-11-11
Add caps to the Range object

thuiop:
As suggested in issue #2985, this PR adds an optional capsize parameter to the Range object for drawing caps at the end, e.g. for error bars.
Here is an example of how it looks : 
```
import seaborn as sns
import seaborn.objects as so

penguins = sns.load_dataset("penguins")
p = (
    so.Plot(penguins, x="body_mass_g", y="species", color="sex")
    .add(so.Range(capsize=0.1), so.Est(errorbar="sd"), so.Dodge())
    .add(so.Dot(), so.Agg(), so.Dodge())
)
```
seaborn_range_caps

### DIFF
--- a/doc/_docstrings/objects.Range.ipynb
+++ b/doc/_docstrings/objects.Range.ipynb
@@ -3,7 +3,6 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "2923956c-f141-4ecb-ab08-e819099f0fa9",
    "metadata": {
     "tags": [
      "hide"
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks for taking a crack at this, but the change is not sufficient — any additional parameters need to be defined as mappable properties, which would mean there would need to be a new `capsize` property. I haven't decided whether that's something I want to add yet. In the mean time, you can put caps on error bars with the `Dash` mark...

thuiop:
Oh, I see. I had thought about the `Dash` mark but this is not very convenient, especially for something as common as having caps at the end of error bars. Feel free to do whatever you want with the PR.

mwaskom:
It'll be more convenient when https://github.com/mwaskom/seaborn/pull/3120 lands.

PR #3135, opened 2022-11-11
Fix bug with boolean values for color

thuiop:
Fixes the bug from #3106 by using a Nominal scale for boolean instead of Continuous.


mwaskom:
I think this should be fixed in `Continuous` but making sure theres a cast to float before the subtraction. Then I also think there could be a `Boolean` or `Binary` scale type. I'm not sure about defaulting to `Nominal`.

thuiop:
I do agree that there could be a `Boolean` scale ; however, I do feel that boolean data is more at home with `Nominal` rather than `Continuous` as there is a discrete number of values.

mwaskom:
In any case, this is not the right place to add the default scale because color properties are not the only ones that run into the problem.

Issue #3118, opened 2022-10-29
`stripplot` and `swarmplot` fail on empty data when `palette` is set


MaozGelbart:
You can do the following and get an empty plot:
```python
sns.stripplot(x=[], y=[], hue=[], palette=None)
```
But providing a palette raises an error message that is not so informative:
```python
sns.stripplot(x=[], y=[], hue=[], palette=[])
```
```
<stdin>:1: FutureWarning: Passing `palette` without assigning `hue` is deprecated.
<stdin>:1: UserWarning: Ignoring `palette` because no `hue` variable has been assigned.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../seaborn/seaborn/categorical.py", line 2537, in stripplot
    p.plot_strips(
  File ".../seaborn/seaborn/categorical.py", line 265, in plot_strips
    offsets = self._nested_offsets(width, dodge)
  File ".../seaborn/seaborn/categorical.py", line 240, in _nested_offsets
    n_levels = len(self._hue_map.levels)
TypeError: object of type 'NoneType' has no len()
```
The same call using `scatterplot` results with an empty plot (as expected).
The general use case is when a Dataframe is provided as input data along with x, y and hue variables, and also a palette but the Dataframe is (unfortunately) empty. Something like:
```python
tips = sns.load_dataset("tips")
tips = tips.drop(tips.index)
sns.stripplot(data=tips, x="sex", y="tip", hue="smoker", palette=[])

mwaskom:
I think most functions short circuit on empty x/y to avoid downstream special casing.

mwaskom:
Closed by #3136 (although it would still be good to have these functions short circuit since otherwise you end up playing edge case whack-a-mole with empty data).

PR #3136, opened 2022-11-11
Fix issue when an empty palette is provided for an empty dataset

thuiop:
Fixes issue #3118, by checking whether `self._hue_palette.levels` is `None` before calling `len` on it.


mwaskom:
As mentioned in the issue, I think that most functions short-circuit when there's no x/y data to avoid the need for this downstream special-casing.

For this to be accepted, there would need to be a test replicating the original problem.

thuiop:
Actually it seems that only categorical plots would be concerned by the issue, as in the other cases `len` is never called on `self._hue_map.levels`. Anyway, I added a small test and checked that it failed before this PR and is now passing.

PR #3138, opened 2022-11-11
Fixed the issue with wrong width in log scale for Bar(s)

thuiop:
Fixes the issue #2907, but in a hacky way. The core issue is that currently, the bars' positions relative to each other are defined by a coordinate (say x) and their width. When applying a transform, what happens is that x is transformed (into `transformed_x`), and the new width is obtained with `transformed_width = transform(x+width/2)-transform(x-width/2)`. While this is actually correct, it falls apart when passing the coordinates to the matplotlib API, which asks for the bottom left of the rectangle. It is currently obtained by doing `transformed_x-transformed_width/2`, whereas it should really be `transform(x-width/2)`, which is obviously not the same in the case of the log.
In this PR, I solve the problem by storing the `transform(x-width/2)` value and retrieving it when needed, but this feels a bit weird to have it as an additional variable ; however we cannot ditch the normal `transformed_x` as it is needed for other things than the bars. There is probably some underlying flow change that would make this more straightforward but I expect this to be a big change.

### DIFF
--- a/seaborn/_core/plot.py
+++ b/seaborn/_core/plot.py
@@ -1356,6 +1356,7 @@ class Plotter:
 
             grouping_vars = mark._grouping_props + default_grouping_vars
             split_generator = self._setup_split_generator(grouping_vars, df, subplots)
+
             mark._plot(split_generator, scales, orient)
 
         # TODO is this the right place for this?
@@ -1397,6 +1398,7 @@ class Plotter:
                     out_df.loc[values.index, f"{orient}base"] = (
                         transform(values - width / 2)
                     )
+
         return out_df
 
     def _generate_pairings(
--- a/seaborn/_core/plot.py
+++ b/seaborn/_core/plot.py
@@ -1356,7 +1356,6 @@ class Plotter:
 
             grouping_vars = mark._grouping_props + default_grouping_vars
             split_generator = self._setup_split_generator(grouping_vars, df, subplots)
-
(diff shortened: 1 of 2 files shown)


mwaskom:
Is this necessary, since `Mark._plot` has access to the scales?

thuiop:
Hm. Then it may be possible to do it in a better way, but I fear it will still feel weird ; in the` _unscale_coords` function, where the magic happens, it is the `axis.get_transform().inverted()` which is used (which ultimately comes from the scales, I agree). Perhaps the best way to do it would be to get rid of the `_unscale_coords` function in Plot altogether and do the relevant conversions in the `Mark` directly, but this is quite a larger step.

mwaskom:
> Perhaps the best way to do it would be to get rid of the _unscale_coords function in Plot altogether and do the relevant conversions in the Mark directly

That seems like it would be unnecessarily repetitive in most cases?

thuiop:
> Is this necessary, since `Mark._plot` has access to the scales?

Actually, while looking at it, I would say it is, because in `Mark._plot` `data` already contains the transformed `x` and width and not the originals. It is surely possible to retrieve the original x and width with the scales but this would probably be uglier than the solution I proposed.

thuiop:
(superseded by #3217)

mwaskom:
Thanks for tracking down the root cause!

PR #3141, opened 2022-11-14
Fix a copy & paste error in the docstring for FacetGrid.set_titles

hagau:
As the title says, it fixes a copy & paste error in the docstring for `FacetGrid.set_titles` :smiley:.


mwaskom:
Thanks, nice catch.

PR #3177, opened 2022-12-06
COMPAT: Change pandas use_inf_as_null to reflect 2.0 deprecation

mroeschke:
pandas 2.0 will enforce the deprecation of the option `use_inf_as_null` in favor of `use_if_as na`, so changing the usages I found in seaborn


mwaskom:
Thanks!

PR #3223, opened 2023-01-15
Add default theme configuration to Plot

mwaskom:
This PR introduces a mechanism for changing the appearance of _all_ plots made with `so.Plot`. It adds a configuration interface, accessible through `so.Plot.config` (this is an attribute of the class, not an instance). I expect to add a number of configuration objects; this PR adds `so.Plot.config.theme`, which exposes a dictionary of rcParams that will be used by default in all `Plot` plots, unless overridden by `so.Plot().theme()`. So to change the appearance of all plots, you could put at the top of your notebook:

```python
theme = sns.axes_style("whitegrid") | sns.plotting_context("talk")
so.Plot.config.theme.update(theme)
```
then do
```python
so.Plot()
```

To use the matplotlib globals (whether determined by a `matplotlibrc` file or functions that modify the global state, you'd only need do)

```python
so.Plot.config.theme.update(mpl.rcParams)  # or plt.rcParams
```

I've added a nice `_repr_html_` for the theme config object, so you can see the available parameters and what they're set to more easily:

image

(One could even imaging making this an editable table for interactive updating. We'll see...)

Also, we actually store the theme defaults in a matplotlib `RcParams` object, so we get to take advantage of their validation:

```python
so.Plot.config.theme["axes.linewidth"] = "thick"
```
```python-traceback
Cell In [9], line 1
----> 1 so.Plot.config.theme["axes.linewidth"] = "thick"

File ~/miniconda/envs/py310/lib/python3.10/site-packages/matplotlib/__init__.py:651, in RcParams.__setitem__(self, key, val)
    649         cval = self.validate[key](val)
    650     except ValueError as ve:
--> 651         raise ValueError(f"Key {key}: {ve}") from None
    652     dict.__setitem__(self, key, cval)
    653 except KeyError as err:

ValueError: Key axes.linewidth: Could not convert 'thick' to float
```

One possibility that I'm undecided on is whether there should be more direct access to seaborn's named themes, i.e. something like

```python
so.Plot.config.theme.set(style="ticks", context="paper")
```

I'm a little unsure of how/where best to document this, but for now I'm adding a configuration section right on the `Plot` API docs page.

cf. https://github.com/mwaskom/seaborn/discussions/3005

### DIFF
--- a/doc/_docstrings/objects.Plot.config.ipynb
+++ b/doc/_docstrings/objects.Plot.config.ipynb
@@ -1,0 +1,124 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a38a6fed-51de-4dbc-8d5b-4971d06acf2e",
+   "metadata": {
+    "tags": [
+     "hide"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "import seaborn.objects as so"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "id": "38081259-9382-4623-8d67-09aa114e0949",
+   "metadata": {},
+   "source": [
+    "Theme configuration\n",
+    "^^^^^^^^^^^^^^^^^^^\n",
+    "\n",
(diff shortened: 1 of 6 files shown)


RSelvaratnam:
is so.plot available in the stable release?

PR #3234, opened 2023-01-24
Retrieve list of available datasets from a github url with correct CORS headers set.

koenvo:
This PR is a fix for https://github.com/mwaskom/seaborn-data/issues/24. It assumes there is a file called `dataset_names.txt` in the `seaborn-data` repository containing all the available dataset names. Each line should contain the name of the file without extension.

### DIFF
--- a/seaborn/utils.py
+++ b/seaborn/utils.py
@@ -19,7 +19,7 @@ from seaborn.external.version import Version
 from seaborn.external.appdirs import user_cache_dir
 
 __all__ = ["desaturate", "saturate", "set_hls_values", "move_legend",
-           "despine", "get_dataset_names", "get_data_home", "load_dataset", "DATASET_NAMES_URL"]
+           "despine", "get_dataset_names", "get_data_home", "load_dataset"]
 
 DATASET_NAMES_URL = "https://raw.githubusercontent.com/mwaskom/seaborn-data/master/dataset_names.txt"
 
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks — I think we'll need to do a PR on the other repo first for this to work?

seaborn/utils.py
```diff
@@ -19,7 +19,9 @@
 from seaborn.external.appdirs import user_cache_dir
 
 __all__ = ["desaturate", "saturate", "set_hls_values", "move_legend",
-           "despine", "get_dataset_names", "get_data_home", "load_dataset"]
+           "despine", "get_dataset_names", "get_data_home", "load_dataset", "DATASET_NAMES_URL"]
```
mwaskom, line 22: Wh add to `__all__`? I don't think this should be exposed as public API.

koenvo:
> Thanks — I think we'll need to do a PR on the other repo first for this to work?

Yes! Let me create one.

seaborn/utils.py
```diff
@@ -21,6 +21,8 @@
 __all__ = ["desaturate", "saturate", "set_hls_values", "move_legend",
            "despine", "get_dataset_names", "get_data_home", "load_dataset"]
 
+DATASET_NAMES_URL = "https://raw.githubusercontent.com/mwaskom/seaborn-data/master/dataset_names.txt"
```
mwaskom, line 24: Suggested refactor that will also fix the lint errors:

```suggestion
DATASET_SOURCE = "https://raw.githubusercontent.com/mwaskom/seaborn-data/master"
DATASET_NAMES_URL = f"{DATASET_SOURCE}/dataset_names.txt"
```

And then in `load_dataset` change use `DATASET_SOURCE` to build the url for the csv file.

mwaskom:
Thanks @koenvo

PR #3261, opened 2023-02-15
New marks: Axline, Axhline, Axvline. 

matulad:
This PR adds some new objects to the great Seaborn Objects API. 
It's my first contribution to `seaborn` package so I'm happy to hear your thoughts.

## New objects

* **`so.Axhline()`**
  * Implements a horizontal line defined by `y` coordinate (mappableFloat).
  * I've been thinking about utilizing `yintercept` (as is used by `ggplot2::geom_hline()` I'm familiar with), but the `y` seems to be a bit more convenient & momrable).
  * It inherits form `so.Path()` object so it's capable to use all its aesthetic mappings, too.
  * A docstring provided.
```{python}
import seaborn as sns
import seaborn.objects as so

tips["tip_slope"] = tips["tip"]/tips["total_bill"]

p = (
  so.Plot(tips, x="total_bill", y="tip", color="sex")
  .add(so.Dot(pointsize=1, color="black"))
)

( 
  p
  .add(so.Axhline(y=1, color="red"))  # adding fixed horizontal line
  .add(so.Axhline(), so.Agg(group_by_orient=False))  # adding subpopulation-level average; for group_by_orient see bellow
)
```

* **`so.Axvline()`**
  * Implements vertical line mark. The implementation details are pretty much the same as for the `so.Axhline()`.
  * The only difference: the position of the line is given by `x` property.
  * A docstring provided.
```{python}
(
  p
  .add(so.Axvline(x=1, color="red"))   # adding fixed vertical line
  .add(so.Axvline(), so.Agg(group_by_orient=False), orient="y")  # adding subpopulation-level average
)
```

* **`so.Axline()`**
  * Implements arbitrary line mark. 
  * The mark is specified using `intercept` and `slope`. 
  * Docstring provided.
```{python}
(
    p
    .add(so.Axline(intercept=1, slope=0.1, color="red"), legend=None)  # adding arbitrary line
    .add(so.Axline(intercept=0), so.AggCustom({"slope": "mean"}), slope="tip_slope") 
)
```

Besides, I've prepared and utilized two more aggregation objects.  
* **`so.AggCustom`**
  * This enables us to aggregate more than one column at a time.  
    * Originaly, it has been developed as a supporting structure for the `so.Axline()` object - to enable aggregating `intercept` and `slope` properties. 
    * (Moreover, I've been considering using e.g. `xintercept` instead of `x` aesthetics for the `so.Axvline()` because that is the way I'm used to from `geom_vline()`. In that case, the `so.AggCustom({"intercept": "mean", "slope": "mean"}) would be necessary, too. ... Although I have made other impl. decision, I believe we may find the `AggCustom()` helpful while implementing other marks..)
  * The `func` can be either:
    * the same as we use in `Agg` --> it will be applied for value property (and other property, too, if `group_by_orient` is set to `False` -- se bellow for details). 
    * dict of str/callables --> it's passed to the `GroupBy.agg()` method.
    * ~list of str/callables~ __Not yet.__ It works when no other grouping takes a place. In case of nonempty grouper, it produces multiheader in `groupby.py`, row 104 which needs to be reshaped. I haven't followed this path yet because there is no clean way how user can interact with the output (that is, the column describing which `func` is related to given value) e.g. to produce line labels or map line properties  
  * Another parameter is `group_by_orient` (defaults to `True`) which turns off grouping by orient property (handy for continuous orient variables, ofc).
  * A docstring is provided 
```{python}
# THIS IS NOT SUPPORTED, yet. But it works if there is no grouping at use:

# example 1: this doesn't work as 'color': 'sex' is applied.
p.add(so.Axhline(), so.AggCustom({"y": ["min", "mean", "max"]}))  

# example 2: this works, no other grouping
so.Plot(tips, "total_bill", "tip").add(so.Axhline(), so.AggCustom({"y": ["min", "mean", "max"]})) 
```
* **`so.Agg2d`**
  * This is just a shortcut for `so.AggCustom({"x": func, "y": func})`.
  * Strictly speaking, I do not have some good usecase for this one, yet (it has been developed in the stage where I've been experimenting with `so.Axline()` parametrization using `x` and `y` properties). But it works just fine and I believe it will show up to be handy, too :). 
  * A docstring provided.
```{python}
(
  p
  .add(so.Axhline(), so.Agg2d())
  .add(so.Axvline(), so.Agg2d())
)
```

## Changes 

* **`so.Agg` now enables to change `group_by_orient` to `False` (defaults to `True`).**
  *  The usecase: now we're able aggregate the `y` over the whole population to get e.g. population-level (or subgroup-level) mean in case the `x` axis is continuous:
  * In further development, i'd love to make it scale-dependent (False for continuous, True for discrete). But this is up to another discussion...

```{python}
p.add(so.Axhline(), so.Agg(group_by_orient=False))
```

* **`GroupBy.agg()` now implements empty grouper, too.**
  * Originally, It was throwing `ValueError` & there was a note: _We will need to see whether there are valid usecases that end up here_
  * The usecase: I'd like to get population-level stat:
```{python}
(
  so.Plot(tips, x="total_bill", y="tip")
  .add(so.Dot(pointsize=1, color="black"))
  .add(so.Axhline(), so.Agg(group_by_orient=False))  # this used to throw the error
)
```
 
* **`GroupBy._get_groups()` drops grouping variables IF present among aggregated columns**
  * Yeah, this one may be a bit controversal, but still I find this very usefull tool.
  * It solves the problem of providing custom aesthetics and applying aggregations on them (an example follows).
    * Another option would be to adjust grouping-assembling process. But this seemed to be a better option.
  * It enables me to do not bother with `group_by_orient` that much (e.g., examples with `so.Agg2d()`). 
```{python}
p.add(so.Axline(slope=0), so.AggCustom({"intercept": "mean"}), intercept="tip")
```

## Tests

* There is only one test not-passing: `tests/_core/test_groupby/test_at_least_one_grouping_variable_required` and it's of the obvious reason. I've got rid of it.
* No unit tests have been added so far. I'd like to hear your opinion first before investing my effort..

Ught, I hope I've covered all the main changes. 
As I mentioned above, I'm pretty new in contributing to your package. I'd be glad to hear your thoughts on proposed changes, too. 

Best,
Dominik


mwaskom:
Hi, thanks for your interest and enthusiasm. There's a lot going on in this PR. It's very hard to give a meaningful review to this much code touching so many different parts of the library.

As mentioned in the [contributing guidelines](https://github.com/mwaskom/seaborn/blob/242a3312b6742f5b255c7ad6bccb7228839c9419/.github/CONTRIBUTING.md#new-features), I really strongly prefer that people raise feature requests for discussion. I would have told you that this work overlaps in a number of ways with plans / prototypes on my roadmap, and (as you've encountered) it raises some tricky issues with the current architecture that require some careful thought. But while I'll revisit these questions as the objects interface, it's not my primary focus at this moment. Sorry, I'm sure that's rather discouraging.

thuiop:
As a side note : I have myself a similar object for plugging into the axline matplotlib API, which does both the horizontal and the vertical one ; one caveat is definitely that so.Agg does not work if there is no grouping variable, which means I may have to introduce an artificial one.
Also, although I see the use of the vertical and horizontal lines, I'm curious about what use case you have for the slope/intercept ones ?

matulad:
@thuiop - I'm using arbitrary lines quite a lot to either higlight some trend or to give some reference. 
Use Cases: 
* `y=x` line to higlight the difference while plotting scatter plot of related variables (e.g. paired measurements)
* add a regression line `y~ax+b` (in future, I've been thinking to add a stats object to do the calculation for you to not bother yourself with fitting model aside)
* ... and some other less common but still useful cases, e.g. visualising some constraint like `x+y < 7` etc.

@mwaskom - Oh, ok. Yeah, it's discouraging tbh. I've decided to finish my work first based on your reply in [the discussion](https://github.com/mwaskom/seaborn/discussions/3133). ... Anyway, would you mind to invest a bit of your time to review at least the isolated stuff (`so.Axhline()` and `so.Axvline()`)? It would be great to have at least something merged after the effort I've invested.. And I believe it would be handy for the API users, too.

Concerning the other stuff - ok, I see it somehow overlaps your future plans. But could you be a bit more specific? Is there a possibility to take a look on your roadmap to know with what can I help you with? Thanks a lot.

thuiop:
For cases like y=x or some arbitrary constraints, I see little point in having a seaborn object ; I usually make those using the matplotlib API directly. For regressions, I think you can just use so.PolyFit and so.Line ?

thuiop:
I believe PolyFit is basically a placeholder now, to provide basic functionality, but may change in the future, which is why the doc is incomplete.

As for having more objects, I think the answer is in #3133 : it is definitely intended to have more objects, but probably once the API is more stable. Adding the objects now would mean more maintenance and discourage changing the API when it may be necessary. Having them in the Discussions seems a good compromise to me : accessible to other users, but without the pressure of maintenance.

frfeng:
Rather than creating new marks, it seems not too bad to just use `so.Line` to add arbitrary lines?

```python
import seaborn as sns
import seaborn.objects as so

tips["tip_slope"] = tips["tip"]/tips["total_bill"]

p = (
  so.Plot(tips, x="total_bill", y="tip", color="sex")
  .add(so.Dot(pointsize=1, color="black"))
  .add(so.Line(color="red"), data=pd.DataFrame({'x':[0, 50]}).assign(y=1), x='x', y='y', color=None)
  .add(so.Line(color="blue"), data=pd.DataFrame({'y':[0, 10]}).assign(x=0), x='x', y='y', color=None)
  .add(so.Line(color="green"), data=pd.DataFrame({'x':[0, 50]}).assign(y=lambda df: 2 + 0.1 * df['x']), x='x', y='y', color=None)
)
```

matulad:
> Rather than creating new marks, it seems not too bad to just use `so.Line` to add arbitrary lines?

Thanks for your suggestion. Yeah, I know it's possible. But it forces me to prepare some temporal objects beforehand (see the example below) & I find it a bit clumpsy (contrary to elegant piping of `so`).

Usecase
- I'm plotting subgroups distributio (e.g. via KDE)
- I'd like to higlight group means using vertical line
- Now, I need to precalculate the values to be used there (or calculate it within the layer specification which is much even more clumsy)
- // moreover, In case of change of grouping you need to precalculate the temp df. It would be more convenient to have just the `x = mean(var | group)` mapping withing the so.Axvline geom...

frfeng:
Thanks. I'm not sure I see the example that you mentioned. A MWE might help for showing your use case.

PR #3264, opened 2023-02-16
New Mark: StepLine

matulad:
... OK, it's maybe a bit stupid thing to make a PR without issue for the second time, sorry for that. But I've already implemented and documented all necessary things before the feedback provided in [the other PR](https://github.com/mwaskom/seaborn/pull/3261). So I just put it here and wait for your feedback. I promiss to raise an issue in further occasions 😃 .

The PR is rather small and only adds strightforward way to __plot a stepwise function__. The inspiration to implement it comes from @thuiop's [suggestion](https://github.com/mwaskom/seaborn/discussions/3250). I thought it would be a bit more convenient to have a mark for step functions as ggplot2 ecosystem has the [`geom_step()`](https://ggplot2.tidyverse.org/reference/geom_path.html).

Usage:
```{python}
import seaborn.objects as so
from seaborn import load_dataset,
dowjones = load_dataset("dowjones")

so.Plot(dowjones.head(50), "Date", "Price").add(so.StepLine())
```


thuiop:
You may want to be careful with that, the ecdfplot used "steps-pre" for the reverse orientation, otherwise the plot will not be what you expect. I'm not sure there are many use cases for an ecdf plot on the y axis but you never know.

mwaskom:
Having a `Step` mark is on the roadmap but it is indeed not so simple because, at least in the ECDF case, the direction of the step should depend on the orientation (and probably be otherwise configurable). Also, a Step mark would ideally be capable of representing a histogram, but `Hist` currently emits the center coordinate and the width, which is not where you would want to step.

PR #3269, opened 2023-02-19
Fix typo in documentation of "seaborn.jointplot"

yvonnefroehlich:
In the documentation of [`seaborn.jointplot`](https://seaborn.pydata.org/generated/seaborn.jointplot.html) the docstring for the `hue`  parameter is stated twice. This PR removes the manual added text to show only the (identical) text set via [`_docstring.py`](https://github.com/mwaskom/seaborn/blob/master/seaborn/_docstrings.py).


mwaskom:
Nice catch, thanks!

PR #3290, opened 2023-03-08
Docs fix - heatmap.ipynb.

EitanHemed:
1. Using keyword-arguments on `.pivot` call, to avoid FutureWarning when pivoting only using positional arguments (see current version of [sns.heatmap docs](https://seaborn.pydata.org/generated/seaborn.heatmap.html#seaborn.heatmap#examples) - first example). To verify the validity of the change, this code can be used:
```
import seaborn as sns
import pandas as pd
positional = sns.load_dataset("glue").pivot("Model", "Task", "Score")
kw = sns.load_dataset("glue").pivot(index="Model", columns="Task", values="Score") 
pd.testing.assert_frame_equal(positional, kw)
```

2. Removed empty code cell in the end of the document.

### DIFF
--- a/doc/_docstrings/heatmap.ipynb
+++ b/doc/_docstrings/heatmap.ipynb
@@ -179,7 +179,7 @@
     "ax.set(xlabel=\"\", ylabel=\"\")\n",
     "ax.xaxis.tick_top()"
    ]
-  },
+  }
  ],
  "metadata": {
   "kernelspec": {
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks — wouldn't worry too much about the empty cell at the end of the notebook, it is removed when they are processed into docs. But also no problem in taking it out.

Issue #3230, opened 2023-01-21
TypeError with `dashes`  kwarg to lineplot since 0.12


matthew-brett:
We just noticed that code that used to work for 0.11, now causes a TypeError.

Here is a mimimal reproducible example:

```python
import numpy as np
import pandas as pd

import seaborn as sns

df = pd.DataFrame(np.random.normal(size=(10, 2)))

sns.lineplot(data=df, dashes=[(5, 5), (10, 10)])
```

This gives the expected plot with Seaborn < 0.12.  For versions 0.12 through current master (72faf27f), it gives:

```
Traceback (most recent call last):
  File "/Volumes/zorg/mb312/tmp/sns_tester.py", line 8, in <module>
    sns.lineplot(data=df, dashes=[(5, 5), (10, 10)])
  File "/Volumes/zorg/mb312/dev_trees/seaborn/seaborn/relational.py", line 645, in lineplot
    p.plot(ax, kwargs)
  File "/Volumes/zorg/mb312/dev_trees/seaborn/seaborn/relational.py", line 459, in plot
    lines = ax.plot(sub_data["x"], sub_data["y"], **kws)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/axes/_axes.py", line 1743, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/axes/_base.py", line 273, in __call__
    yield from self._plot_args(this, kwargs)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/axes/_base.py", line 418, in _plot_args
    return [func(x[:, j % ncx], y[:, j % ncy], kw, kwargs)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/axes/_base.py", line 418, in <listcomp>
    return [func(x[:, j % ncx], y[:, j % ncy], kw, kwargs)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/axes/_base.py", line 312, in _makeline
    seg = mlines.Line2D(x, y, **kw)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/lines.py", line 390, in __init__
    self.update(kwargs)
  File "/Users/mb312/Library/Python/3.9/lib/python/site-packages/matplotlib/artist.py", line 998, in update
[...]

PR #3316, opened 2023-04-08
Bugfix issue 3230

jonascristens:
Dear @mwaskom,

I would love to contribute to the amazing project, so I have tried to resolve issue [3230](https://github.com/mwaskom/seaborn/issues/3230).

Issue: The `kws` dict contained a list of tuples of `dashes` instead of single tuple for plotting.
Solution: Revert code partially back to changes made [here](https://github.com/mwaskom/seaborn/pull/2449/files)
Prevention: add an additional test to check for the existing issue

### DIFF
--- a/tests/test_relational.py
+++ b/tests/test_relational.py
@@ -680,6 +680,12 @@ class TestRelationalPlotter(Helpers):
         for text in g.legend.texts:
             assert float(text.get_text()) > 1e7
 
+    def test_lineplot_2d_dashes(self, long_df):
+
+        ax = lineplot(data=long_df[["x", "y"]], dashes=[(5, 5), (10, 10)])
+        assert ax.get_lines()[0]._us_dashSeq
+        assert ax.get_lines()[3]._us_dashSeq
+
 
 class TestLinePlotter(SharedAxesLevelTests, Helpers):
 
--- a/tests/test_relational.py
+++ b/tests/test_relational.py
@@ -683,8 +683,8 @@ class TestRelationalPlotter(Helpers):
     def test_lineplot_2d_dashes(self, long_df):
 
         ax = lineplot(data=long_df[["x", "y"]], dashes=[(5, 5), (10, 10)])
-        assert ax.get_lines()[0]._us_dashSeq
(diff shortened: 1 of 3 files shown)


mwaskom:
Thanks for taking a crack at this bug. It looks like you've run black over the files you've modified, which makes it very difficult to assess the functional changes to the code. Can you please revert the stylistic changes so the diff just reflects the bugfix? Thanks!

jonascristens:
The black changes have been reversed, thanks for the quick feedback!

tests/test_relational.py
```diff
@@ -680,6 +680,12 @@ def test_legend_has_no_offset(self, long_df):
         for text in g.legend.texts:
             assert float(text.get_text()) > 1e7
 
+    def test_lineplot_2d_dashes(self, long_df):
+
+        ax = lineplot(data=long_df[["x", "y"]], dashes=[(5, 5), (10, 10)])
+        assert ax.get_lines()[0]._us_dashSeq == (5, 5)
+        assert ax.get_lines()[3]._us_dashSeq == (10, 10)
```
mwaskom, line 687: Please do not access private attributes of third-party objects

mwaskom:
Thanks but I don't think this is the right solution. Better to handle the bug where it was introduced, in the process of reconciling parameters in the body of `lineplot`. I am pretty sure that you just need to change this line:

https://github.com/mwaskom/seaborn/blob/master/seaborn/relational.py#L599

There are two ways to have a defined `style` mapping: explicitly, where the local `style` variable is not None, and implicitly, when using wide-form data. The change added here to allow `dashes` to be a literal specification of dashes when not using a style mapping forgot about the second (implicit) way. So basically I think this just needs to change the check for `style is None` to `"style" not in p.variables`.

tests/test_relational.py
jonascristens: Solved thansk for the feedback!

jonascristens:
Thanks for the feedback, I've adjusted the code as requested. I now have a better understanding of what is happening exactly.

mwaskom:
Thanks!

Issue #3272, opened 2023-02-22
Layered pointplot getting clipped


mwaskom:
Following the refactor in #3270 

```python
sns.pointplot(
    x=pd.Series(["a", "b", "c"], index=[1, 2, 3]),
    y=pd.Series([1, 2, 3], index=[1, 2, 3])
)

sns.pointplot(
    x=pd.Series(["b", "c", "d"], index=[1, 2, 3]),
    y=pd.Series([3, 4, 5], index=[1, 2, 3])
)
```
<img width=300 src=https://user-images.githubusercontent.com/315810/220614667-ae1588d0-fc0d-4ba8-ae34-c79f6764d7da.png />

The same issue does not exist in `barplot`:

```python
sns.barplot(
    x=pd.Series(["a", "b", "c"], index=[1, 2, 3]),
    y=pd.Series([1, 2, 3], index=[1, 2, 3]),
)
sns.barplot(
    x=pd.Series(["b", "c", "d"], index=[1, 2, 3]),
    y=pd.Series([3, 4, 5], index=[1, 2, 3]),
    width=.5,
)
```
<img width=300 src=https://user-images.githubusercontent.com/315810/220614839-59419a8d-cee8-427e-a7a5-3d7bb0dcfeb7.png />

PR #3317, opened 2023-04-08
Fixed mwaskom/seaborn#3272

Jakal02:
This pull request fixes bug #3272. Now, sequential calls to seaborn.pointplot, where the oriented axis is categorical, will display all data properly. As a quick demo, here is the output of my code on the original example:
```
sns.pointplot(
    x=pd.Series(["a", "b", "c"], index=[1, 2, 3]),
    y=pd.Series([1, 2, 3], index=[1, 2, 3])
)

sns.pointplot(
    x=pd.Series(["b", "c", "d"], index=[1, 2, 3]),
    y=pd.Series([3, 4, 5], index=[1, 2, 3])
)
```

Inside `categorical.py::plot_points`, the variables `positions` is used to reindex the data about to be plotted, which is needed to include null-values for certain categories. However, `positions` took the explicit list of categorical variables passed in, and assigned them values starting from 0, instead of their correct positions with respect to the existing image.

In the above example, during the second call, the Series `["b", "c", "d"]` would set `positions=[0, 1, 2]`, when it should have been `positions = [1,2,3]`. This caused the point for `d` to not be graphed.

If the second series was `["b", "c"]`, only the data for `"b"` would show up, and if the second series was `["c", "d"]`, a spot for `d` would appear, but no data for `c` or `d` would be plotted.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -683,11 +683,12 @@ class _CategoricalPlotterNew(_RelationalPlotter):
         markers = self._map_prop_with_hue("marker", markers, "o", plot_kws)
         linestyles = self._map_prop_with_hue("linestyle", linestyles, "-", plot_kws)
 
+        ax = self.ax
+
         positions = self.var_levels[self.orient]
         if self.var_types[self.orient] == "categorical":
-            min_cat_val = int(self.comp_data[self.orient].min())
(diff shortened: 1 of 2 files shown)


mwaskom:
This look right to me, thanks!

Issue #3277, opened 2023-02-24
Feature Request for Polars library Support on Seaborn


Killpit:
Polars is one of the recent Python library that has all the functionalities in Pandas while being much faster. However, as I looked at the documentation from both Polars library and seaborn, I couldn't find anything other than having to use to_pandas() method.

If polars library support is added to seaborn, it becomes much easier and faster to create visualisations without losing time and the data frame becomes much tidier.

thuiop:
This would basically involve rewriting the whole library...

jcmkk3:
> This would basically involve rewriting the whole library...

It depends on what "support" means. It seems like the best short-term solution would be to read dataframes with pandas [`api.interchange.from_dataframe`](https://pandas.pydata.org/docs/reference/api/pandas.api.interchange.from_dataframe.html#pandas.api.interchange.from_dataframe) function. This should allow seaborn to read data from any libraries that implement the [dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html) (polars just being one [example](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.__dataframe__.html))

This won't perform the compute with polars, but it will make things more seamless for the users. With newer versions of pandas and the right settings (I'm not sure exactly what those are), it should be possible to make them zero copy with the arrow format and since pandas is starting to perform more of its compute with arrow, there wouldn't even be much of a performance difference compared to polars.

mwaskom:
Yeah, this isn't a very clear request .

> The best short-term solution would be to read dataframes with pandas [api.interchange.from_dataframe](https://pandas.pydata.org/docs/reference/api/pandas.api.interchange.from_dataframe.html#pandas.api.interchange.from_dataframe) function.

That is indeed on the roadmap for seaborn v0.13 now that it's landed in pandas stable.

> If polars library support is added to seaborn, it becomes much easier and faster to create visualisations without losing time and the data frame becomes much tidier.

[...]

PR #3340, opened 2023-04-25
Support interchange protocol

MarcoGorelli:
In https://github.com/mwaskom/seaborn/issues/3277#issuecomment-1444886408, it was mentioned

> That [api.interchange.from_dataframe] is indeed on the roadmap for seaborn v0.13 now that it's landed in pandas stable

so I figured I'd open a PR to drive the conversation forwards

Just wanted to check:
- would you still be open to this?
- given that pandas 2.0.0 and 2.0.1 have some small bugs in the interchange protocol (for categorical columns, and those with missing values - the rest should already work) would it be OK to set `2.0.2` (probably out mid-May) as the minimum pandas version to try interchanging from?

Trying this out with the latest pandas nightly, the `docs/tutorial` notebooks all seem to "just work" with non-pandas DataFrames - here's an example:

image


mwaskom:
Yes, definitely still open to this, but I don't think that adding it to every interface function is the right approach. Most functions call down to one (of a couple, with gradual standardization) internal functions for parsing the data specification.

> given that pandas 2.0.0 and 2.0.1 have some small bugs in the interchange protocol (for categorical columns, and those with missing values - the rest should already work) would it be OK to set 2.0.2 (probably out mid-May) as the minimum pandas version to try interchanging from?

I'm not sure I see the argument for gating it by pandas version. If someone passes a DataFrame that pandas 2.0.0 can handle, why not let it? (Also I thought that the exchange protocol was introduced in 1.5?)

seaborn/utils.py
```diff
@@ -889,3 +891,13 @@ def _disable_autolayout():
 def _version_predates(lib: ModuleType, version: str) -> bool:
     """Helper function for checking version compatibility."""
     return Version(lib.__version__) < Version(version)
+
+
+def try_convert_to_pandas(data: object | None) -> pd.DataFrame:
```
mwaskom, line 896: Why `object | None` and not `Any`?

seaborn/utils.py
```diff
@@ -889,3 +891,13 @@ def _disable_autolayout():
...
+    if data is None:
+        return None
+    elif isinstance(data, pd.DataFrame):
```
mwaskom, line 899: Is this guard important? will passing a `pandas.DataFrame` to ` pd.api.interchange.from_dataframe` be costly? I'd expect it to be a no-op without looking closer.

mwaskom:
Also is there an analogous solution for `pd.Series`? I guess many other DataFrame libraries don't 'have a concept of an "indexed" column so many not — those should go through the array protocol? I ask because most seaborn functions can accept vector data passed directly to `x`/`y`/etc. and I'd want to be similarly flexible.

seaborn/utils.py
MarcoGorelli: you're right, thanks, it is in-fact a no-op

https://github.com/pandas-dev/pandas/blob/360bf218d68c703911731aec58a52b6501b2f4ce/pandas/core/interchange/from_dataframe.py#L48-L49

However, seaborn still supports versions of pandas older than those which support the interchange protocol, so I introduced this to keep it a no-op in such cases

seaborn/utils.py
MarcoGorelli: just because `Any` turns off the type checker so I tend to avoid using it unless I have to, whereas as `object` prevents me from making assumptions about what properties the variable might have

MarcoGorelli:
Looks like the `x=<vector>, y=<vector>` syntax already works for non-pandas Series, as they're converted to numpy arrays anyway

https://github.com/mwaskom/seaborn/blob/54c36b74bbd4f7dfd5f933fcdc5e81fbb3fc0a45/seaborn/regression.py#L49-L50

For example, the following just works

image

MarcoGorelli:
Thanks - what's a more modern function you'd recommend testing? The notebooks from `docs/_tutorial` seem to work, e.g.

image

MarcoGorelli:
I tried amending `load_dataset` to return a non-pandas (polars) dataframe, and executing all the notebooks in `doc/_tutorial`, and it all worked

I've opened a PR to my own fork demonstrating this - it's easiest to see the diff, and the outputs, with reviewnb: https://app.reviewnb.com/MarcoGorelli/seaborn/pull/2/

I had to use the latest pandas nightly (in order to get some recent fixes which will be in version 2.0.2), which I installed with:
```
pip uninstall pandas -y
pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple pandas
```

To execute all the notebooks, I just did
```
git ls-files doc/_tutorial/*.ipynb | xargs jupyter nbconvert --to notebook --execute --inplace
```

seaborn/_oldcore.py
```diff
@@ -922,7 +924,7 @@ def _assign_variables_longform(self, data=None, **kwargs):
                     val in data
                     or (isinstance(val, (str, bytes)) and val in index)
                 )
-            except (KeyError, TypeError):
+            except (KeyError, TypeError, ValueError):
```
MarcoGorelli, line 927: if `val` is a non-pandas Series, then checking `val in data` will throw `ValueError`

tests/conftest.py
```diff
@@ -1,9 +1,26 @@
+import os
+
 import numpy as np
 import pandas as pd
 
 import pytest
 
 
+def maybe_convert_to_polars(df):
```
MarcoGorelli, line 9: updating several fixtures to return `polars.DataFrame` instead of `pandas.DataFrame`

MarcoGorelli:
I've added a CI job to test this, in which the fixtures are amended to return a polars DataFrame instead of a pandas one (where possible)

There are currently some CI failures in that job, which also appear when using the latest pandas nightly - might have to wait for pandas 2.0.2 to come out then, to use that in this CI workflow. Fortunately, that's expected to be quite soon, on Monday

mwaskom:
Hi @MarcoGorelli this is a very helpful proof of concept but just a heads up that I don't think I want to maintain this sort of test-everything-on-polars approach on an ongoing basis, so no need to kill yourself tying up every edge case in the way that tests are written.

MarcoGorelli:
😄 sure, I'll see if there's a simpler way round this, thanks

MarcoGorelli:
closing in favour of https://github.com/mwaskom/seaborn/pull/3369

PR #3359, opened 2023-05-10
TST: Make test_index_alignment_between_series less brittle

StefRe:
`test_index_alignment_between_series` fails on Windows but passes on Linux. The reason is that the indices of the expected Series are created using the default dtype of an integer np.arange which is int64 on Linux but int32 on Windows (see for instance https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine). The DataFrame default RangeIndex is always of dtype int64.  

With this PR, the range is created with an explicit dtype to make the test independent of the default integer type.

### DIFF
--- a/tests/_core/test_data.py
+++ b/tests/_core/test_data.py
@@ -187,8 +187,9 @@ class TestPlotData:
 
         p = PlotData(None, {"x": x, "y": y})
 
-        x_col_expected = pd.Series([10, 20, 30, np.nan, np.nan], np.arange(1, 6))
-        y_col_expected = pd.Series([np.nan, np.nan, 300, 400, 500], np.arange(1, 6))
+        idx_expected = [1, 2, 3, 4, 5]
+        x_col_expected = pd.Series([10, 20, 30, np.nan, np.nan], idx_expected)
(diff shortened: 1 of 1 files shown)


tests/_core/test_data.py
```diff
@@ -187,8 +187,9 @@ def test_index_alignment_between_series(self):
 
         p = PlotData(None, {"x": x, "y": y})
 
-        x_col_expected = pd.Series([10, 20, 30, np.nan, np.nan], np.arange(1, 6))
-        y_col_expected = pd.Series([np.nan, np.nan, 300, 400, 500], np.arange(1, 6))
+        idx_expected = np.arange(1, 6, dtype=pd.RangeIndex(0).dtype)
```
mwaskom, line 190: Would it be less roundabout to use `np.intp` here? It is meant to represent the platform-appropriate indexing integer.

mwaskom: No strong preference just thought it looked weird to instantiate an index just to grab its dtype. You could also just use a `RangeIndex` as the `idx_expected`...

StefRe: I eventually went for the simplest variant `[1, 2, 3, 4, 5]` which corresponds to the indices used for the series.

PR #3366, opened 2023-05-16
DOC: Fix some typos

StefRe:
There are also varying spellings for
* program***m***atically/programatically (6 and 2 occurences, repsectively)
* paramet***e***rize/parametrize (6 and 5 occurences, the latter is the version used by `pytest`)

I guess this is an American vs British spelling thing, not sure which version you want to stick with?


mwaskom:
Thanks!

PR #3369, opened 2023-05-21
Support polars and other data libraries via dataframe interchange

### LINKED ISSUE #3368: seaborn.object: `TypeError: `x` given by both name and position` raised by polars and not by pandas
I am plotting a simple histogram from the result of a `.value_counts()`, so I just have two columns (see collapsible element)

```
┌───────────────────────────────────┬────────┐
│ source                            ┆ counts │
│ ---                               ┆ ---    │
│ str                               ┆ u32    │
╞═══════════════════════════════════╪════════╡
│ Professioni sanitarie - Milano (… ┆ 40     │
│ Veterinaria - 2003                ┆ 25     │
│ Veterinaria - 2002                ┆ 31     │
│ Odontoiatria - 2010               ┆ 40     │
│ Professioni sanitarie - Bologna … ┆ 28     │
└───────────────────────────────────┴────────┘
```

The code I use for plotting is:

```python
(
    prod.select(pl.col("content").struct.field("source").value_counts())
    .unnest("source")
    .pipe(so.Plot, x="source", y="counts")
    .add(so.Bar())
)
```

But this error is raised: `TypeError: 'x' given by both name and position.`

<details>
<summary><i>Full traceback</i></summary>

```
Cell In[30], line 2
      1 (
----> 2     prod.select(pl.col("content").struct.field("source").value_counts())
      3     .unnest("source")
      4     # .to_pandas()
      5     .pipe(so.Plot, x="source", y="counts")
      6     .add(so.Bar())
      7 )

File [~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/polars/dataframe/frame.py:4448](https://file+.vscode-resource.vscode-cdn.net/Users/lucabaggi/Documents/dev/futura/futura-grader/notebooks/~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/polars/dataframe/frame.py:4448), in DataFrame.pipe(self, function, *args, **kwargs)
   4381 def pipe(
   4382     self,
   4383     function: Callable[Concatenate[DataFrame, P], T],
   4384     *args: P.args,
   4385     **kwargs: P.kwargs,
   4386 ) -> T:
   4387     """
   4388     Offers a structured way to apply a sequence of user-defined functions (UDFs).
   4389 
   (...)
   4446 
   4447     """
-> 4448     return function(self, *args, **kwargs)

File [~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/seaborn/_core/plot.py:208](https://file+.vscode-resource.vscode-cdn.net/Users/lucabaggi/Documents/dev/futura/futura-grader/notebooks/~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/seaborn/_core/plot.py:208), in Plot.__init__(self, data, *args, **variables)
    200 def __init__(
    201     self,
    202     *args: DataSource | VariableSpec,
    203     data: DataSource = None,
    204     **variables: VariableSpec,
    205 ):
    207     if args:
--> 208         data, variables = self._resolve_positionals(args, data, variables)
    210     unknown = [x for x in variables if x not in PROPERTIES]
    211     if unknown:

File [~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/seaborn/_core/plot.py:262](https://file+.vscode-resource.vscode-cdn.net/Users/lucabaggi/Documents/dev/futura/futura-grader/notebooks/~/Documents/dev/futura/futura-grader/.venv/lib/python3.9/site-packages/seaborn/_core/plot.py:262), in Plot._resolve_positionals(self, args, data, variables)
    260 if var is not None:
    261     if name in variables:
--> 262         raise TypeError(f"`{name}` given by both name and position.")
    263     # Keep coordinates at the front of the variables dict
    264     # Cast type because we know this isn't a DataSource at this point
    265     variables = {name: cast(VariableSpec, var), **variables}

TypeError: `x` given by both name and position.
```
</details>

However, when I transform to pandas, the plot is drawn correctly.

```python
(
    prod.select(pl.col("content").struct.field("source").value_counts())
    .unnest("source")
    .to_pandas()
    .pipe(so.Plot, x="source", y="counts")
    .add(so.Bar())
)
```

mwaskom:
This PR leverages the [dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html) to let seaborn consume objects from other dataframe libraries, such as polars.

Dataframes are converted to pandas objects upon consumption, and that is what is used internally by seaborn, so seaborn's statistical operations don't take advantage of any parallelism / out of core / etc. functionality offered by these libraries. While that would be ideal, I don't see it happening any time soon. 

Nevertheless, this should make it easy to prep data in a library of choice and then pipe it to seaborn without thinking too much about the representation.

For testing, my approach is to use a simple mock object that is not (by inheritance) a pandas DataFrame, but that does "support" the interchange protocol. I think this is sufficient, with the assumption that pandas / other data libraries will together correctly implement the dataframe interchange itself. Testing whether that works correctly feels out of scope for seaborn's unit tests.

I wouldn't be surprised to learn of various edge cases as this roles out to people using alternative dataframe libraries heavily (I only did some light testing with polars using toy datasets) so we'll address those as they happen.

Thanks to @MarcoGorelli for getting the ball rolling with #3340 and advising on the approach.
Closes #3368

### DIFF
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -61,7 +61,12 @@ class PlotData:
         self.names = names
         self.ids = ids
 
-        self.frames = {}  # TODO this is a hack, remove
+        # The reason we possibly have a dictionary of frames is to support the
+        # Plot.pair operation, post scaling, where each x/y variable needs its
+        # own frame. This feels pretty clumsy and there are a bunch of places in
+        # the client code with awkard if frame / elif frames constructions.
+        # It would be great to have a cleaner abstraction here.
+        self.frames = {}
 
         self.source_data = data
         self.source_vars = variables
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -282,7 +282,7 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
         )
         raise TypeError(msg)
 
-    if _version_predates(pd, "2.0.2 "):
+    if _version_predates(pd, "2.0.2"):
         msg = (
             "DataFrame interchange with pandas<2.0.2 has some known issues. "
             f"You are using pandas {pd.__version__}. "
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -304,6 +304,12 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
         warnings.warn(msg, stacklevel=2)
 
     try:
+        # This is going to convert all columns in the input dataframe, even though
+        # we may only need one or two of them. It would be more efficient to select
+        # the columns that are going to be used in the plot prior to interchange.
+        # Solving that in general is a hard problem, especially with the objects
+        # interface where variables passed in Plot() may only be referenced later
+        # in Plot.add(). But noting here in case this seems to be a bottleneck.
         return pd.api.interchange.from_dataframe(data)
     except Exception as err:
         msg = (
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -52,6 +52,14 @@ class PlotData:
         variables: dict[str, VariableSpec],
     ):
 
+        if isinstance(data, pd.DataFrame) or hasattr(data, "__dataframe__"):
+            # Check for pd.DataFrame inheritance could be removed once
+            # minimal pandas version supports dataframe interchange (1.5.0).
+            data = convert_dataframe_to_pandas(data)
+        elif data is not None and not isinstance(data, Mapping):
+            err = f"Data source must be a DataFrame or Mapping, not {type(data)!r}."
+            raise TypeError(err)
+
         frame, names, ids = self._assign_variables(data, variables)
 
         self.frame = frame
@@ -75,7 +83,7 @@ class PlotData:
         variables: dict[str, VariableSpec] | None,
     ) -> PlotData:
         """Add, replace, or drop variables and return as a new dataset."""
-        # Inherit the original source of the upsteam data by default
+        # Inherit the original source of the upstream data by default
         if data is None:
             data = self.source_data
 
@@ -118,7 +126,7 @@ class PlotData:
 
     def _assign_variables(
         self,
-        data: DataSource,
+        data: DataFrame | Mapping | None,
         variables: dict[str, VariableSpec],
     ) -> tuple[DataFrame, dict[str, str | None], dict[str, str | int]]:
         """
@@ -147,6 +155,8 @@ class PlotData:
 
         Raises
         ------
+        TypeError
+            When data source is not a DataFrame or Mapping.
         ValueError
             When variables are strings that don't appear in `data`, or when they are
             non-indexed vector datatypes that have a different length from `data`.
@@ -162,15 +172,12 @@ class PlotData:
         ids = {}
 
         given_data = data is not None
-        if data is not None:
-            source_data = data
-        else:
+        if data is None:
             # Data is optional; all variables can be defined as vectors
             # But simplify downstream code by always having a usable source data object
             source_data = {}
-
-        # TODO Generally interested in accepting a generic DataFrame interface
-        # Track https://data-apis.org/ for development
+        else:
+            source_data = data
 
         # Variables can also be extracted from the index of a DataFrame
         if isinstance(source_data, pd.DataFrame):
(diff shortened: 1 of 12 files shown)


MarcoGorelli:
thanks for the ping

I tried this out, but got a failure for the `doc/_docstrings/FacetGrid.ipynb` notebook:
```python
tips = pl.from_pandas(sns.load_dataset('tips'))
g = sns.FacetGrid(tips, col="time",  row="sex")
g.map(sns.scatterplot, "total_bill", "tip")
```

```python-traceback
SchemaError                               Traceback (most recent call last)
Cell In[4], line 2
      1 g = sns.FacetGrid(tips, col="time",  row="sex")
----> 2 g.map(sns.scatterplot, "total_bill", "tip")

File ~/seaborn-dev/seaborn/axisgrid.py:720, in FacetGrid.map(self, func, *args, **kwargs)
    717         warnings.warn(warning)
    719 # Iterate over the data subsets
--> 720 for (row_i, col_j, hue_k), data_ijk in self.facet_data():
    721 
    722     # If this subset is null, move on
    723     if not data_ijk.values.size:
    724         continue

File ~/seaborn-dev/seaborn/axisgrid.py:674, in FacetGrid.facet_data(self)
    670 # Here is the main generator loop
    671 for (i, row), (j, col), (k, hue) in product(enumerate(row_masks),
    672                                             enumerate(col_masks),
    673                                             enumerate(hue_masks)):
--> 674     data_ijk = data[row & col & hue & self._not_na]
    675     yield (i, j, k), data_ijk

File ~/seaborn-dev/.venv/lib/python3.10/site-packages/polars/series/series.py:439, in Series.__and__(self, other)
    437 if not isinstance(other, Series):
    438     other = Series([other])
--> 439 return self._from_pyseries(self._s.bitand(other._s))

SchemaError: cannot unpack series of type `list[bool]` into `bool`
```

seaborn/_core/data.py
```diff
@@ -258,3 +265,26 @@ def _assign_variables(
...
+            "Support for non-pandas DataFrame objects requires a version of pandas"
+            "that implements the DataFrame interchange protocol. Please upgrade "
+            "your pandas version or coerce your data to pandas before passing "
+            "it to seaborn."
+        )
+        raise TypeError(msg)
+
+    try:
+        return pd.api.interchange.from_dataframe(data)
+    except Exception as err:
```
MarcoGorelli, line 285: I'd strongly suggest setting 2.0.2 as the minimum version to try interchanging from, because there's some pretty basic mistakes in earlier versions:
```python
In [1]: df = pl.DataFrame({'a': [1,2,3]})

In [2]: pd.api.interchange.from_dataframe(df[1:])
Out[2]:
                 a
0                2
1                3
2  125822987010162
```
(😱 )

mwaskom:
> I tried this out, but got a failure for the doc/_docstrings/FacetGrid.ipynb notebook

Right, this just addresses the [objects interface](https://seaborn.pydata.org/tutorial/objects_interface.html); the older code will need to be handled separately.

seaborn/_core/data.py
mwaskom: Ah, I didn't realize there were silent failure modes.

mwaskom: I do sort of continue to think that it's not seaborn's business to pin behaviors on specific upstream iterations, but I guess "sometimes wrong" is worse than "sometimes errors"...

MarcoGorelli:
I see, thanks!

I tried
```
git ls-files doc/_docstrings/objects*.ipynb | xargs jupyter nbconvert --to notebook --execute --inplace
git ls-files doc/_tutorial/objects_interface.ipynb | xargs jupyter nbconvert --to notebook --execute --inplace
```
and both commands pass without errors (in my `demo-support-interchange-protocol-2`, where I make `load_dataset` return a polars DataFrame)

seaborn/_core/data.py
MarcoGorelli: sure, your decision (obvs), just pointing this out because as the conversion happens internally, it wouldn't be obvious to users that the issue is in pandas, and so you'd likely get tonnes of issues in your issue tracker

seaborn/_core/data.py
```diff
@@ -258,3 +265,26 @@ def _assign_variables(
...
+        msg = (
+            "Support for non-pandas DataFrame objects requires a version of pandas"
+            "that implements the DataFrame interchange protocol. Please upgrade "
+            "your pandas version or coerce your data to pandas before passing "
+            "it to seaborn."
+        )
+        raise TypeError(msg)
+
+    try:
+        return pd.api.interchange.from_dataframe(data)
```
MarcoGorelli, line 284: just sharing this plotly PR for reference: https://github.com/plotly/plotly.py/pull/4286/files

if it's possible to tell in advance which columns need converting to pandas, then it may be more efficient only select them before hand using `__dataframe__().select_columns_by_name`

mwaskom: In general that is somewhere between difficult and hard. For instance, as currently written, when you do `Plot(data).add(so.Dots(), x="x", y="y')`, the data interchange will happen when the `Plot()` constructor executes but the columns that are needed from the constructor aren't known until `Plot.add()` executes.

I've added a comment to the code noting that this is a known inefficiency but I don't think there's much that can be done about it save a large rewrite of how data ingest works so I am going to say we need to live with it for now.

MarcoGorelli: sounds good, thanks for explaining!

PR #3400, opened 2023-06-23
Change get_linestyle to get_dashes for Collections

oscargus:
There is an upcoming change in Matplotlib that unifies the return of `get_linestyle`. Instead, based on the current tests, one should use `get_dashes`. (Will produce a warning for a few releases and then return the linestyle string as other artists.)

~Also added a test as there is `get_dashes` for lines (now).~ Too long since I did the MPL PR clearly, but there will be a `get_dashes` as part of the PR.

(Couldn't run the tests locally because of issues with pywintypes so please bear with me in case it is not working out of the box...)

### DIFF
--- a/tests/_marks/test_line.py
+++ b/tests/_marks/test_line.py
@@ -181,7 +181,7 @@ class TestPaths:
 
         assert same_color(lines.get_color().squeeze(), m.color)
         assert lines.get_linewidth().item() == m.linewidth
-        assert lines.get_linestyle()[0] == (0, list(m.linestyle))
+        assert lines.get_dashes()[0] == (0, list(m.linestyle))
 
     def test_mapped_properties(self):
 


mwaskom:
Thanks @oscargus. Can you link to the relevant matplotlib pr / discussion so i can understand the context here better?

oscargus:
Sure, should have done that from the start: https://github.com/matplotlib/matplotlib/pull/23056

A few things are happening with the goal to unify the handling of line styles (same return types, same entry supported, same error checking).

On the MPL dev-call yesterday the conclusion was that we will warn when using `Collection.get_linestyle` and refer to `get_dashes` instead for a number of releases and since Seaborn was one of the packages that we could identify that used this, I thought I'd give some heads up to possibly avoid getting warning for a release or so.

Seeing the failed test, I realize that some additional thinking is required... (But most changes survived...)

oscargus:
Pushed a new version with the working part. This should work for recent MPL versions as well as forthcoming with or without the proposed unification (and no warning for collections).

mwaskom:
Thanks!

PR #3424, opened 2023-07-21
Address inf_as_na pandas deprecation

mroeschke:
Will be deprecated in pandas 2.1 https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html#deprecations

### DIFF
--- a/seaborn/_base.py
+++ b/seaborn/_base.py
@@ -1121,13 +1121,12 @@ class VectorPlotter:
                 parts = []
                 grouped = self.plot_data[var].groupby(self.converters[var], sort=False)
                 for converter, orig in grouped:
-                    with pd.option_context('mode.use_inf_as_na', True):
-                        orig = orig.dropna()
-                        if var in self.var_levels:
-                            # TODO this should happen in some centralized location
-                            # it is similar to GH2419, but more complicated because
-                            # supporting `order` in categorical plots is tricky
-                            orig = orig[orig.isin(self.var_levels[var])]
+                    orig = orig.replace(np.inf, np.nan).dropna()
+                    if var in self.var_levels:
+                        # TODO this should happen in some centralized location
+                        # it is similar to GH2419, but more complicated because
+                        # supporting `order` in categorical plots is tricky
+                        orig = orig[orig.isin(self.var_levels[var])]
                     comp = pd.to_numeric(converter.convert_units(orig)).astype(float)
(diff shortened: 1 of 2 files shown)


seaborn/_base.py
```diff
@@ -1121,13 +1121,12 @@ def comp_data(self):
...
                 grouped = self.plot_data[var].groupby(self.converters[var], sort=False)
                 for converter, orig in grouped:
-                    with pd.option_context('mode.use_inf_as_na', True):
-                        orig = orig.dropna()
-                        if var in self.var_levels:
-                            # TODO this should happen in some centralized location
-                            # it is similar to GH2419, but more complicated because
-                            # supporting `order` in categorical plots is tricky
-                            orig = orig[orig.isin(self.var_levels[var])]
+                    orig = orig.replace(np.inf, np.nan).dropna()
```
mwaskom, line 1124: What about negative inf?

mroeschke: Good call. Added `-np.inf`

mwaskom:
It looks like this is still failing on a test case with a datetime series. I don't know exactly what the issue is but looking into it raises a behavior I don't understand.

Here's a simple example:

```python
t = pd.date_range("2010-01-01", periods=3, freq="d").to_series()
t.iloc[1] = np.inf
t.replace(np.inf, np.nan)
```
```
2010-01-01   2010-01-01
2010-01-02          NaT
2010-01-03   2010-01-03
Freq: D, dtype: datetime64[ns]
```

That does what I expect (nice of pandas to coerce the `np.nan` to `NaT`).

But with negative infinity, it isn't properly coerced to a missing value and, weirder, the output suddenly has an object dtype?

```python
t = pd.date_range("2010-01-01", periods=3, freq="d").to_series()
t.iloc[1] = np.inf
t.replace(-np.inf, np.nan)
```
```
2010-01-01    2010-01-01 00:00:00
2010-01-02                    inf
2010-01-03    2010-01-03 00:00:00
Freq: D, dtype: object
```

mwaskom:
@mroeschke any idea why this is still failing? FWIW seaborn tests are now failing on the pandas rc as a result of the change so it would be good to sort this out.

mroeschke:
Okay got this to pass. I think the tested version of pandas (1.2) has a bug where `replace` with inf didn't work, but the mask solution here should be equivalent

mwaskom:
Nice!

PR #3425, opened 2023-07-22
Fix the type of `Plot.add` `transforms` argument to accept `Move`, not `Mark`

qiemem:
Fixes the type in accordance with the intention of #2948. Presumably, this was just an autocomplete typo since docs and functionality all specify `Move` rather than `Mark`.


mwaskom:
Oops, nice catch, thanks.

PR #3427, opened 2023-07-23
seaborn.objects: Add formatting string argument to Text mark

qiemem:
Adds a `fmt` argument to the `so.Text` mark, allowing the user to pass in a [format string](https://docs.python.org/3/library/string.html#formatstrings) to format the text before display. I also considered having it just take in a function, similar to `.label`'s arguments. Let me know if that would be preferred.

### DIFF
--- a/doc/_docstrings/objects.Text.ipynb
+++ b/doc/_docstrings/objects.Text.ipynb
@@ -156,9 +156,31 @@
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


JanPalasek:
@mwaskom is there some problems with this PR?
For instance I find myself needing this functionality in a lot of plots. To bypass this, I use apply with pandas to create a new formatted column.

PR #3436, opened 2023-08-07
Hash-pin GitHub Actions

### LINKED ISSUE #3435: Hash-pin GitHub Actions
Hey, it's Pedro and I'm back (see #3319 and #3344) with another security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes will be automatically updated by dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

I'll send a PR pinning the Actions along with this issue.

pnacht:
Fixes #3435.

This PR hash-pins GitHub Actions to protect the project from supply-chain attacks.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,10 +21,10 @@ jobs:
   build-docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Makes sense. Thanks!

PR #3446, opened 2023-08-24
Update tests for matplotlib 3.8.0rc1

ksunden:
This is mostly a generalization of the tests to work with both old and new matplotlib

The only production code change is adding `.flat` to one line in the matrix annotation code, all other changes are test-only.

Mostly the test changes were allowing the collection being found to be a `QuadContourSet`, as well as allowing outputs like colors or coordinates to be sequences instead of singular values, or adding `.flat` to values which now get returned as input-shaped arrays instead of automatically flattened arrays.

I was able to avoid version checking for most things, aside from the minor tick expected value and explicitly checking the expected type in `ax.collections`. Though there are some version-checks by proxy of `isinstance` checks.

I did add a boolean flag to shift the filtering of empty paths from the test to the helper function for `get_contour_coords`, this is only used in one test for now, but made for a backwards compatible filtering step.

`get_contour_coords` will now return non-rectangular grids (Not actually clear to me that this is _new_) so some tests have additional loops before calling `assert_array_equal` I added assertions that the lengths were the same to avoid `zip` dropping the fact that one list was longer.

I have tested with both matplotlib v3.8.x (the backport branch for the 3.8 series) as well as released 3.7.2.

In my dev environment I am seing some numpy warnings (which are elevated to errors for 2 tests), but other than that all tests pass.

### DIFF
--- a/seaborn/matrix.py
+++ b/seaborn/matrix.py
@@ -252,7 +252,7 @@ class _HeatMapper:
         height, width = self.annot_data.shape
         xpos, ypos = np.meshgrid(np.arange(width) + .5, np.arange(height) + .5)
         for x, y, m, color, val in zip(xpos.flat, ypos.flat,
-                                       mesh.get_array(), mesh.get_facecolors(),
+                                       mesh.get_array().flat, mesh.get_facecolors(),
                                        self.annot_data.flat):
             if m is not np.ma.masked:
                 lum = relative_luminance(color)
--- a/tests/_core/test_scales.py
+++ b/tests/_core/test_scales.py
@@ -191,9 +191,11 @@ class TestContinuous:
 
         n = 3
         a = self.setup_ticks(x, count=2, minor=n)
-        # I am not sure why matplotlib's minor ticks include the
-        # largest major location but exclude the smalllest one ...
-        expected = np.linspace(0, 1, n + 2)[1:]
+        expected = np.linspace(0, 1, n + 2)
+        if _version_predates(mpl, "3.8.0rc1"):
+            # I am not sure why matplotlib <3.8  minor ticks include the
+            # largest major location but exclude the smalllest one ...
+            expected = expected[1:]
         assert_array_equal(a.minor.locator(), expected)
 
     def test_log_tick_default(self, x):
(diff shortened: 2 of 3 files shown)


mwaskom:
Thanks @ksunden for jumping in here with the quick help — I can confirm that this passes all test using the rc on my laptop too!

mwaskom:
Ah dang, not sure these test updates are all correct. There was a (sometimes triggered) accession of the `contours.collections` inside seaborn's main codebase that was populating the collections list as previously expected. So the `filter_empty` addition allowed tests to pass using that data, but not once I fixed the deprecated (as opposed to broken) pattern :(

ksunden:
Ah, darn, that was the trickiest bit that I was not quite as confident of (but it _appeared_ to work for both old and new versions of mpl, so I went with it).


If I understand what's happening, the following patch should work:

```diff
diff --git a/tests/test_distributions.py b/tests/test_distributions.py
index 3b8418fb..54de2b04 100644
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -2438,13 +2438,13 @@ class TestDisPlot:
         z = [0] * 80 + [1] * 20

         g = displot(x=x, y=y, col=z, kind="kde", levels=10)
-        l1 = sum(bool(get_contour_coords(c)) for c in g.axes.flat[0].collections)
-        l2 = sum(bool(get_contour_coords(c)) for c in g.axes.flat[1].collections)
+        l1 = sum(len(get_contour_coords(c, True)) for c in g.axes.flat[0].collections)
+        l2 = sum(len(get_contour_coords(c, True)) for c in g.axes.flat[1].collections)
         assert l1 > l2

         g = displot(x=x, y=y, col=z, kind="kde", levels=10, common_norm=False)
-        l1 = sum(bool(get_contour_coords(c)) for c in g.axes.flat[0].collections)
-        l2 = sum(bool(get_contour_coords(c)) for c in g.axes.flat[1].collections)
+        l1 = sum(len(get_contour_coords(c, True)) for c in g.axes.flat[0].collections)
+        l2 = sum(len(get_contour_coords(c, True)) for c in g.axes.flat[1].collections)
         assert l1 == l2

     def test_bivariate_hist_norm(self, rng):
```

i.e. using the `filter_empty` behavior for one additional test, and changing from `bool` to `len`, as there may be more than one now

(I just commented out the line which accessed `cset.collections`, so whatever the proper fix on that side is up to you :))

PR #3447, opened 2023-08-25
Squash a large number of warnings in the tests

mwaskom:
Various small edits to:

- Maintain forward compatibility with upcoming changes in dependencies
- Avoid triggering some numpy warnings around missing data and discrete mapping logic
- Silence a couple of innocuous warnings there's no clear way to work around

All that remains here are https://github.com/matplotlib/matplotlib/pull/26300, https://github.com/matplotlib/matplotlib/issues/26596 and ~https://github.com/pytest-dev/pytest-xdist/issues/840~ (decided to ignore the latter in the test config, since there's no apparent movement towards dealing with it upstream despite lots of reports).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,5 @@ exclude = ["doc/_static/*.svg"]
 
 [tool.pytest.ini_options]
 filterwarnings = [
-   "error",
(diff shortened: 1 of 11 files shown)


mwaskom:
Merging over codecov as it's just flagging a line that is version gated to a forthcoming matplotlib release.

PR #3452, opened 2023-08-27
Deprecate accepting an array-typed palette, which is not intentional

### LINKED ISSUE #3450: Warning "passing `palette` without assigning `hue`"
**Context**
Seems related to https://github.com/mwaskom/seaborn/issues/2429, but [the associated gist](https://gist.github.com/mwaskom/9937c53eeb8b9794a450bf1f20b03e2f) didn't help.

I aim to produce a `barplot` with the color varying on the continuous y-axis. It works fine, except that I can't get rid of a new warning message:

```
Passing `palette` without assigning `hue` is deprecated and will be removed in v0.14.0. Assign the `x` variable to `hue` and set `legend=False` for the same effect.

  ax = sns.barplot(
/Users/vincentmaladiere/mambaforge/envs/skrub/lib/python3.10/site-packages/seaborn/_base.py:206: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if palette in QUAL_PALETTES:
```

I use the 0.13.0.dev0 version of seaborn on the following script:
```python
import numpy as np
import seaborn as sns


x = np.array([0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0])
y = np.array([17583, 16411, 14014, 10706, 14455, 11458, 16209])

norm = plt.Normalize(y.min(), y.max())
cmap = plt.get_cmap("magma")

ax = sns.barplot(
    x=x,
    y=y,
    palette=cmap(norm(y)),
)
plt.xticks(rotation=30)
plt.ylabel(None)
plt.tight_layout()
```
With this output:
Screen Shot 2023-08-27 at 10 08 47

**Expected behavior**
No warning messages.

**Version**
Python 3.10.11
matplotlib 3.7.1
seaborn 0.13.0.dev0

mwaskom:
Closes #3450

We don't document ndarray as a valid type for `palette` but it has worked by happenstance previously. Some changes in numpy are causing one of the operations we perform on the palette object to break. This PR (1) works around those numpy changes by catching array types and coercing them to lists, and (2) fires a warning about upcoming removal of support of array-typed palettes.

Note that arrays are already disallowed in the objects interface:
```
TypeError: Scale values for color with a Nominal mapping must be string, list, tuple, or dict; not <class 'numpy.ndarray'>.
```

### DIFF
--- a/seaborn/_base.py
+++ b/seaborn/_base.py
@@ -108,6 +108,15 @@ class HueMapping(SemanticMapping):
 
         data = plotter.plot_data.get("hue", pd.Series(dtype=float))
 
+        if isinstance(palette, np.ndarray):
+            msg = (
+                "Numpy array is not a supported type for `palette`. "
+                "Please convert your palette to a list. "
+                "This will become an error in v0.14"
+            )
+            warnings.warn(msg, stacklevel=4)
+            palette = palette.tolist()
+
         if data.isna().all():
             if palette is not None:
                 msg = "Ignoring `palette` because no `hue` variable has been assigned."
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -451,6 +451,14 @@ class TestSizeMapping:
         with pytest.raises(ValueError):
             SizeMapping(p, sizes="bad_size")
 
+    def test_array_palette_deprecation(self, long_df):
+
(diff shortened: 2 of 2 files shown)


Vincent-Maladiere:
Thank you for this :)

mwaskom:
Thank *you* for testing out the development version!

PR #3513, opened 2023-10-05
No. but also...

beyarkay:
I'm all for opinionated colour palettes and funny comments, but it would be nice to get _some_ guidance towards goodness, instead of just a fence around the badness.


mwaskom:
No.  ;)

mwaskom:
(Maybe the joke is past its time — though I’m not really interested in debating that — but the point is not actually to prevent people from using jet or to educate them. It’s just an Easter egg, explaining the joke is both no longer funny and still unhelpful.)

PR #3516, opened 2023-10-08
FIX: handle potential TypeError when determining variable type

theOehrly:
I ran into an issue with Seaborn 0.13.0 that wasn't present in previous versions. The following code works fine in Seaborn 0.12.2 but raises ``numpy.core._exceptions._UFuncInputCastingError`` on 0.13.0.

```
from datetime import timedelta
import seaborn as sns

data = {
    'delta': [timedelta(1), timedelta(5), timedelta(3)],
    'x': [1, 2, 3]
}

sns.scatterplot(data, x='x', y='delta')
plt.show()

```

The reason are numpy's default casting rules and the fact that the check changed from ``np.isin(vector, [0, 1, np.nan]`` to ``np.isin(vector, [0, 1])`` in the latest release. That means that previously the comparison was with ``0.0`` and ``1.0`` as float dtype, because the ``nan`` made NumPy cast the whole test array to float. Now the comparison is against int dtype.

In my opinion, the actual dtypes are not relevant here. But we should not assume that the default casting rules allow that this comparison is possible. Simply because the data type of ``vector`` is not know as the data is provided by the user. 
Therefore, I'd say it is best to handle the ``_UFuncInputCastingError`` (which is a subclass of ``TypeError``) and in that case decide that the data is not binary/boolean. The then remaining checks should not have similar issues.

### DIFF
--- a/seaborn/_base.py
+++ b/seaborn/_base.py
@@ -1520,8 +1520,13 @@ def variable_type(vector, boolean_type="numeric"):
         warnings.simplefilter(
             action='ignore', category=(FutureWarning, DeprecationWarning)
         )
-        if np.isin(vector, [0, 1]).all():
-            return VariableType(boolean_type)
+        try:
+            if np.isin(vector, [0, 1]).all():
+                return VariableType(boolean_type)
+        except TypeError:
+            # .isin comparison is not guaranteed to be possible under NumPy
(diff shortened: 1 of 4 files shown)


mwaskom:
Thanks @theOehrly, great catch

PR #3534, opened 2023-10-19
fall back to to_pandas if interchange protocol fails

### LINKED ISSUE #3533: Try `to_pandas` rather than erroring if interchanging to pandas doesn't work?
Here's an example of some code which currently raises:
```python
import seaborn as sns
import polars as pl
df = pl.DataFrame({'a': [1,2,3], 'b': [4,5,6], 'c': [[1,2], [4,5], [6,7]]})
sns.catplot(
    data=df,
    x="a",
    y="b",
)
```

There's a really long error message, but the gist of it is
```
ValueError: data type List(Int64) not supported by the interchange protocol
```

Indeed, just doing
```python
pd.api.interchange.from_pandas(df)
```
would raise the same error

I'd like to suggest that, when converting to pandas, if the interchange protocol fails, there be a `to_pandas` fallback.

Because that at least works here (for reference, plotly do the same - try interchanging first, and if that fails, check if there's `to_pandas`, and if so, use that)

MarcoGorelli:
Closes #3533

### DIFF
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -312,11 +312,12 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
         # in Plot.add(). But noting here in case this seems to be a bottleneck.
         return pd.api.interchange.from_dataframe(data)
     except Exception as err:
-        if hasattr(data, 'to_pandas'):
-            try:
-                return data.to_pandas()  # type: ignore[attr-defined]
-            except Exception:
-                pass
+        try:
+            # An object's own to_pandas method may be more flexible than the interchange
+            # See https://github.com/mwaskom/seaborn/pull/3534 for more context
+            return data.to_pandas()  # type: ignore[attr-defined]
+        except Exception:
+            pass
         msg = (
             "Encountered an exception when converting data source "
             "to a pandas DataFrame. See traceback above for details."
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -312,8 +312,18 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
         # in Plot.add(). But noting here in case this seems to be a bottleneck.
         return pd.api.interchange.from_dataframe(data)
     except Exception as err:
-        msg = (
-            "Encountered an exception when converting data source "
-            "to a pandas DataFrame. See traceback above for details."
-        )
-        raise RuntimeError(msg) from err
+        if hasattr(data, 'to_pandas'):
+            try:
+                return data.to_pandas()  # type: ignore[attr-defined]
+            except Exception:
+                msg = (
+                    "Encountered an exception when converting data source "
+                    "to a pandas DataFrame. See traceback above for details."
+                )
+                raise RuntimeError(msg) from err
+        else:
+            msg = (
+                "Encountered an exception when converting data source "
+                "to a pandas DataFrame. See traceback above for details."
+            )
+            raise RuntimeError(msg) from err
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -316,14 +316,9 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
             try:
                 return data.to_pandas()  # type: ignore[attr-defined]
             except Exception:
-                msg = (
-                    "Encountered an exception when converting data source "
-                    "to a pandas DataFrame. See traceback above for details."
-                )
-                raise RuntimeError(msg) from err
-        else:
-            msg = (
-                "Encountered an exception when converting data source "
-                "to a pandas DataFrame. See traceback above for details."
-            )
-            raise RuntimeError(msg) from err
+                pass
+        msg = (
+            "Encountered an exception when converting data source "
+            "to a pandas DataFrame. See traceback above for details."
+        )
+        raise RuntimeError(msg) from err
--- a/tests/_core/test_data.py
+++ b/tests/_core/test_data.py
@@ -430,6 +430,19 @@ class TestPlotData:
             PlotData(mock_long_df, {"x": "x"})
 
     @pytest.mark.skipif(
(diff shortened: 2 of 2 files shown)


mwaskom:
I've got no issues with this but it would be helpful to understand better when it's relevant. Is the idea that a library might have bugs in the hooks that `frame_dataframe`calls but not in `to_pandas`? Do we often expect those to go through very different code paths?

seaborn/_core/data.py
```diff
@@ -312,8 +312,18 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
...
         return pd.api.interchange.from_dataframe(data)
     except Exception as err:
-        msg = (
-            "Encountered an exception when converting data source "
-            "to a pandas DataFrame. See traceback above for details."
-        )
-        raise RuntimeError(msg) from err
+        if hasattr(data, 'to_pandas'):
+            try:
+                return data.to_pandas()  # type: ignore[attr-defined]
```
mwaskom, line 317: j/w but can type checkers not figure out that this attribute must exist in this context?

seaborn/_core/data.py
```diff
@@ -312,8 +312,18 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
...
+            except Exception:
+                msg = (
+                    "Encountered an exception when converting data source "
+                    "to a pandas DataFrame. See traceback above for details."
+                )
+                raise RuntimeError(msg) from err
```
mwaskom, line 323: Not wild about the copy-paste here, I guess there's no super clean way to write this sort of nested exception handling, but can we please at least just define `msg` once?

(I guess technically you don't even need to raise from the inner `except` block, because the outer raise statement doesn't need to be guarded by an `else`. You could either return from `to_pandas` or just `pass` on any exception that comes out of it and end up back at the original `raise` statement. Not 100% sure that is cleaner / more explicit though).

mwaskom (changes requested):

MarcoGorelli:
> I've got no issues with this

Great, thanks!

> Is the idea that a library might have bugs in the hooks that frame_dataframe calls but not in to_pandas?

A library can control its own `to_pandas` method, but for `pd.api.interchange.from_dataframe`, they're tied down to whatever decision pandas (and the dataframe consortium) make.

In particular, if there's a datatype not supported by the consortium, then interchanging to pandas will raise regardless of whether that column will be used or not. Doesn't look like there's any real prospect of this changing https://github.com/data-apis/dataframe-api/issues/288

_maybe_ I can get an argument add to pandas (like, `from_dataframe(..., ignore_unsupported_columns: bool)`)

mwaskom (approved):

mwaskom:
Thanks for explaining. TBH as an observer (but interested party) these nuances make it fairly confusing to understand what the "right" way to generically consume a dataframe object is. Not sure that that's really actionable feedback (and based on the linked thread it looks like you agree). So this feels slightly hacky but I think reasonable given the circumstances.

MarcoGorelli:
I think the ideal way would be something like this:

https://github.com/plotly/plotly.py/blob/57e4d1d33c67c5cc715bec5c3c240dd6f4c3b10d/packages/python/plotly/plotly/express/_core.py#L1428-L1431

```python
                columns = list(necessary_columns)
                df = pd.api.interchange.from_dataframe(
                    df.__dataframe__().select_columns_by_name(columns)
                )
```
but I understand that knowing in advance exactly which columns you need might not be possible without a major refactor

mwaskom:
I gather that a relevant example looks like this:

```python
import polars as pl
pol_df = pl.DataFrame({
    "a": pl.Series([1, 2, 3, 4]),
    "b": pl.Series([[1, 2], [3], [4, 5, 6], []]),
})

# Works, "b" column is object-typed
pol_df.to_pandas()

# Raises with "large_list<item: int64> not supported by interchange protocol"
pd.api.interchange.from_dataframe(pol_df)
```

(Actually, first-class list-type support would be pretty handy for seaborn 🤔...)

This makes sense to me as a current source of disagreement that we should work around.

seaborn/_core/data.py
```diff
@@ -312,6 +312,11 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
         # in Plot.add(). But noting here in case this seems to be a bottleneck.
         return pd.api.interchange.from_dataframe(data)
     except Exception as err:
+        if hasattr(data, 'to_pandas'):
+            try:
+                return data.to_pandas()  # type: ignore[attr-defined]
+            except Exception:
+                pass
```
mwaskom, line 319:
```suggestion
        try:
            # A data object's own to_pandas method may be more flexible than the interchange logic
            # See https://github.com/mwaskom/seaborn/pull/3534 for more context
            return data.to_pandas()  # type: ignore[attr-defined]
        except Exception:
            pass
```

krz:
I'm really hoping for a fix, because this is an annoying problem I sometimes encounter.
I'm an R convert who loves both `polars` and `seaborn.objects`

For example, this works without `to_pandas()`

```
import seaborn.objects as so
import polars as pl

penguins = pl.read_csv("https://raw.githubusercontent.com/mwaskom/seaborn-data/master/penguins.csv")

(
    so.Plot(penguins, x="bill_length_mm", y="bill_depth_mm")
    .add(so.Dot())
)
```

However, this does not work:

```
economics = pl.read_csv("https://raw.githubusercontent.com/tidyverse/ggplot2/main/data-raw/economics.csv", try_parse_dates=True, dtypes={"pop": pl.Float32})

(
    so.Plot(economics, x="date", y="uempmed")
    .add(so.Path())
)
```

`to_pandas()` works:

```
(
    so.Plot(economics.to_pandas(), x="date", y="uempmed")
    .add(so.Path())
)
```

MarcoGorelli:
hi @mwaskom - just for my understanding, is there anything holding this back? or will it go in with the 0.14 release?

PR #3535, opened 2023-10-21
Print list of palettes on invalid palette choice

beyarkay:
(Just a small thing, feel free to reject.)

This PR simply prints out some valid colour palettes if the user provides an invalid one. It doesn't print out _every_ palette.

### DIFF
--- a/seaborn/palettes.py
+++ b/seaborn/palettes.py
@@ -60,6 +60,7 @@ QUAL_PALETTES = list(QUAL_PALETTE_SIZES.keys())
 
 class _ColorPalette(list):
     """Set the color palette in a with statement, otherwise be a list."""
+
     def __enter__(self):
         """Open the context."""
         from .rcmod import set_palette
@@ -93,6 +94,7 @@ class _ColorPalette(list):
 
 def _patch_colormap_display():
     """Simplify the rich display of matplotlib color maps in a notebook."""
+
     def _repr_png_(self):
         """Generate a PNG representation of the Colormap."""
         import io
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks for this, but I'm not sure it's helpful to provide a list of valid palettes that doesn't include all valid palettes.

And even though it doesn't print *every* palette (seaborn palette specification is programmatic and that wouldn't really be possible), it sure does print a lot of them. That produces a huge error message and I'm not sure the UX of searching through it is really a net benefit. Things would look even more overwhelming if e.g. you were using `colorcet` which adds several dozen additional palettes.

If the objective here is to help people who misspell a colormap (and I agree, it can be really hard to remember the exact spelling of those colorbrewer maps), I'd be in favor an approach that did a fuzzy match with a suggestion, similar to what Python is trying to do with its attribute errors lately. (E.g. "No palette named 'YlGrBu', did you mean 'YlGnBu'"?) Though of course that is trickier to implement!

beyarkay:
Yeah I did underestimate the number of available colour palettes a bit 😅 and looking at it with fresh eyes I'd agree that the massive list is a bit much and not super helpful.

I don't think seaborn has a dependency on a fuzzy matching library (correct me if I'm wrong) and I'm not sure I have time to implement one myself at the moment. It feels a bit overkill to bring in a dependency just for this error message, unless you think it would be useful to have in the future (?).

An alternative would be to link the user to either the seaborn "Choosing color palettes" docs or the matplotlib ["Colormap reference"](https://matplotlib.org/stable/gallery/color/colormap_reference.html).

mwaskom:
> I don't think seaborn has a dependency on a fuzzy matching library (correct me if I'm wrong) and I'm not sure I have time to implement one myself at the moment.

Correct, and I am also -1 on adding a new dependency on a third party library just for this. But Python itself is making these kind of suggestions as of 3.11 so it is worth looking into how that works and whether there is something in the stdlib that could be leveraged. What I know of off the top of my head is `difflib`, and a simple PoC using that is straightforward:

```python
import difflib
arg = "YlGrBu"
scores = pd.Series({name: difflib.SequenceMatcher(a=arg, b=name).ratio() for name in mpl.colormaps})
scores.sort_values(ascending=False).head()
```
```
YlGnBu      0.833333
YlGnBu_r    0.714286
YlGn_r      0.666667
YlOrBr      0.666667
RdYlBu      0.666667
```

I am not certain that this is the best way to do it, and there are a number of questions about how exactly to go about implementing this in the right way (e.g. you'd probably want a threshold on the score, things like `_r` should get special handling, etc.).

PR #3538, opened 2023-10-22
Update old mpl color tutorial link

beyarkay:
The old link https://matplotlib.org/2.0.2/users/colormaps.html is marked by matplotlib as out of date, so this replaces it with the new link https://matplotlib.org/stable/users/explain/colors/colormaps.html that seems to have basically the same information (although it did get shuffled around a wee bit).


mwaskom:
Thanks!

PR #3540, opened 2023-10-25
FIX #3537 Disabling the legend based on the x-axis and hue values

### LINKED ISSUE #3537: catplot with redundant hue assignment creates empty legend with title
```python
sns.catplot(tips, x="day", y="total_bill", hue="day", col="sex", row="smoker", kind="box", height=3)
```
image

Set `legend=False` to workaround, but with default `legend='auto'` this should be disabled due to the redundancy.

abdulwaheedsoudagar:
Disable legend in case of redundancy
`sns.catplot(tips, x="day", y="total_bill", hue="day", col="sex", row="smoker", kind="box", height=3)`

image

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2763,6 +2763,10 @@ def catplot(
         elif x is not None and y is not None:
             raise ValueError("Cannot pass values for both `x` and `y`.")
 
+    # Check for x and hue values are same
+    if x == hue:
+        legend = False
+
     p = Plotter(
         data=data,
         variables=dict(x=x, y=y, hue=hue, row=row, col=col, units=units),
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2763,10 +2763,6 @@ def catplot(
         elif x is not None and y is not None:
             raise ValueError("Cannot pass values for both `x` and `y`.")
 
-    # Check for x or y and hue values are same
-    if x == hue or y == hue:
-        legend = False
-
     p = Plotter(
         data=data,
         variables=dict(x=x, y=y, hue=hue, row=row, col=col, units=units),
@@ -3100,7 +3096,18 @@ def catplot(
         ax.legend_ = None
 
     if legend and "hue" in p.variables and p.input_format == "long":
-        g.add_legend(title=p.variables.get("hue"), label_order=hue_order)
+        # Obtaining the column names for hue, x, and y.
+        hue_value = p.variables.get("hue")
+        x_value = p.variables.get("x")
+        y_value = p.variables.get("y")
+        # Obtaining the datatype of hue
+        hue_type = p.var_types.get("hue")
+        if legend == 'auto' and hue_value in {x_value, y_value} and hue_type != 'numeric':
+            # Setting the title of hue to None if hue matches "x" or "y", or if the data type of "hue" is numeric.
+            title_hue = None
+        else:
+            title_hue = hue_value
+        g.add_legend(title=title_hue, label_order=hue_order)
 
     if data is not None:
         # Replace the dataframe on the FacetGrid for any subsequent maps
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2763,8 +2763,8 @@ def catplot(
         elif x is not None and y is not None:
             raise ValueError("Cannot pass values for both `x` and `y`.")
 
-    # Check for x and hue values are same
-    if x == hue:
+    # Check for x or y and hue values are same
+    if x == hue or y == hue:
         legend = False
 
     p = Plotter(
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -409,12 +409,10 @@ class _CategoricalPlotter(VectorPlotter):
                     data[col] = inv(data[col])
 
     def _configure_legend(self, ax, func, common_kws=None, semantic_kws=None):
-
         if self.legend == "auto":
             show_legend = not self._redundant_hue and self.input_format != "wide"
         else:
             show_legend = bool(self.legend)
-
         if show_legend:
             self.add_legend_data(ax, func, common_kws, semantic_kws=semantic_kws)
             handles, _ = ax.get_legend_handles_labels()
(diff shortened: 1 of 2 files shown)


mwaskom:
Thanks for taking a crack at this, but it's not a sufficient change. First, that simple comparison will raise if any of the variables are specified using vector data (i.e. pandas series or numpy arrays, where the `==` operator works element-wise and returns a vector). Second, we want to continue supporting a (redundant) legend when the user asks for it explicitly by passing `legend=True`. You'll want to take a look at how this is currently handled in the axes-level functions, i.e. in this method: https://github.com/mwaskom/seaborn/blob/863539d71e281d88a93de64581827b166d9e2f22/seaborn/categorical.py#L411

abdulwaheedsoudagar:
> Thanks for taking a crack at this, but it's not a sufficient change. First, that simple comparison will raise if any of the variables are specified using vector data (i.e. pandas series or numpy arrays, where the `==` operator works element-wise and returns a vector). Second, we want to continue supporting a (redundant) legend when the user asks for it explicitly by passing `legend=True`. You'll want to take a look at how this is currently handled in the axes-level functions, i.e. in this method:
> 
> https://github.com/mwaskom/seaborn/blob/863539d71e281d88a93de64581827b166d9e2f22/seaborn/categorical.py#L411

Thank you for responding. I have a brief inquiry. The problem of the legend not showing up occurs when the setting is "legend = auto." Yet, when it's set to "legend = full" or "legend = brief," it appears correctly, when hue and x values are same. Should the legend only be displayed when it's set to "True", or also when it's "full" or "brief"?

mwaskom:
> Should the legend only be displayed when it's set to "True", or also when it's "full" or "brief"?

`"True"` (as a string type) is not a valid argument for `legend` but `True` (as a boolean type) is. The only place where we want conditional logic about a redundant legend is when `legend="auto"`. Otherwise it should appear when requested (with `True`, `"brief"`, or `"full"`) and be suppressed when not (with `False`).

abdulwaheedsoudagar:
> > Should the legend only be displayed when it's set to "True", or also when it's "full" or "brief"?
> 
> `"True"` (as a string type) is not a valid argument for `legend` but `True` (as a boolean type) is. The only place where we want conditional logic about a redundant legend is when `legend="auto"`. Otherwise it should appear when requested (with `True`, `"brief"`, or `"full"`) and be suppressed when not (with `False`).

Thanks, Based on your comment I have modified the code, legend title is disabled based on x and y values and hue datatype. 
In function _configure_legend, legend title is not decided, it is decieded in line 3098, so I added conditions for this https://github.com/abdulwaheedsoudagar/seaborn/blob/d4ec331b57a58287b05cc49191b4b5b54797118f/seaborn/categorical.py#L3098 statement.

seaborn/categorical.py
```diff
@@ -3096,7 +3096,20 @@ def catplot(
...
+        # Obtaining the column names for hue, x, and y.
+        hue_value = p.variables.get("hue")
+        x_value = p.variables.get("x")
+        y_value = p.variables.get("y")
+        # Obtaining the datatype of hue
+        hue_type = p.var_types.get("hue")
+        if legend == 'auto' and hue_value in {x_value, y_value} \
+                and hue_type != 'numeric':
+            # Setting the title of hue to None if hue matches "x" or
+            # "y", or if the data type of "hue" is numeric.
```
mwaskom, line 3108: I don't think you don't need all of this complexity; as in `_configure_legend` you can take advantage of the `_redundant_hue` property to cover the edge case.

Additionally, the conditional should guard the entire `add_legend` call, not just what the title gets set to. The plot should not have a legend when `legend="auto"` and the hue variable is redundant.

mwaskom (changes requested):

seaborn/categorical.py
abdulwaheedsoudagar: Initially, I started with what you suggested and found that in the current implementation when the x and hue columns are the same `_redundant_hue ` is set as True and `show_legend `as False, which causes the legend data not to display, but the legend column is still displayed. so I went with modifying the title
image
Also if I comment the `if show_legend` code in function `_configure_legend`, even then legend title(that is day in this example) will be displayed. what I understood is in this function legend data is decided but not title
```
def _configure_legend(self, ax, func, common_kws=None, semantic_kws=None):

    if self.legend == "auto":
        show_legend = not self._redundant_hue and self.input_format != "wide"
    else:
        show_legend = bool(self.legend)

    if show_legend:
        self.add_legend_data(ax, func, common_kws, semantic_kws=semantic_kws)
        handles, _ = ax.get_legend_handles_labels()
        if handles:
            ax.legend(title=self.legend_title)
```

mwaskom: I am not suggesting you modify `_configure_legend`, I am suggesting you base the changes to the body of `catplot` off of it. I think you just need the following patch

```diff
diff --git a/seaborn/categorical.py b/seaborn/categorical.py
index bb40d5d3..be68617a 100644
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -3095,7 +3095,12 @@ def catplot(
         g._update_legend_data(ax)
         ax.legend_ = None
 
-    if legend and "hue" in p.variables and p.input_format == "long":
+    if legend == "auto":
+        show_legend = not p._redundant_hue and p.input_format != "wide"
+    else:
+        show_legend = bool(legend)
+
+    if show_legend:
         g.add_legend(title=p.variables.get("hue"), label_order=hue_order)
 
     if data is not None:
```

There is a bit of copy-paste from `_configure_legend` that could perhaps get abstracted away, but it's not a big deal.

abdulwaheedsoudagar: Ah, I overlooked that the class p contains the variable _redundant_hue. I've made the appropriate changes. Thanks

mwaskom:
Can you please add a test? Essentially you want to assert that the `FacetGrid` returned by `catplot` under this condition has a null `legend` attribute.

abdulwaheedsoudagar:
> Can you please add a test? Essentially you want to assert that the `FacetGrid` returned by `catplot` under this condition has a null `legend` attribute.

Sure, I will add

mwaskom (approved):

mwaskom:
Thanks @abdulwaheedsoudagar

Issue #3511, opened 2023-10-02
lineplot of empty dataframe with hue in seaborn 0.13.0


maximilianmordig:
MWE
```
df1 = pd.DataFrame({}, columns=["aa", "bb", "cc"]) # empty dataframe
# df1 = pd.DataFrame([(1, 2, 3), (2, 1, 3)], columns=["aa", "bb", "cc"]) # with this, it works
sns.lineplot(df1, x="aa", y="bb") # works
sns.lineplot(df1, x="aa", y="bb", hue="cc") # does not work
```

Error happens with seaborn 0.13.0, but not with 0.12.2:

Error:
```
File .../python3.10/site-packages/seaborn/relational.py:507, in lineplot(data, x, y, hue, size, style, units, palette, hue_order, hue_norm, sizes, size_order, size_norm, dashes, markers, style_order, estimator, errorbar, n_boot, seed, orient, sort, err_style, err_kws, legend, ci, ax, **kwargs)
    504 color = kwargs.pop("color", kwargs.pop("c", None))
    505 kwargs["color"] = _default_color(ax.plot, hue, color, kwargs)
--> 507 p.plot(ax, kwargs)
    508 return ax

File .../python3.10/site-packages/seaborn/relational.py:274, in _LinePlotter.plot(self, ax, kws)
    266 # TODO How to handle NA? We don't want NA to propagate through to the
    267 # estimate/CI when some values are present, but we would also like
    268 # matplotlib to show "gaps" in the line when all values are missing.
   (...)
    271 
    272 # Loop over the semantic subsets and add to the plot
    273 grouping_vars = "hue", "size", "style"
--> 274 for sub_vars, sub_data in self.iter_data(grouping_vars, from_comp_data=True):
    276     if self.sort:
    277         sort_vars = ["units", orient, other]

File .../python3.10/site-packages/seaborn/_base.py:938, in VectorPlotter.iter_data(self, grouping_vars, reverse, from_comp_data, by_facet, allow_empty, dropna)
    935 for var in grouping_vars:
    936     grouping_keys.append(levels.get(var, []))
--> 938 iter_keys = itertools.product(*grouping_keys)
    939 if reverse:
    940     iter_keys = reversed(list(iter_keys))

TypeError: 'NoneType' object is not iterable
```

mwaskom:
Fixed by #3569

mwaskom:
Thanks for reporting with a simple MWE btw!

PR #3569, opened 2023-11-22
[Fix issue][#3511]lineplot of empty dataframe with hue in seaborn 0.13.0

Noko-Github:
Corresponds to the case where grouping_var has None object.

Related issue: https://github.com/mwaskom/seaborn/issues/3511

### DIFF
--- a/seaborn/_base.py
+++ b/seaborn/_base.py
@@ -933,8 +933,8 @@ class VectorPlotter:
 
             grouping_keys = []
             for var in grouping_vars:
-                key = levels.get(var) if levels.get(var) is not None else []
-                grouping_keys.append(key)
+                key = levels.get(var)
+                grouping_keys.append([] if key is None else key)
 
             iter_keys = itertools.product(*grouping_keys)
             if reverse:
--- a/seaborn/_base.py
+++ b/seaborn/_base.py
@@ -933,10 +933,8 @@ class VectorPlotter:
 
             grouping_keys = []
             for var in grouping_vars:
-                if key := levels.get(var):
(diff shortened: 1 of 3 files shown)


mwaskom:
Thanks for taking a crack at this. Can you please add a test case that would fail with the current behavior (and pass with your fix)? Thanks!

seaborn/_base.py
```diff
@@ -933,7 +933,8 @@ def iter_data(
 
             grouping_keys = []
             for var in grouping_vars:
-                grouping_keys.append(levels.get(var, []))
+                key = levels.get(var) if levels.get(var) is not None else []
+                grouping_keys.append(key)
```
mwaskom, line 937: Total nitpick, but the duplicate call to `levels.get` above looks cumbersome, maybe move the ternary down here? e.g.

```suggestion
                key = levels.get(var)
                grouping_keys.append([] if key is None else key)
```

mwaskom (changes requested):
Thanks, a couple nitpicks but looks like the right approach.

Noko-Github:
Thank you for your review!
I incorporated feedback.

mwaskom:
Thanks @Noko-Github 

I think it's possible that this could also be fixed upstream by making the default `levels` value an empty list rather than `None`. I am not certain whether that would cause any other problems though and this is simple and seems to work so 👍

mwaskom (approved):

mwaskom:
Btw if you make the PR message say "Closes #issue" or "Fixes #issue" it will close the issue automatically when merged.

Issue #3566, opened 2023-11-20
Problems when setting positions in boxplot() (mainly on log-scale axis)


leoluecken:
Hey everybody,
Thanks for adding `native_scale` to boxplot in 0.13!! I've been waiting for this! :)
Now, I tried to do some tweaking dodging boxpositions manually, and encountered the following (I'm on 0.13.0, and matplotlib 3.7.2):

My code:
```python
import seaborn as sns
import matplotlib.pyplot as plt

data = {
    "x":[0.1,0.1,0.1,0.1,1,1,1,1],
    "y":[1,2,3,4,1,2,3,4]
}

xvals = sorted(set(data["x"]))

# logscale dodging
boxpositions = [np.exp(np.log(x)+0.5) for x in xvals]

fig, ax = plt.subplots()
ax.set_xscale("log")
sns.boxplot(data, x="x", y="y", width=0.3, native_scale=True, positions=boxpositions, ax=ax)
plt.show()
```

produces the following image:

![image](https://github.com/mwaskom/seaborn/assets/6482401/c8823211-8a36-4fb5-80f7-6dd419a40781)

This doesn't get better when trying to use log_scale instead of native_scale:

```python
sns.boxplot(data, x="x", y="y", width=0.3, log_scale=True, positions=boxpositions, ax=ax)
```
→

![image](https://github.com/mwaskom/seaborn/assets/6482401/87bbfb63-c96d-46d2-b010-d0b089c94a54)

It works more or less fine on linear scale:

```python
import seaborn as sns
import matplotlib.pyplot as plt

data = {
    "x":[0.1,0.1,0.1,0.1,1,1,1,1],
    "y":[1,2,3,4,1,2,3,4]
}

xvals = sorted(set(data["x"]))

# linscale dodging
boxpositions = [x+0.5 for x in xvals]

fig, ax = plt.subplots()
sns.boxplot(data, x="x", y="y", width=0.3, positions=boxpositions, ax=ax)
plt.show()
```

Only the xlim is not updated:

![image](https://github.com/mwaskom/seaborn/assets/6482401/2bc05e99-bb88-44ed-866d-823c3392cd49)

Cheers,
Leo

mwaskom:
The problem is that `positions` is a keyword argument of the underlying matplotlib function. Matplotlib itself has no idea how to draw a boxplot on a log axis, seaborn needs to do all of the positional adjustments itself to make things like dodging on a log scale work (it's quite complicated and annoying).

[...]

PR #3576, opened 2023-11-29
Rejecting parameter 'positions' for boxplot(), refs #3566

leoluecken:
Hi, this is a first attempt for the boxplot position rejection (#3566). I'm not sure about the style. Please review and let me know if something should be changed. 
Cheers

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -1577,6 +1577,10 @@ def boxplot(
     fliersize=None, hue_norm=None, native_scale=False, log_scale=None, formatter=None,
     legend="auto", ax=None, **kwargs
 ):
+    
+    if "positions" in kwargs:
+        msg = "boxplot() does not support parameter 'positions'. Consider to use native_scale=True and specify positions via parameter 'x'."
+        raise ValueError(msg)
 
     p = _CategoricalPlotter(
         data=data,
--- a/tests/test_categorical.py
+++ b/tests/test_categorical.py
@@ -1168,7 +1168,25 @@ class TestBoxPlot(SharedAxesLevelTests, SharedPatchArtistTests):
         g = catplot(**kwargs, kind="box")
 
         assert_plots_equal(ax, g.ax)
-
+        
+        
+    @pytest.mark.parametrize("positions", [None, [1, 2, 3, 4]])
+    def test_reject_boxpositions(self, positions):
+        
+        kwargs = {
+            "x":[1, 2, 3, 4],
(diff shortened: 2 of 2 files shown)


seaborn/categorical.py
```diff
@@ -1577,6 +1577,10 @@ def boxplot(
     fliersize=None, hue_norm=None, native_scale=False, log_scale=None, formatter=None,
     legend="auto", ax=None, **kwargs
 ):
+    
+    if "positions" in kwargs:
+        msg = "boxplot() does not support parameter 'positions'. Consider to use native_scale=True and specify positions via parameter 'x'."
```
mwaskom, line 1582:
```suggestion
        msg = "boxplot does not support parameter `positions`. Consider using `native_scale=True`.
```
Nit: For a horizontal boxplot you'd specify the positions with `y` so let's make the suggestion less specific. Note that I've also tweaked the formatting.

seaborn/categorical.py
```diff
@@ -1577,6 +1577,10 @@ def boxplot(
...
+        raise ValueError(msg)
```
mwaskom, line 1583: Small thing but `TypeError` is probably a better error class here, that is what you get if you passed an undefined parameter. Alternatively, `RuntimeError` since `positions` is technically valid but produces incorrect behavior.

tests/test_categorical.py
```diff
@@ -1168,7 +1168,25 @@ def test_vs_catplot(self, long_df, wide_df, null_df, flat_series, kwargs):
...
+            "y":[1, 1, 2, 2]
+        }
+            
+        try:
+            boxplot(**kwargs, positions=positions)
+            err = None
+        except ValueError as e:
+            err = e
+        
+        assert(err is not None)
```
mwaskom, line 1187: It would be slightly more idiomatic I think to use `pytest.raises` here, ideally with the `match` parameter to make sure we're getting the error we expect.

mwaskom (changes requested):
Thanks, I think raising an error here makes sense; it looks like `positions` errored on <=0.12 anyway since seaborn was already passing `positions` directly. (In 0.13, there was an effort to allow users to override more of the matplotlib params that seaborn sets defaults for, though as you discovered in the original issue, there's no guarantee that the overridden behavior will be correct).

leoluecken:
Hi, I came back to this now and updated my branch. How do I proceed? Should I open a new PR or can this be updated somehow? Sorry, I'm not very familiar with this process... :)

Issue #3563, opened 2023-11-17
Using weighted mean estimator for bootstrapped confidence intervals in seaborn plots


iainAtIon:
Hi, I would like to use a weighted mean estimator for calculating confidence intervals on various seaborn plots. In the past I have done this via a 'hack' suggested [here](https://github.com/mwaskom/seaborn/issues/722) which uses complex numbers to encode the data and its weights before passing to a seaborn plotting function.

Unfortunately, as of upgrading to seaborn v0.13.0, this approach no longer works as it seems like the complex numbers are cast to reals at some point in the plotting process (and hence lose part of the data). This had previously worked up until v0.12.2. 

I appreciate this was always a bit of a hack, but would either of the following be possible:
a) Add native support for weighted mean estimators to the seaborn plotting functions or,
b) Restore this hacky behaviour for now in a future release

I have tried alternatives such as storing the data and its weights in tuples or dataclasses, however neither of these approaches work as the data types are not numeric.

Language and package versions:
- Python v3.11.5
- numpy v1.26.2
- matplotlib v3.8.1
- pandas v2.1.3

Example code:
```
import pandas as pd
import seaborn as sns
import numpy as np

randomGenerator = np.random.default_rng(123)

values = randomGenerator.normal(10, 5, size=(100,))
weights = randomGenerator.uniform(size=(100,))

dataFrame = pd.DataFrame({'values': values, 'weights': weights})
dataFrame['valuesWithWeights'] = dataFrame['values'] + 1j * dataFrame['weights']

def WeightedMean(valuesWithWeights, **kwargs):
    values, weights = np.real(valuesWithWeights), np.imag(valuesWithWeights)
    weightedSum = np.sum((weights * values)) / np.sum(weights)
    return weightedSum

sns.barplot(data=dataFrame, y='valuesWithWeights', estimator=WeightedMean)
```

Output using seaborn v0.12.2

![image](https://github.com/mwaskom/seaborn/assets/143531681/1811057b-c024-4e56-bbff-3319896ff324)

Output using seaborn v0.13.0

```
[c:\Temp\seaborn_test\seaborn-venv\Lib\site-packages\matplotlib\cbook.py:1699](file:///C:/Temp/seaborn_test/seaborn-venv/Lib/site-packages/matplotlib/cbook.py:1699): ComplexWarning: Casting complex values to real discards the imaginary part
  return math.isfinite(val)
[...]

PR #3580, opened 2023-12-04
Add support for weights in objects.Est

mwaskom:
This PR enhances `seaborn.objects.Est` to support a weighted mean, with confidence intervals:

```python
(
    so.Plot(diamonds, x="color", y="price")
    .add(so.Dot(), so.Est(), weight="carat")
    .add(so.Range(), so.Est(), weight="carat")
)
```

(I don't know that this plot makes any particular sense, but it demonstrates how you'd use it...)

It is using `np.average` under the hood.

Currently support is limited to `func="mean", errorbar=("ci", ...)`.  There may be other estimators / error bar methods were it makes sense to compute a weighted version (e.g. weighted median/percentiles) although we would need to implement those in seaborn as I don't believe there's an implementation in our dependencies. I think this is fine and probably satisfies the majority of use cases; we can expand support later if there's enough interest.

(As a side note, we should probably error when `func != "mean"` and `errorbar=("se", ...)` since we don't generically compute the parameteric standard error of whatever estimator is passed).

See previous discussion in #3563, which is about the function interface. I actually think that it will be relatively straightforward to add first-class support in the function interface too on top of these changes. I might follow up here or maybe do that in another PR...

### DIFF
--- a/doc/_docstrings/objects.Est.ipynb
+++ b/doc/_docstrings/objects.Est.ipynb
@@ -110,9 +110,27 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "df807ef8-b5fb-4eac-b539-1bd4e797ddc2",
+   "metadata": {},
+   "source": [
+    "To compute a weighted estimate (and confidence interval), assign a `weight` variable in the layer where you use the stat:"
+   ]
+  },
+  {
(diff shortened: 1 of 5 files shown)


mwaskom:
Going to merge this and then follow up on the function interface in a separate PR.

PR #3600, opened 2023-12-22
FIX: Enable xticklabels for all bottom axes

### LINKED ISSUE #3548: seaborn.objects: facet doesn't show x tick labels for the bottom subplots in each column if the corresponding last row is empty
See the title. In the example below, I would expect the subplots H and I show the x tick labels, just like subplot J. Thank you. 

```python
import seaborn as sns
import seaborn.objects as so

p = so.Plot(diamonds, x="carat", y="price").add(so.Dots())
p.facet("color", wrap=3)
```
test

MaozGelbart:
Prior to this change, wrapped faceted plots (with shared axis) are created by deletion of some axes from a larger rectangular grid, so axes that are not on the bottom row lose their tick labels. Current code makes sure that all xtick labels are visible, but doesn't ensure that they're on. The proposed PR makes sure all "bottom" axes have their xticklabels on.  
Fixes #3548 :
Screenshot 2023-12-22 at 12 43 57

### DIFF
--- a/tests/_core/test_plot.py
+++ b/tests/_core/test_plot.py
@@ -1852,6 +1852,12 @@ class TestLabelVisibility:
         for s in subplots[1:]:
             ax = s["ax"]
             assert ax.xaxis.get_label().get_visible()
+            # mpl3.7 added a getter for tick params, but both yaxis and xaxis return
+            # the same entry of "labelleft" instead of  "labelbottom" for xaxis
+            if not _version_predates(mpl, "3.7"):
+                assert ax.xaxis.get_tick_params()["labelleft"]
+            else:
+                assert len(ax.get_xticklabels()) > 0
             assert all(t.get_visible() for t in ax.get_xticklabels())
 
         for s in subplots[1:-1]:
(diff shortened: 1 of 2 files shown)


tests/_core/test_plot.py
```diff
@@ -1852,6 +1852,12 @@ def test_1d_column_wrapped(self):
         for s in subplots[1:]:
             ax = s["ax"]
             assert ax.xaxis.get_label().get_visible()
+            # mpl3.7 added a getter for tick params, but both yaxis and xaxis return
+            # the same entry of "labelleft" instead of  "labelbottom" for xaxis
+            if not _version_predates(mpl, "3.7"):
+                assert ax.xaxis.get_tick_params()["labelleft"]
+            else:
+                assert len(ax.get_xticklabels()) > 0
```
mwaskom, line 1860: I'm confused about how this is flagged as not tested since the pinned build log appears to have matplotlib 3.4 installed.

MaozGelbart: Unsure about this comment by codecov, but its webpage for this test suite suggests it is tested (lines marked as green). See [here](https://app.codecov.io/gh/mwaskom/seaborn/pull/3600/blob/tests/_core/test_plot.py#L1860).

mwaskom: Agreed, the full code coverage report looks fine, not sure what's up with the bot.

mwaskom:
Thanks!

PR #3602, opened 2023-12-23
Add a missing raise in kdeplot and slightly improve lmplot signature

hamdanal:
I noticed these while working on the stubs at typeshed. They felt small enough to include in a single PR.

1. There is a missing `raise ` before `TypeError(msg)` in `kdeplot`
2. The `data` parameter of `lmplot` is required but it has a default of `None`. 20 lines later there is a check that raises a `TypeError` if `data is None`. Removing the default makes sense here to signal to the user that they must pass this argument. It also has two more advantages. The first is that your IDE will now helpfully yell at you if you don't pass the argument. The second is minor but it will allow us to keep running the stub tests for this function at typeshed where we removed the default which made the stub go out of sync with the runtime implementation so we had to skip its tests to pass CI. Note that there is no user facing change here as calling the function without passing `data` still raises a `TypeError` with a clear message.

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -1593,7 +1593,7 @@ def kdeplot(
     # Handle (past) deprecation of `data2`
     if "data2" in kwargs:
         msg = "`data2` has been removed (replaced by `y`); please update your code."
-        TypeError(msg)
+        raise TypeError(msg)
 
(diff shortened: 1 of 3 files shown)


mwaskom:
Thanks!

PR #3607, opened 2023-12-30
Cleanup compat code for obsolete versions of dependencies

hamdanal:
This PR removes compatibility code for features from matplotlib<1.4, numpy<1.20, and Python<3.8 as these versions are now obsolete according to the dependencies listed in pyrpoject.toml. There is one test change that was necessary, please see inline comment for explanation.

### DIFF
--- a/tests/_core/test_properties.py
+++ b/tests/_core/test_properties.py
@@ -3,6 +3,7 @@ import numpy as np
 import pandas as pd
 import matplotlib as mpl
 from matplotlib.colors import same_color, to_rgb, to_rgba
+from matplotlib.markers import MarkerStyle
 
 import pytest
 from numpy.testing import assert_array_equal
@@ -20,7 +21,7 @@ from seaborn._core.properties import (
     Marker,
     PointSize,
 )
-from seaborn._compat import MarkerStyle, get_colormap
+from seaborn._compat import get_colormap
 from seaborn.palettes import color_palette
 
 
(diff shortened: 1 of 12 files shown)


tests/_core/test_properties.py
```diff
@@ -358,6 +359,17 @@ class TestMarker(ObjectPropertyBase):
...
+        a_path, b_path = a.get_path(), b.get_path()
+        assert_array_equal(a_path.vertices, b_path.vertices)
+        assert_array_equal(a_path.codes, b_path.codes)
+        assert a_path.simplify_threshold == b_path.simplify_threshold
+        assert a_path.should_simplify == b_path.should_simplify
+
+        assert a.get_joinstyle() == b.get_joinstyle()
+        assert a.get_transform().to_values() == b.get_transform().to_values()
+        assert a.get_fillstyle() == b.get_fillstyle()
+
```
hamdanal, line 372: This test change was needed because here we need to compare `MarkerStyle` objects. This class does not define `__eq__` method so it is hard to compare its instances against each other. This was not needed before because the old `MarkerStyle` compatibility function returned the same matplotlib `MarkerStyle` object without creating a copy. This allowed the test to pass earlier becuase `marker1 == marker2` falls back to `marker1 is marker2` which was true when the object returned was not a copy but is no longer the case. Instead we now compare the attributes of the marker objects.

mwaskom:
Nice — thanks!

PR #3611, opened 2024-01-06
microscopic typofix: "swaarms"

aalok-sathe:
appears here: https://seaborn.pydata.org/generated/seaborn.swarmplot.html


mwaskom:
Thanks! would also have accepted swårms :)

PR #2727, opened 2022-01-18
[DO NOT MERGE] Test pandas 1.4 release candidate

jorisvandenbossche:
Similarly as https://github.com/mwaskom/seaborn/pull/2608, not to merge, but just to test the pandas release candidate on your CI.

PR #3615, opened 2024-01-12
[DO NOT MERGE] Test pandas 2.2 release candidate

jorisvandenbossche:
Similarly as https://github.com/mwaskom/seaborn/pull/2727, not to merge, but just to test the pandas release candidate on your CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,6 +57,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         install: [full]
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks! Looks like one test failure based on a construct used in a test, easy for seaborn to adapt (although the error message is weird as the capitalization in the error message doesn't match what is in the code — maybe double check that's not a pandas bug?)

But also a tremendous amount of `FutureWarning` for upcoming deprecations that [did not go through a `DeprecationWarning` schedule](https://github.com/pandas-dev/pandas/issues/54970) :/

jorisvandenbossche:
> Looks like one test failure based on a construct used in a test, easy for seaborn to adapt (although the error message is weird as the capitalization in the error message doesn't match what is in the code — maybe double check that's not a pandas bug?)

Yes, that's indeed easy to fix here, and have also already reported that on the pandas side: https://github.com/pandas-dev/pandas/issues/56847#issuecomment-1889375264. It was not really documented that those lower case versions also work, but it's probably a bug in the deprecation that those stopped working.

mwaskom:
Thanks for flagging!

PR #3620, opened 2024-01-21
Adapt to deprecations in pandas 2.2.0

mwaskom:
😠

### DIFF
--- a/tests/test_categorical.py
+++ b/tests/test_categorical.py
@@ -2078,7 +2078,7 @@ class TestBarPlot(SharedAggTests):
 
     def test_datetime_native_scale_axis(self):
 
-        x = pd.date_range("2010-01-01", periods=20, freq="m")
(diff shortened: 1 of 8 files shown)


tests/test_categorical.py
```diff
@@ -2078,7 +2078,7 @@ def test_xy_native_scale_log_transform(self):
 
     def test_datetime_native_scale_axis(self):
 
-        x = pd.date_range("2010-01-01", periods=20, freq="m")
+        x = pd.date_range("2010-01-01", periods=20, freq="MS")
```
jorisvandenbossche, line 2081: FWIW if you want to have the same values as before you need "ME" (month end, not month start). Not sure whether that matters for the test, though (you test below with "2003-02-28", which is a "month end")

mwaskom: Ah yeah thanks, the test date here being a "month end" is just a coincidence and not critical to the test logic.

Issue #2351, opened 2020-11-13
Pass parameters to `statsmodels.lowess` in a call to `regplot`


RuolinZheng08:
This feature might have been requested before: May we customize the `lowess` parameter in `regplot` and pass arguments to `statsmodels` function `nonparametric.smoothers_lowess.lowess(endog, exog, frac=0.6666666666666666, it=3, delta=0.0, xvals=None, is_sorted=False, missing='drop', return_sorted=True)`?

I saw the mention of a `seaborn.lowessplot` function mentioned in #607. Is there any plan to implement that?

Thanks for your hard work on this beautiful visualization library.

mwaskom:
Yes, I would like to expose an interface for LOWESS-based regression plots that allows more control over the smoothing parameters. I'm also interested in other enhancements, like using statsmodel's new GAM tools for faster local smoothing on large datasets (similar to what ggplot does) and getting confidence intervals on the prediction. I continue to think the right way forward is to separate out the nonparametric regressionss into their own function. This is tentatively on the roadmap for a 0.12 release. It's not something that should happen before the regression module can be refactored to use the new core data infrastructure introduced in 0.11.

leoluecken:
I also needed this and did a first draft for the implementation. See PR to follow.

PR #3626, opened 2024-01-29
Exposing lowess params in regplot, refs #2351

leoluecken:
This restricts user access to the parameters `frac`, `it`, and `delta`, since I wasn't sure how much of the other parameters would be assumed to hold on the seaborn side.

Also note, that the test `test_residplot_lowess_with_kws` still fails. This is because the original test `test_residplot_lowess` does as well.


leoluecken:
Please note that I didn't test the documentation!

PR #3635, opened 2024-02-18
Test on nightly numpy

mwaskom:
Don't merge


rgommers:
Hi @mwaskom, since this is green that looks like a good sign. Can I ask if you know if recently released `seaborn` versions are compatible with NumPy 2.0?

mwaskom:
Ah, no, I put this up a while ago before realizing that there wasn’t actually a 2.0rc out yet. I don’t thing I ever succeeded in installing numpy here. Sorry for the confusion.

rgommers:
Ah okay, thanks. There is an RC now. `pip install numpy --pre` should do it.

mwaskom:
Yep I’ve been tracking your issue, just have no bandwidth for seaborn right now 👍

PR #3683, opened 2024-04-25
Test on numpy 2.0 preview

mwaskom:


mwaskom:
cc @rgommers, looks like seaborn is numpy 2.0 compatible after removing one reference to a deprecated name inside of a test (#3685)

rgommers:
Great, thanks @mwaskom. I'll list the latest release on PyPI (`0.13.2`) as already compatible then, since it looks like nothing relevant was merged since that release.

PR #3684, opened 2024-04-25
Relax typing strictness for Plot.theme

### LINKED ISSUE #3682: Relax typing strictness for Plot.theme
Plot.theme currently accepts a `dict[str, Any]`, which unfortunately throws a pytype typing error when passed another similar mapping, like an immutable dict.

```python
def accepts_dict(d: dict[str, Any]) -> None:
    print(d)

accepts_dict(immutabledict.immutabledict({'a': 1}))
# Error: Function accepts_dict was called with the wrong arguments [wrong-arg-types]
#        Expected: (d: Dict[str, Any], ...)
#        Actually passed: (d: immutabledict.immutabledict[nothing, nothing])
```

Using collections.abc.Mapping doesn't throw an error.

```python
from collections.abc import Mapping

def accepts_mapping(d: Mappingf[str, Any]) -> None:
    print(d)

accepts_mapping(immutabledict.immutabledict({'a': 1}))
```

Would you mind relaxing the type specificity for this? Thanks!

michaelquinn32:
Long-term it might help to look at more of the specific types in annotations and adjust them, but I'd prefer to remain incremental for now.

Fixes #3682

### DIFF
--- a/seaborn/_core/plot.py
+++ b/seaborn/_core/plot.py
@@ -9,7 +9,7 @@ import itertools
 import textwrap
 from contextlib import contextmanager
 from collections import abc
-from collections.abc import Callable, Generator
+from collections.abc import Callable, Generator, Mapping
 from typing import Any, List, Literal, Optional, cast
 from xml.etree import ElementTree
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Thanks!

Issue #3690, opened 2024-05-10
Relplot refline error in situations when using concatenated dataframes


zacharygibbs:
This error occurs when using relplot then adding a refline. From investigation, it appears that replot is duplicating the data. In this example, when i concatentate my 3 dataframes, I did not use 'ignore_index'; therefore there are duplicate indicies in the input data. 

The problem is solved when I use ignore_index, or feed the data in with as `df.reset_index()`, however, the error message was not useful in discovering this! After tracking down the relplot source code, it appears the problem is related to the `grid_data` merging at the end of the function. I was able to solve this by skipping the "merge" if all of the columns are already present. I will plant to submit a pull request.

the error was: `ValueError: operands could not be broadcast together with shapes (45000,) (15000,) `
(the input data was 15000 rows long with 3 different "hue" variables)

Reproducible example:

```python
import pandas as pd
import numpy as np

import seaborn as sns
import matplotlib.pyplot as plt

n_items = 5000
n_floats = 5
n_categorical = 3

df1 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df1 = df1.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df2 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df2 = df2.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df3 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df3 = df3.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df = pd.concat([df1.assign(origin=1), df2.assign(origin=2), df3.assign(origin=3)])
print(df)

fg=sns.relplot(data=df, x='float1', y='float2', hue='origin', row='categorical1')
print('main', fg.data.shape)
fg.refline(y=0.5)



plt.show()
```

zacharygibbs:
@mwaskom do you not take contributions from outside? I saw you closed my PR.

PR #3692, opened 2024-05-10
BUGFIX: Relplot Error adding refline when duplicate indicies present

zacharygibbs:
This is related to issue #3690 

In this case, replot was creating duplicated data when it didn't need to when the input data had duplicate indicies, which caused the refline addition to fail.

I was able to solve this by modifying the `grid_data` merge step at the end of the relplot function to only merge when there's actually something to merge!

# Print Debugging before fix (self.data - dataframe shape):
```
relplot (15000, 4)
relplot_before_grid_data (15000, 4)
relplot_after_grid_data (45000, 9)
main (45000, 9)
ValueError: operands could not be broadcast together with shapes (45000,) (15000,) 
```

# Print Debugging After (self.data - dataframe shape)

```
relplot (15000, 4)
relplot_before_grid_data (15000, 4)
relplot_after_grid_data (15000, 4)
main (15000, 4)
```

# reproducible example

(From original issue, for reference - reproducible example)
```python
import pandas as pd
import numpy as np

import seaborn as sns

n_items = 5000
n_floats = 5
n_categorical = 3

df1 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df1 = df1.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df2 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df2 = df2.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df3 = pd.DataFrame(
    np.random.random((n_items, n_floats)),
    columns=[f'float{i}' for i in range(n_floats)]
)
df3 = df3.assign( **{f'categorical{i}': np.random.randint(0, 15, n_items) for i in range(n_categorical)})

df = pd.concat([df1.assign(origin=1), df2.assign(origin=2), df3.assign(origin=3)])
print(df)

fg=sns.relplot(data=df, x='float1', y='float2', hue='origin', row='categorical1')
print('main', fg.data.shape)
fg.refline(y=0.5)



plt.show()
```

### DIFF
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -833,8 +833,6 @@ def relplot(
         height=height, aspect=aspect, dropna=False,
         **facet_kws
     )
-    
-    print('relplot', g.data.shape)
 
     # Draw the plot
     g.map_dataframe(func, **plot_kws)
--- a/seaborn/relational.py
+++ b/seaborn/relational.py
@@ -833,6 +833,8 @@ def relplot(
         height=height, aspect=aspect, dropna=False,
         **facet_kws
     )
+    
+    print('relplot', g.data.shape)
 
     # Draw the plot
     g.map_dataframe(func, **plot_kws)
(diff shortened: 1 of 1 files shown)


zacharygibbs:
Any word on this pull request? Seems like a simple fix, what's the hold up?

Issue #3646, opened 2024-03-01
Value error in histplot with binwidth larger than the double of data range


jhncls:
`sns.histplot([1, 2, 3], binwidth=7)` crashes with

"ValueError: `bins` must be positive, when an integer"

It is related to #2721, which is marked as solved, but the error also happens in the dev version.

The cause seems to be [line 136 in counting.py](https://github.com/mwaskom/seaborn/blob/9bdfc7484c15894dd35d58d3f16475389dfe4fe5/seaborn/_stats/counting.py#L136) with `bins = int(round((stop - start) / binwidth))`, setting it to `0` for small ranges.  Changing this to `bins = max(1, int(round((stop - start) / binwidth)))` would probably solve it adequately.

(The code also has problems when `binwidth=0` (division by zero) or negative (this makes `bins` negative, which causes numpy to protest)).

By the way, `binwidth` is ignored when `discrete=True`. As an example, `sns.histplot(titanic[titanic['who'] == 'woman'], x='age', binwidth=5)` has some bins with 5 and others with 6 ages. Of course, people who really care can provide their own custom bin edges.

mwaskom:
Should probably just reject unless `0 < binwidth <= binrange`. Not sure any other parametrization makes sense.

> By the way, binwidth is ignored when discrete=True. 

Yeah, the parameter is [documented as](https://seaborn.pydata.org/generated/seaborn.histplot.html) 

> If True, default to `binwidth=1` ...

Maybe "default to" implies that you can override it and should be "set `binwidth=1`

jhncls:
> Should probably just reject unless 0 < binwidth <= binrange. Not sure any other parametrization makes sense.

When the data isn't known in advance, in some situations only 0 or 1 data points will pop up. Without setting binwidth, now an empty plot is returned for 0 data points, and a symmetric bin of 1 wide for only one data point.  With binwidth set, the code could behave similarly, using that binwidth. A plot showing something a bit reasonable could be friendlier than an error message.

> If True, default to `binwidth=1` ...

Well, my comment was a suggestion to also allow other integer binwidths for discrete data, shifting the edges by a half. (But maybe this would complicate things too much. Next, people will be asking for discrete units measured fractions, e.g. 1/10ths.)

Erotemic:
I got this error when all the data in the column of interest was the same, and was pretty confused as to the problem. After figuring it out, I made a MWE:

```python
import seaborn as sns
import pandas as pd
sns.set()

[...]

PR #3724, opened 2024-07-05
fix: bin width < half spread  #3646

Gornoka:
prevents edge-case when a bin only has elements with spread < 0.5*bin_width .

I would personally prefer this option, with ceil instead of min(1,..).
This would alter previous behavior for binning in general( creating an additional bin in ~50% of cases),I would be interested to know if that is acceptable here or if you would like to use the min approach instead.

### DIFF
--- a/examples/histogram_binning_options.py
+++ b/examples/histogram_binning_options.py
@@ -1,0 +1,50 @@
+"""
+Histogram binning options on data that does not fit the bin size.
+=================================================================
+
+"""
+import pandas as pd
+
+import seaborn as sns
+import matplotlib.pyplot as plt
+
+if __name__ == '__main__':
+    df_round_up = pd.DataFrame({"x": list(range(58))})
+    df_round_down = pd.DataFrame({"x": list(range(52))})
+    n_plots = 4
+
+    fig, ax = plt.subplots(n_plots, 2, figsize=(5 * 2, 5 * n_plots))
+    fig.suptitle("Histogram binning options\nrange(58) and range(52)")
+    fig.tight_layout()
+
+    hist_n_bins_ru = sns.histplot(df_round_up,
+                                  x="x", bins=10, ax=ax[0][0])
+    hist_n_bins_ru.set_title("bins=10")
+    hist_n_bins_rd = sns.histplot(df_round_down,
+                                  x="x", bins=10, ax=ax[0][1])
+    hist_n_bins_rd.set_title("bins=10")
+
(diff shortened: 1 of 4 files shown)


mwaskom:
What is the edge case? Is there an issue that you can link to? Is there a failing test that you could add?

Gornoka:
> What is the edge case? Is there an issue that you can link to? Is there a failing test that you could add?

this is related to #3646

mwaskom:
> I would personally prefer this option, with ceil instead of min(1,..).

Do you mean `max(1, ...)`?

That sounds like a better approach to me — IIUC the linked issue, the edge case is entirely about cases where the default bin rule gives "less than 1 bin" and that doesn't make sense. Why would we want to change other plots?

Gornoka:
> > I would personally prefer this option, with ceil instead of min(1,..).
> 
> Do you mean `max(1, ...)`?
> 
> That sounds like a better approach to me — IIUC the linked issue, the edge case is entirely about cases where the default bin rule gives "less than 1 bin" and that doesn't make sense. Why would we want to change other plots?

Mainly I would want to fix the issue with no bins, personally i would prefer the behaviour with ceil plots instead round plots. But I understand that some people may expect the old behavior, that is why I asked for opinions on that.

I will update this PR to use the max approach, the min was just a mixup, sorry for the confusion.

mwaskom:
> personally i would prefer the behaviour with ceil plots instead round plots
why though? do you have a reason?

Gornoka:
> > personally i would prefer the behaviour with ceil plots instead round plots
> 
> why though? do you have a reason?

When I specify `binwidth` my assumption was that I would get a histogram with that bin width.
When I specify `bins` I would get a histogram with that number of bins.
When I specify the `binrange` I would get whatever is in that range depending on the other settings.

Example graphs of the current behavior below.
bin_sizes

I was expecting `bindwidth` to create new bins until it fits my whole data range .
If I have data that does not fit perfectly in the bin width, I should get the last bin semi-empty, 
resulting in a round up of n_bins. (like it does in the bottom row)

If the range is not specified explicitly seaborn will shrink or extend the bins to fit the bins to evenly over the data.
That is a good decision to make the histogram easier to read correctly, and when using round the shrunk/extended bin width is closer to what the user asked for. So probably my preference is wrong here for many applications and I will have to restort to setting both `binrange` and `binwidth` in the future.

Gornoka:
are there any next steps for this pr, / any requested changes or do you want to kill it  @mwaskom ?

Issue #3728, opened 2024-07-12
Incorrect plotting of exactly overlapping scatter with `hue` and `hue_order`


eloyvallinaes:
While working with `sns.scatterplot` for representing locations on a grid, I discovered an issue where using `hue` and `hue_order` produces an incorrect plot: markers that should be perfectly overlapping—they have identical (`x`, `y`) coordinates—are drawn at a small offset, such that the edge of one can be seen intersecting the other. Here's a minimal example that reproduces the issue with `matplotlib 3.9.1` and `seaborn 0.13.2`:
```python
import matplotlib
import matplotlib.pyplot as plt
import seaborn as sns
import pandas as pd

df = pd.DataFrame.from_dict({
    'x': [6.3, 6.3, 6.3, 6.3, 6.633333, 6.633333, 6.633333, 6.633333, 33.48, 33.48, 33.48, 33.48, 33.813333, 33.813333, 33.813333, 33.813333],
    'y': [-12.42, -12.42, -4.0, -4.0, -12.42, -12.42, -4.0, -4.0, -12.42, -12.42, -4.0, -4.0, -12.42, -12.42, -4.0, -4.0],
    'locid': ['loc1', 'loc1', 'loc1', 'loc1', 'loc2', 'loc2', 'loc2', 'loc2', 'loc1', 'loc1', 'loc1', 'loc1', 'loc2', 'loc2', 'loc2', 'loc2']
})

sns.scatterplot(
    data=df,
    x='x',
    y='y',
    marker="o",
    hue='locid',
    hue_order=['loc1'],
)
print('Pandas version: ', pd.__version__)  # 2.2.2
print('Matplotlib version: ', matplotlib.__version__)  # 3.9.1
print('Seaborn version: ', sns.__version__)  # 0.13.2
```
That code produces the following plot:
![bugPlot](https://github.com/user-attachments/assets/9f3b7511-f90c-48ca-b5e0-4bde179dc146)
where at each corner, the edge of the second marker is clearly seen to intersect the face of the first

From my brief dive into this problem:
1. As in the example, it doesn't matter whether a tall stack of markers are made to overlap: there's only to points with the exact (6.3, -12.42) coordinates and the problem is there.
2. The issue is seaborn-specific. Using matplotlib's `plt.scatter` does yield a correct plot.
3. Both `hue` and `hue_order` need to be used in order for the issue to appear. Slicing the data with `df[df.locid == 'loc1']` makes a correct plot.
[...]

Issue #3601, opened 2023-12-23
scatterplot crashes or produces unexpected results when {`hue`,`style`}`_order` do not contain all `hue`/`style` values


mmore500:
**bug description:** scatterplot plots hollow points when `hue_order` is a strict subset of `hue` values within dataframe and crashes when `style_order` is a strict subset of `style` values with in dataframe (i.e., the dataframe contains `hue` or `style` values not present in `hue_order`/`style_order`). Was able to reproduce in versions 0.13.0, 0.12.0, and 0.11.0

**expected behavior:** scatterplot would should plot the subset of data with values specified in `hue_order` and `style_order`, like current behavior of `lineplot`, `kdeplot`, etc.

**related issues:** none obvious, #3575 has different stack trace and does not occur with 0.12.x versions of seaborn

if wanted, I'd be happy to look into contributing a fix

## `scatterplot`: plots hollow points when `set(hue_order) < set(df[hue])`

> ```python3
> sns.scatterplot(
>     data=sns.load_dataset("diamonds").dropna(),
>     x="price",
>     y="carat",
>     hue="clarity",
>     hue_order=["I1", "IF"],
> )
> plt.gca().set_facecolor("gray")
> ```
> ![image](https://github.com/mwaskom/seaborn/assets/10763333/139cb60f-4432-4955-9954-055b3e2ff7b6)

## `scatterplot`: crashes when `set(style_order) < set(df[style])`

> ```python3
> sns.scatterplot(
>     data=sns.load_dataset("diamonds").dropna(),
>     x="price",
>     y="carat",
>     style="clarity",
>     style_order=["I1", "IF"],
> )
> ```
>
> ```
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/home/mmore500/2023-12-09/outset/env210/lib64/python3.10/site-packages/seaborn/relational.py", line 624, in scatterplot
>     p.plot(ax, kwargs)
>   File "/home/mmore500/2023-12-09/outset/env210/lib64/python3.10/site-packages/seaborn/relational.py", line 443, in plot
>     p = [self._style_map(val, "path") for val in data["style"]]
>   File "/home/mmore500/2023-12-09/outset/env210/lib64/python3.10/site-packages/seaborn/relational.py", line 443, in <listcomp>
>     p = [self._style_map(val, "path") for val in data["style"]]
[...]

PR #3730, opened 2024-07-15
Nullify edgecolor if facecolor is null, preserve input color otherwise

eloyvallinaes:
In connection with #3728 and #3601 added logic to `relational._ScatterPlotter.plot` such that:

- Marker with null `facecolor` also have null `edgecolor`.
- Value supplied through parameter `edgecolor`, including default white, is retained for all other markers.


mwaskom:
Hm not sure this is the right fix; users may set a null facecolor to show line-art markers on purpose.

PR #3733, opened 2024-07-22
Add `inner="median"` option to violin plot

fdamken:
Sometime, one does not want to plot the 25% and 75% quantiles but only the median, especially when previously outliers were removed from the data.

I added a new option for the `inner` keyword of `violinplot` to accept `"median"` which in turn only plots the median.


mwaskom:
Thanks, but I think I am :-1: on additional options for violinplots — I also don't understand why outlier removal would mean that you don't want to see the full quartiles.

EwoutH:
@fdamken could you maybe provide a few examples in which this option is useful (preferably comparing to the current plots without `inner="median"`)?

Issue #3664, opened 2024-03-25
jointplot with kind="hex" fails with datetime64[ns]


fynnkroeger:
Minimal example:
```python
import seaborn as sns
import numpy as np
dates = np.array(["2023-01-01", "2023-01-02", "2023-01-03"], dtype="datetime64[ns]")
sns.jointplot(x=dates, y=[1, 2, 3], kind="hex")
```
Error:
```
Traceback (most recent call last):
  File "/.../seaborn_bug.py", line 21, in <module>
    sns.jointplot(x=dates, y=[1, 2, 3], kind="hex")
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/seaborn/axisgrid.py", line 2307, in jointplot
    x_bins = min(_freedman_diaconis_bins(grid.x), 50)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/seaborn/distributions.py", line 2381, in _freedman_diaconis_bins
    iqr = np.subtract.reduce(np.nanpercentile(a, [75, 25]))
TypeError: the resolved dtypes are not compatible with subtract.reduce. Resolved (dtype('<M8[ns]'), dtype('<M8[ns]'), dtype('<m8[ns]'))
```
I think this should work, as datetime64[ns] is the default type for datetimes in pandas. It works when I omit `kind="hex"` or use `kind="kde"`. The error was in version 0.13.2. In 0.11.2 I got a error in the same cases but it was a integer overflow with numpy during conversions.

athompson1991:
Hi, I'm not especially experienced in open source development, I think would count as my first contribution, but I just opened a pull request on this. The initial bug was easy enough to sort out, but there were some downstream complications I describe in the pull request. Let me know if this is an appropriate fix, glad to help as best I can.

PR #3742, opened 2024-08-04
Test for datetime input on jointplot, code to pass test (#3664)

athompson1991:
The initial error comes from the `_freedman_diaconis_bins` function, which I fixed with a straightforward check on data type. However, once the code got through that hurdle, it runs into a more complicated blocker with an overflow error derived from matplotlib's handling of dates as integer. I get around the overflow issue with some manipulation of the object state, but some of the fix is admittedly imperfect. Nonetheless, this code alleviates the original problem seen with `datetime64[ns]` handling, see plot below for example code and plot:

```python
import numpy as np
from seaborn import axisgrid as ag

dates = np.arange("2024-01-01", "2024-04-01", dtype="datetime64[D]").astype("datetime64[ns]")
random_dates = np.random.choice(dates, size=1000)
y = np.random.normal(size=1000)
plot = ag.jointplot(x=random_dates, y=y, kind="hex")

plot.fig.savefig("out.png")
```

out

### DIFF
--- a/seaborn/axisgrid.py
+++ b/seaborn/axisgrid.py
@@ -1704,6 +1704,7 @@ class JointGrid(_BaseGrid):
         self.ax_joint = ax_joint
         self.ax_marg_x = ax_marg_x
         self.ax_marg_y = ax_marg_y
+        self.is_datetime_x = False
 
         # Turn off tick visibility for the measure axis on the marginal plots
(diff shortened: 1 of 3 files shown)


mwaskom:
This isn't the way we handle datetime data in seaborn; should be using the matplotlib units system. Also I don't understand why it would only be implemented for the x axis.

PR #3762, opened 2024-10-03
Exclude tests/* from coverage

gliptak:
https://app.codecov.io/gh/mwaskom/seaborn


mwaskom:
No thanks, this is intentional.

PR #3782, opened 2024-11-09
fix: use PyCapsule Interface instead of Dataframe Interchange Protocol

### LINKED ISSUE #3756: Use Arrow PyCapsule Interface instead of Dataframe Interchange Protocol
This is something I've chatted with @MarcoGorelli offline about. At the time it was implemented in seaborn, the Dataframe Interchange Protocol was the best option for exchanging dataframe-like data. However, since that was implemented in seaborn, the [PyArrow Capsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) has come along and solved many of the issues that the DataFrame Interchange Protocol left open.

Without knowing the current state of the interchange implementation of seaborn, switching to the PyArrow Capsule Interface should solve at least the following issues:

- It will add support for polars and other dataframe libraries (https://github.com/mwaskom/seaborn/issues/3277 and https://github.com/mwaskom/seaborn/issues/3188)
- It will use the Arrow type system, which supports aggregate types (https://github.com/mwaskom/seaborn/issues/3533)
- The wonkiness of pandas' type system won't be inherited by seaborn (potentially solving https://github.com/mwaskom/seaborn/issues/3519)

The interface has been adopted by a good deal of projects already, some of which are being tracked in https://github.com/apache/arrow/issues/39195

MarcoGorelli:
closes #3756 
closes https://github.com/mwaskom/seaborn/issues/3533
I'm hoping that this can supersede https://github.com/mwaskom/seaborn/pull/3534

This means that you get support for quite a lot more, e.g.:

- **DuckDB**:

image

- **cuDF** (their interchange protocol implementation is currently broken anyway https://github.com/rapidsai/cudf/issues/17282)

- **Polars**: it fixes the issue reported in #3533, because the PyCapsule interface actually supports nested data types:

image

In addition, this has **no effect on existing pandas users**, as there's already an early return for pandas https://github.com/MarcoGorelli/seaborn/blob/0bd85071284d45f38cbf419b8cf1efb2179eda24/seaborn/_core/data.py#L284-L285

---

I'm sorry for having introduced the Interchange Protocol in the first place. It's turned out to be fairly problematic, see https://github.com/pandas-dev/pandas/issues/56732#issuecomment-2466301769 as the associated discussion for more context

---

cc @willayd for comments

### DIFF
--- a/seaborn/_core/data.py
+++ b/seaborn/_core/data.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 from collections.abc import Mapping, Sized
 from typing import cast
-import warnings
 
 import pandas as pd
 from pandas import DataFrame
@@ -269,9 +268,9 @@ class PlotData:
 
 def handle_data_source(data: object) -> pd.DataFrame | Mapping | None:
     """Convert the data source object to a common union representation."""
-    if isinstance(data, pd.DataFrame) or hasattr(data, "__dataframe__"):
+    if isinstance(data, pd.DataFrame) or hasattr(data, "__arrow_c_stream__"):
         # Check for pd.DataFrame inheritance could be removed once
-        # minimal pandas version supports dataframe interchange (1.5.0).
+        # minimal pandas version supports PyCapsule Interface (2.2).
         data = convert_dataframe_to_pandas(data)
     elif data is not None and not isinstance(data, Mapping):
         err = f"Data source must be a DataFrame or Mapping, not {type(data)!r}."
@@ -285,35 +284,32 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
     if isinstance(data, pd.DataFrame):
         return data
 
-    if not hasattr(pd.api, "interchange"):
-        msg = (
-            "Support for non-pandas DataFrame objects requires a version of pandas "
-            "that implements the DataFrame interchange protocol. Please upgrade "
-            "your pandas version or coerce your data to pandas before passing "
-            "it to seaborn."
-        )
-        raise TypeError(msg)
+    if hasattr(data, '__arrow_c_stream__'):
+        try:
+            import pyarrow
+        except ImportError as err:
+            msg = "PyArrow is required for non-pandas Dataframe support."
+            raise RuntimeError(msg) from err
+        if _version_predates(pyarrow, '14.0.0'):
+            msg = "PyArrow>=14.0.0 is required for non-pandas Dataframe support."
+            raise RuntimeError(msg)
+        try:
+            # This is going to convert all columns in the input dataframe, even though
+            # we may only need one or two of them. It would be more efficient to select
+            # the columns that are going to be used in the plot prior to interchange.
+            # Solving that in general is a hard problem, especially with the objects
+            # interface where variables passed in Plot() may only be referenced later
+            # in Plot.add(). But noting here in case this seems to be a bottleneck.
+            return pyarrow.table(data).to_pandas()
+        except Exception as err:
+            msg = (
+                "Encountered an exception when converting data source "
+                "to a pandas DataFrame. See traceback above for details."
+            )
+            raise RuntimeError(msg) from err
 
-    if _version_predates(pd, "2.0.2"):
-        msg = (
-            "DataFrame interchange with pandas<2.0.2 has some known issues. "
-            f"You are using pandas {pd.__version__}. "
-            "Continuing, but it is recommended to carefully inspect the results and to "
-            "consider upgrading."
-        )
-        warnings.warn(msg, stacklevel=2)
-
-    try:
-        # This is going to convert all columns in the input dataframe, even though
-        # we may only need one or two of them. It would be more efficient to select
-        # the columns that are going to be used in the plot prior to interchange.
-        # Solving that in general is a hard problem, especially with the objects
-        # interface where variables passed in Plot() may only be referenced later
-        # in Plot.add(). But noting here in case this seems to be a bottleneck.
-        return pd.api.interchange.from_dataframe(data)
-    except Exception as err:
-        msg = (
-            "Encountered an exception when converting data source "
-            "to a pandas DataFrame. See traceback above for details."
-        )
-        raise RuntimeError(msg) from err
+    msg = (
+        "Expected object which implements '__arrow_c_stream__' from the "
+        f"PyCapsule Interface, got: {type(data)}"
+    )
+    raise TypeError(msg)
--- a/tests/_core/test_data.py
+++ b/tests/_core/test_data.py
@@ -1,6 +1,7 @@
 import functools
 import numpy as np
 import pandas as pd
+from seaborn.external.version import Version
 
 import pytest
 from numpy.testing import assert_array_equal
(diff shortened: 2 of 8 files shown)


WillAyd:
Implementation wise I think this looks great. Nice work @MarcoGorelli

tests/_core/test_data.py
```diff
@@ -405,8 +405,8 @@ def test_bad_type(self, flat_list):
             PlotData(flat_list, {})
 
     @pytest.mark.skipif(
-        condition=not hasattr(pd.api, "interchange"),
-        reason="Tests behavior assuming support for dataframe interchange"
+        condition=not hasattr(pd.DataFrame, "__arrow_c_stream__"),
```
WillAyd, line 408: Not sure if its a better way of expressing things but this was added in pandas 2.2

MarcoGorelli: thanks - i think we should just conditionally skip based on the pyarrow version, seeing as that's what's doing the interchanging now, that's what i've added now

WillAyd: Do we need to bump the minimum version of pandas anywhere here though? Not familiar with the CI setup, but I would imagine that this PR  will not work with pandas < 2.2

MarcoGorelli: for pandas there's already an early return, so there's no issue for pandas users

https://github.com/MarcoGorelli/seaborn/blob/0bd85071284d45f38cbf419b8cf1efb2179eda24/seaborn/_core/data.py#L284-L285

seaborn/_core/data.py
```diff
@@ -285,35 +284,32 @@ def convert_dataframe_to_pandas(data: object) -> pd.DataFrame:
...
-        msg = (
-            "Encountered an exception when converting data source "
-            "to a pandas DataFrame. See traceback above for details."
-        )
-        raise RuntimeError(msg) from err
+    if hasattr(data, '__arrow_c_stream__'):
+        try:
+            import pyarrow
+        except ImportError as err:
+            msg = "PyArrow is required for non-pandas Dataframe support."
```
mwaskom, line 291: Is this generally a dependency of non-pandas dataframe libraries now? Or could this change introduce a regression for e.g. polars users who are currently leveraging the dataframe interchange protocol?

tests/_core/test_data.py
```diff
@@ -404,11 +405,11 @@ def test_bad_type(self, flat_list):
         with pytest.raises(TypeError, match=err):
             PlotData(flat_list, {})
 
-    @pytest.mark.skipif(
-        condition=not hasattr(pd.api, "interchange"),
-        reason="Tests behavior assuming support for dataframe interchange"
-    )
     def test_data_interchange(self, mock_long_df, long_df):
+        pytest.importorskip(
```
mwaskom, line 409: Nice, TIL

mwaskom (commented):
Cool, thanks for this. I think my one question is about how compatible this will be for users that are currently benefitting from the (seemingly more-or-less built-in) interchange protocol. Do we need to provide backwards compatibility for them?

seaborn/_core/data.py
MarcoGorelli: Thanks for your review!

Polars doesn't depend on PyArrow, but `polars.DataFrame.to_pandas` always requires PyArrow. So, in practice, anyone working with both dataframe libraries may well already have PyArrow already installed

To avoid requiring PyArrow for the cases when it's not necessary, one way could be to do something like:
- try using the interchange protocol
- if it raises, then fall back to the PyCapsule Interface (which currently requires PyArrow)

This has the upside of not requiring PyArrow in some cases, but the downside of hiding issues where the interchange protocol silently produces invalid results

It may be possible to do this PyCapsule Interface conversion in the future without PyArrow but with something lighter instead, like arro3 by @kylebarron (who I'm ccing in case he has comments too)

What would be your preference?

kylebarron: Some polars users may not have pyarrow installed. If seaborn needs to get pandas data, the only production-ready way to do `Arrow -> pandas` that I know of is using pyarrow. 

As Marco mentions I'm working on [arro3](https://github.com/kylebarron/arro3), which is a minimal library for Arrow in Python, but Pandas interop is not a primary concern, and it's not production-ready today.

WillAyd: FWIW pandas 3.x is going to strongly incentivize users to install PyArrow, although it stops short of outright requiring it. In theory, the only people that shouldn't have PyArrow installed are those that operate in space/resource constrained environments, probably in headless environments like AWS Lambda where seaborn won't be used

Of course up to you how much you want to support non-PyArrow configurations, but the dataframe interchange protocol is relatively buggy and gets very little support, so you may find it easier altogether to force users towards PyArrow

MarcoGorelli: cuDF have said that they will deprecate the interchange format: https://github.com/rapidsai/cudf/issues/17282

Plotly have stopped using it, so Seaborn is the only library left using it

At this point, I think there's a greater risk in keeping it - I don't want to force anything here of course, just making sure you're aware

mwaskom: Sure, happy to drop it. I guess just a question of whether it's better to adopt this new thing or just remove the attempt at cross-dataframe compatibility altogether and leave that up to users.

MarcoGorelli: As a user, I find it convenient to be able to pass a dataframe to Seaborn and have it just work, but I can see merits to just telling users to do it themselves too

mwaskom: I mean yeah I was on board with the promise of it "just working" but if the infra isn't there yet or there are lots of edge cases it's easier _for me_ to push dealing with them out to the user :)

WillAyd: Would you be opposed to taking on the PyArrow dependency? I think if you use that, your implementation becomes much simpler.

The downside to requiring PyArrow if it is not already installed is:

  - PyArrow does not currently distribute musl-c wheels
  - PyArrow requires ~40MB download from pypi and ~130MB on disk

MarcoGorelli: Could you please clarify what you mean by "if the infra isn't there yet"?

The Arrow C Interface already has quite widespread adoption and I'm not aware of edge cases in its implementations. @WillAyd wrote about switching his Pantab project over to it in [Leveraging the Arrow C Data Interface](https://willayd.com/leveraging-the-arrow-c-data-interface.html#arrow-c-data-interface-to-the-rescue), and noted

> Almost immediately my issues went away [...] I felt more confident in the implementation and had to deal with less memory corruption / crashes than before. And, perhaps most importantly, I saved a lot of time.

That was nearly a year ago, and given that he's now suggesting it here in Plotly, I'd say that his experience has stayed just as positive

---

Regarding PyArrow dependency, I'll also note that `polars.DataFrame.to_pandas` also requires PyArrow, so any Polars user (such as myself) would already have needed PyArrow installed if they were converting to pandas via the Polars official method

mwaskom: Basically my threshold is "do I need to think about it _at all_". I'm just not interested in the minutia of competing Python dataframe libraries or the various attempts to make them work better together. The previous approach was sold as a simple protocol that always works, but it turns out that wasn't the case. Maybe this new way is better, the problem is I have no real way to say for sure without spending a lot of time learning about something that doesn't interest me.

MarcoGorelli: Shall I close and leave you to remove cross-dataframe compatibility altogether?

MarcoGorelli: 😄 that's understandable

I'm aware that you said that using Narwhals was a complete non-starter, but just to showcase that as a possibility:
```python
import narwhals.stable.v1 as nw
from narwhals.stable.v1.typing import IntoDataFrame
import polars as pl
import pandas as pd

def convert_dataframe_to_pandas(data: IntoDataFrame) -> pd.DataFrame:
    return nw.from_native(data).to_pandas()
```
and then leave it up to Narwhals to convert to pandas in the best way for each input library

Altair, Plotly, and Vegafusion are using it as required dependency now, and Bokeh have a PR in progress to do the same

MarcoGorelli:
thanks all for the discussions

I'll close this due to lack of interest (which is totally understandable, I'm not complaining), and because pandas will use the PyCapsule Interface in `from_dataframe` from pandas 3.0 onwards (https://github.com/pandas-dev/pandas/pull/60739)

PR #3786, opened 2024-11-17
Fix spacing issue in _spacing function

juhabae:
- Fixed `_spacing` function to calculate spacing for grouped data.
- Ensured proper handling of grouped values by using the minimum spacing within each group.
- Verified changes with existing tests.


mwaskom:
There's not enough context here to evaluate sorry.

Issue #3788, opened 2024-11-19
[FEATURE REQUEST] New Colour Palettes


OWissett:
I have a few nice colour palettes, and want to get them included in seaborn.

'synthwave'

![image](https://github.com/user-attachments/assets/eb860131-662b-48fb-ade0-918cb7522f67)

mwaskom:
Hi, it's not necessary to add new colormaps to seaborn to use them, you can simple pass the list of colors to any seaborn function.

PR #3789, opened 2024-11-19
adds synthwave, nature, paultol, and paultol_dark palettes

OWissett:
#3788 

Adds several more colour palettes.


mwaskom:
See response to #3788; seaborn has only a small number of included color palettes and isn't looking to add any new ones, but you don't need these added to the library to use them.

PR #3803, opened 2024-12-16
Support array-like mask in heatmaps

mariosasko:
Besides `np.ndarray` and `pd.DataFrame` objects, I think it would also be nice to support [array-like objects](https://numpy.org/doc/stable/user/basics.interoperability.html#the-array-method) (PyTorch, Jax, PyArrow arrays, etc.) as the heatmap `mask`.


mwaskom:
LGTM thanks

PR #3811, opened 2025-01-09
fix pointplot(native_scale=True, ...)

v4hn:
Just ran into this plotting something.

Sorry if this should be done cleaner.
I removed the backward compatible code because it failed with working native_scale.

### DIFF
--- a/seaborn/categorical.py
+++ b/seaborn/categorical.py
@@ -2496,14 +2496,10 @@ def pointplot(
     if p.plot_data.empty:
         return ax
 
-    if p.var_types.get(p.orient) == "categorical" or not native_scale:
+    if p.var_types.get(p.orient) == "categorical" and not native_scale:
         p.scale_categorical(p.orient, order=order, formatter=formatter)
 
     p._attach(ax, log_scale=log_scale)
-
-    # Deprecations to remove in v0.14.0.
-    hue_order = p._palette_without_hue_backcompat(palette, hue_order)
-    palette, hue_order = p._hue_backcompat(color, palette, hue_order)
 
     p.map_hue(palette=palette, order=hue_order, norm=hue_norm)
     color = _default_color(ax.plot, hue, color, kwargs)


mwaskom:
Please supply an example that reproduces the problem you ran into. This works fine for me:

```python
import seaborn as sns
sns.pointplot(tips, x="size", y="total_bill", native_scale=True)
```
image

v4hn:
I hit this in the middle of designing a plot combining lineplot and barplot and eventually moved to manual x axis assignment for individual plots, so I cannot reproduce this in my original context anymore.

Note that your example cannot illustrate the effect `native_scale` has if I understood it correctly because the values are evenly spaced.
Playing around with your example a bit I can at least produce the issue below, which might reflect my previous situation. I'm not sure whether you consider this a user error or a bug on the framework side:

```
import seaborn as sns
tips['ssize'] = tips['size'].apply(lambda x: x**4).astype('category')
tips['ssize'] = tips['ssize'].cat.reorder_categories(sorted(tips['ssize'].cat.categories), ordered=True)
sns.pointplot(data=tips, x='ssize', y='total_bill', native_scale=True, legend=False)
```
pointplot

In the example I would have expected this result, which I can also get by explicitly converting the category to float:
pointplot

I believe my original example was slightly more convolved with grouped categories though if that can make a difference in interpretation.

I also believe my current patch is insufficient to address the issue as I see errors when I do not specify `legend`.

v4hn:
That does make sense. Thank you for the clarification!

As it does not seem to be a reasonable/useful function call, I would have appreciated a warning when `native_scale=True` is specified with a categorical variable. Something like

    native_scale=True has no effect on categorical variables.
    Did you intend a different dtype?

mwaskom:
I could see how that would be helpful but I'm generally opposed to issuing warnings where the user isn't doing something unambiguously "wrong", otherwise they can get very noisy and annoying.

v4hn:
I do not quite agree as someone who tripped over it, but the argument is valid.

Last suggestion before close then:

The documentation for the parameter (in at least pointplot and boxplot) states

    When True, **numeric or datetime values on the categorical axis** will maintain
    their original scaling rather than being converted to fixed indices.

I would suggest to rephrase this to indicate that the dtype of the categorical axis must not be category for this. Something like:

    When True and dtype of the categorical axis is numeric or datetime,
    the categorical axis will maintain the original scaling rather than convert to even spacing.

v4hn:
Yes, I tried. Do you have a more accurate suggestion? :)

With this insight my personal (user) intuition is that dtype "category" should be treated the same as dtype "object" with regards to the `native_scale` parameter, but you already stated you disagree.

Another suggestion:

> When True, numeric or datetime values on the categorical axis will maintain
> their original scaling rather than being converted to fixed indices.
> The native scale of a Series with the explicit category dtype always uses fixed indices.

PR #3859, opened 2025-06-21
Add `scipy-stubs` as dev dependency

jorenham:
I noticed that `pandas-stubs` was already in the list, so I figured that you might also be interested in [`scipy-stubs`](https://github.com/scipy/scipy-stubs) :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dev = [
     "flake8",
     "mypy",
     "pandas-stubs",
-    "scipy-stubs",
+    "scipy-stubs; python_version>='3.10'",
     "pre-commit",
     "flit",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dev = [
     "flake8",
     "mypy",
     "pandas-stubs",
+    "scipy-stubs",
     "pre-commit",
     "flit",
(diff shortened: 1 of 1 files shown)


mwaskom:
Looks like you need to limit it to 3.11+ per the distribution requirements: https://github.com/scipy/scipy-stubs/blob/907fff919bfee5a2302dd7a99e7c346d33f7a935/pyproject.toml#L32

jorenham:
> Looks like you need to limit it to 3.11+ per the distribution requirements: [scipy/scipy-stubs@`907fff9`/pyproject.toml#L32](https://github.com/scipy/scipy-stubs/blob/907fff919bfee5a2302dd7a99e7c346d33f7a935/pyproject.toml#L32)

That's only if you use `scipy==1.16.*`, because earlier versions of `scipy-stubs`, such as `1.14.1.6`, have Python 3.10. The logic behind this, is that`scipy-stubs==X.Y.Z.*` targets `scipy==X.Y.Z`, and therefore matches its python restrictions. And in the recent 1.16 scipy release, support for python 3.10 was dropped.

mwaskom:
Pulled in scipy stubs in #3954

jorenham:
> Pulled in scipy stubs in #3954

Awesome!

PR #3874, opened 2025-09-24
fix(distributions): Added check if estimate_kws["bins"] is a str. 

Hannes237:
A list/array would make the comparison ambiguous. This prevents you to use 'weights' and 'bins' at the same time


mwaskom:
Fixed in #3944

PR #3911, opened 2026-02-13
Handle common_norm for density stat in Hist

### LINKED ISSUE #3633: `so.Hist` ignores `common_norm=True` for the `"density"` aggregate statistic
### Current behaviour

`so.Hist` ignores `common_norm=True` for the `"density"` aggregate statistic. `sns.histplot` works correctly.

### Desired behaviour

`so.Hist` should take into account `common_norm` when applying density normalization. The objects interface adds some neat functionality for [normalizing per facet](https://github.com/mwaskom/seaborn/issues/3112#issuecomment-1292870232). Currently it can't be used with `"density"`.

### Examples

I would expect the below examples to produce the same histogram.

```python
(
    so.Plot(
        penguins, 
        x='bill_length_mm', 
        color='species'
    )
    .add(
        so.Bars(), 
        so.Hist(
            stat='density',
            common_norm=True,
        ), 
    )
)
```
240214_b

```python
sns.histplot(
    data=penguins, 
    x='bill_length_mm', 
    hue='species',
    stat='density',
    common_norm=True,
)
```
240214_a

### Versions

- seaborn 0.13.2
- matplotlib 3.8.2
- pandas 2.2.0

Mr-Neutr0n:
## Summary

`so.Hist(stat='density', common_norm=True)` was silently ignoring `common_norm` — each group's density was normalized independently instead of across all groups.

The root cause: `_eval()` passed `density=True` to `np.histogram`, which baked in per-group normalization before `_normalize()` (and its `common_norm` grouping logic) ever ran. For every other stat (`percent`, `proportion`, `frequency`), normalization happens inside `_normalize()` where `common_norm` is properly respected — but `density` was the exception.

## Changes

- **`_eval()`**: Always pass `density=False` to `np.histogram` so raw counts are preserved for all stats.
- **`_normalize()`**: Add density normalization (`count / (total_count * bin_width)`) alongside the existing percent/proportion/frequency cases. This way `common_norm` controls the scope of `total_count` for density just like it does for the other stats.

## Test plan

- Added `test_common_norm_density_default` — verifies total area sums to 1 across all groups when `common_norm=True`
- Added `test_common_norm_density_false` — verifies each group individually sums to 1 when `common_norm=False`
- Added `test_common_norm_density_subset` — verifies normalization within subgroups
- All 32 tests in `test_counting.py` pass (29 existing + 3 new)

Closes #3633

### DIFF
--- a/seaborn/_stats/counting.py
+++ b/seaborn/_stats/counting.py
@@ -172,8 +172,7 @@ class Hist(Stat):
         vals = data[orient]
         weights = data.get("weight", None)
 
-        density = self.stat == "density"
-        hist, edges = np.histogram(vals, **bin_kws, weights=weights, density=density)
+        hist, edges = np.histogram(vals, **bin_kws, weights=weights, density=False)
 
         width = np.diff(edges)
         center = edges[:-1] + width / 2
@@ -183,7 +182,9 @@ class Hist(Stat):
     def _normalize(self, data):
 
         hist = data["count"]
-        if self.stat == "probability" or self.stat == "proportion":
+        if self.stat == "density":
+            hist = hist.astype(float) / (hist.sum() * data["space"])
+        elif self.stat == "probability" or self.stat == "proportion":
             hist = hist.astype(float) / hist.sum()
         elif self.stat == "percent":
             hist = hist.astype(float) / hist.sum() * 100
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #3912, opened 2026-02-13
Fix crash with custom tick formatters/locators and Est()

### LINKED ISSUE #3585: so.Est() + tick label formatter => 'PseudoAxis' object has no attribute '_view_interval'
Hi, thanks for cool library, I especially enjoy using objects interface.

While using it I encountered a small issue when trying to customise tick labels using matplotlib formatter in combination with `so.Est()`. If I replace it with `so.Agg()` the issue disappears. I guess this has something to do with the fact that several variables use the same scale (y/ymin/ymax) and `_view_interval` is not initialised properly in that case?

```python
import seaborn as sns
import seaborn.objects as so
import matplotlib as mpl

fmri = sns.load_dataset("fmri")
(
    so.Plot(fmri, x="timepoint", y="signal", color="event")
        .add(so.Band(), so.Est())
        .scale(y=so.Continuous().label(mpl.ticker.PercentFormatter()))  
)
```

```
File ~/.pyenv/versions/miniforge3/lib/python3.9/site-packages/matplotlib/ticker.py:233, in Formatter.format_ticks(self, values)
    231 """Return the tick labels for all the ticks at once."""
    232 self.set_locs(values)
--> 233 return [self(value, i) for i, value in enumerate(values)]

File ~/.pyenv/versions/miniforge3/lib/python3.9/site-packages/matplotlib/ticker.py:233, in <listcomp>(.0)
    231 """Return the tick labels for all the ticks at once."""
    232 self.set_locs(values)
--> 233 return [self(value, i) for i, value in enumerate(values)]

File ~/.pyenv/versions/miniforge3/lib/python3.9/site-packages/matplotlib/ticker.py:1525, in PercentFormatter.__call__(self, x, pos)
   1523 def __call__(self, x, pos=None):
   1524     """Format the tick as a percentage with the appropriate scaling."""
-> 1525     ax_min, ax_max = self.axis.get_view_interval()
   1526     display_range = abs(ax_max - ax_min)
   1527     return self.fix_minus(self.format_pct(x, display_range))

File ~/.pyenv/versions/miniforge3/lib/python3.9/site-packages/seaborn/_core/scales.py:921, in PseudoAxis.get_view_interval(self)
    920 def get_view_interval(self):
--> 921     return self._view_interval

AttributeError: 'PseudoAxis' object has no attribute '_view_interval'
```

Versions:
seaborn: 0.13.0
matplotlib: 3.6.2

Mr-Neutr0n:
## Summary

Fixes #3585

When using `so.Est()` with custom matplotlib formatters (e.g. `PercentFormatter`) or locators (e.g. `LogLocator`), the plot crashes with `AttributeError: 'PseudoAxis' object has no attribute '_view_interval'` (or `'_get_shared_axis'`).

The root cause is that `_get_scale()` returns locator/formatter instances that are shared across multiple calls to `_setup()`. When a scale is set up for derived coordinate variables like `ymin`/`ymax` (which happens with `Est()`), a `PseudoAxis` is created internally, and `set_default_locators_and_formatters` overwrites the shared locator's axis reference to point at this `PseudoAxis`. The main variable's scale then tries to use the locator, which now references a `PseudoAxis` that lacks attributes like `_view_interval`.

## Changes

- **Deepcopy locators and formatter in `_get_scale()`** so each scale setup gets independent instances that won't interfere with each other
- **Initialize `_view_interval` in `PseudoAxis.__init__`** as a defensive measure, matching the existing pattern for `_data_interval`

## Test plan

- [x] Verified fix with both reproduction cases from the issue (`PercentFormatter` and `LogLocator` with `Est()`)
- [x] Full test suite passes (2361 passed, 14 skipped, 6 xfailed)

### DIFF
--- a/seaborn/_core/scales.py
+++ b/seaborn/_core/scales.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import re
-from copy import copy
+from copy import copy, deepcopy
 from collections.abc import Sequence
 from dataclasses import dataclass
 from functools import partial
@@ -85,6 +85,17 @@ class Scale:
 
         major_locator, minor_locator = self._get_locators(**self._tick_params)
         major_formatter = self._get_formatter(major_locator, **self._label_params)
+
+        # Deepcopy locators and formatter so that each scale setup gets
+        # independent instances. Without this, scales for derived coordinate
+        # variables (e.g. ymin/ymax from Est) share locator/formatter objects
+        # with the main variable's scale, and the locator's internal axis
+        # reference gets overwritten to point at a PseudoAxis that lacks
+        # required attributes like _view_interval. See GH#3585.
+        major_locator = deepcopy(major_locator)
+        if minor_locator is not None:
+            minor_locator = deepcopy(minor_locator)
+        major_formatter = deepcopy(major_formatter)
 
         class InternalScale(mpl.scale.FuncScale):
             def set_default_locators_and_formatters(self, axis):
@@ -909,6 +920,7 @@ class PseudoAxis:
         # It appears that this needs to be initialized this way on matplotlib 3.1,
         # but not later versions. It is unclear whether there are any issues with it.
         self._data_interval = None, None
+        self._view_interval = None, None
 
         scale.set_default_locators_and_formatters(self)
         # self.set_default_intervals()  Is this ever needed?


Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #3927, opened 2026-04-07
Add param show_data_points to barplot for plotting individual data points

ym3141:
There has been lots discussion (or push) recently in the academia to plot individual data points along with traditional bar plots. Here I added a new convenient option in the barplot to show the underlying data points by swarmplot.
[Discussion from Nature's editorial team](https://www.nature.com/articles/s41551-017-0079); [Another research](https://pmc.ncbi.nlm.nih.gov/articles/PMC9366861/).


mwaskom:
The way this works in seaborn is that you layer the two plotting functions, see examples in the docs about this. Thanks!

PR #3934, opened 2026-04-30
Fix Scale initialization

QuLogic:
The first parameter to `matplotlib.scale.LinearScale` and `matplotlib.scale.FuncScale` is, and always has been, an `Axis`, not a name, which has never been used.

Matplotlib 3.11 has made that argument optional internally, but it can only do so if the argument is actually an `Axis`.

Firstly, fix that argument so it's actually the `Axis`, and secondly, only pass if for older Matplotlib.

Note, I am still trying to determine if this could be made simpler upstream: https://github.com/matplotlib/matplotlib/issues/31590

### DIFF
--- a/seaborn/_core/scales.py
+++ b/seaborn/_core/scales.py
@@ -36,6 +36,7 @@ from pandas import Series
 
 from seaborn._core.rules import categorical_order
 from seaborn._core.typing import Default, default
+from seaborn.utils import _version_predates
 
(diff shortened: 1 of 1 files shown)


QuLogic:
https://github.com/matplotlib/matplotlib/pull/31621 should make it so that this is no longer an error-inducing problem, but it should still be fixed on this side.

mwaskom:
Thanks!!

PR #3935, opened 2026-05-01
TST: Adjust results for Matplotlib 3.11

QuLogic:
Log scaling underwent some optimization, which produces slightly different results, but still within some small floating-point tolerance.


mwaskom:
Amazing, thank you!

PR #3944, opened 2026-06-09
Fix histplot with weights and array bins

### LINKED ISSUE #3801: Error for `histplot` with weights and explicit bins
When I try to make a `histplot` with weights and explicitly provided bin edges *as a numpy array*, e.g.:
```
import numpy as np, seaborn as sns

sns.histplot(x=np.random.rand(100), weights=np.ones(100), bins=np.linspace(0, 1, 11))
```
I get the error
```
ValueError: The truth value of an array with more than one element is ambiguous. ...
```
It arises in this line:
https://github.com/mwaskom/seaborn/blob/b4e5f8d261d6d5524a00b7dd35e00a40e4855872/seaborn/distributions.py#L404
and should be easily solvable by wrapping it in `is True`:
```
and (estimate_kws["bins"] == "auto") is True
```

its-amann:
Fixes #3801.

This updates the weighted-histogram auto-bin warning guard so it only compares `bins` with `"auto"` when `bins` is a string. Explicit bin edges provided as a NumPy array are now passed through to the histogram estimator instead of raising an ambiguous truth-value error.

Added a regression test covering `histplot(x=..., weights=..., bins=np.array(...))`.

Verification:

- `python -m pytest tests/test_distributions.py::TestHistPlotUnivariate::test_weights_with_array_bins -q`
- `MPLBACKEND=Agg python -m pytest tests/test_distributions.py::TestHistPlotUnivariate -q`
- `MPLBACKEND=Agg python -m pytest tests/test_distributions.py::TestHistPlotUnivariate::test_weights_with_missing tests/test_distributions.py::TestHistPlotUnivariate::test_weight_norm tests/test_distributions.py::TestHistPlotUnivariate::test_weights_with_array_bins -q`
- `MPLBACKEND=Agg python -m pytest tests/test_statistics.py::TestHistogram::test_array_bins -q`
- `git diff --check`

### DIFF
--- a/seaborn/distributions.py
+++ b/seaborn/distributions.py
@@ -401,6 +401,7 @@ class _DistributionPlotter(VectorPlotter):
 
         auto_bins_with_weights = (
             "weights" in self.variables
+            and isinstance(estimate_kws["bins"], str)
             and estimate_kws["bins"] == "auto"
             and estimate_kws["binwidth"] is None
             and not estimate_kws["discrete"]
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -1461,6 +1461,19 @@ class TestHistPlotUnivariate(SharedAxesLevelTests):
         total_weight = null_df[["x", "s"]].dropna()["s"].sum()
         assert sum(bar_heights) == pytest.approx(total_weight)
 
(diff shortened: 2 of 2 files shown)


mwaskom:
Thanks!

PR #3945, opened 2026-06-12
Improve build instructions for docs

SuperCowProducts:
These are the commands I used to create a PDF. LaTeX errors, yet a 1000-page PDF is there in all its goodness. I think the important things to point out for users like me wanting to build from source are:
1. use Python 3.12
2. Run `python -m ipykernel install --user --name=myenv --display-name "Python (myenv)"`
3. `make latexpdf` may be unsupported, but it works! (it would also be nice to provide a download link for it, it is much appreciated rather than having to build each time with high risk of failing :) )


mwaskom:
Repo uses uv now.

PR #3982, opened 2026-08-31
Fix common normalization for density histograms

### LINKED ISSUE #3633: `so.Hist` ignores `common_norm=True` for the `"density"` aggregate statistic

yushuosun:
`Hist._eval` currently asks NumPy to normalize each semantic group independently when `stat="density"`. By the time seaborn applies its `common_norm` grouping, the original counts have already been lost.

This keeps raw histogram counts in `_eval` and moves density normalization into `_normalize`, alongside the other normalized statistics. As a result:

- the combined area is 1 with the default `common_norm=True`;
- each group has area 1 with `common_norm=False`;
- list-valued `common_norm` normalizes within each requested subset.

The full `_stats` test suite passes (74 passed, 2 skipped), including three regression tests added here.

This revives the approach from #3911, which was withdrawn by its author while the issue remained open. Thanks @Mr-Neutr0n for the original diagnosis and test cases.

Closes #3633.


yushuosun:
I reopened this on latest master as: https://github.com/mwaskom/seaborn/pull/3989

yushuosun:
I opened a refreshed PR against current master: https://github.com/mwaskom/seaborn/pull/3991

yushuosun:
Superseded by #3991 refresh.

PR #3989, opened 2026-09-09
Fix common normalization for density histograms (refresh)

yushuosun:
Reopened on top of latest master to avoid stale base from #3982.

Replaces old PR branch: #3982


yushuosun:
Superseded by newer refresh PRs 3992/3991.

PR #3983, opened 2026-08-31
Handle markers=False in wide-form scatterplots

yushuosun:
Closes #3652.

Wide-form scatterplots automatically map columns onto `style`. With
`markers=False`, that mapping intentionally contains no marker or path, but
`_ScatterPlotter` still attempted to look both up and raised a `KeyError`.

This keeps Matplotlib's default marker when no marker mapping is present and
only applies per-point paths when markers are actually mapped.

Tests:

- `MPLBACKEND=Agg python -m pytest tests/test_relational.py -q` (121 passed)
- `python -m ruff check seaborn/relational.py tests/test_relational.py`

PR #3990, opened 2026-09-09
Handle markers=False in wide-form scatterplots (refresh)

yushuosun:
Reopened on top of latest master to avoid stale base from #3983.

Replaces old PR branch: #3983


yushuosun:
Superseded by newer refresh PRs 3992/3991.

